PR #32, opened 2018-01-05
fix/admin: add admin field to JWTs

rudyardrichter:


CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# `fence` Changelog
```
Avantol13: I love the idea of a changelog, but if we're doing this manually, it's going to take a lot of diligence and willingness to update this. And then we need to know when to update and bump the version. Might be nice to automatically pull in completed issues somehow? idk, just ranting about it. Like the idea, not sure if we'll be able to maintain a manually updated changelog

setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name='fence',
-    version='0.1.0',
+    version='0.2.0',
```
Avantol13: Do we have a notion of when a minor vs major bump is required? This seems like a breaking change. I'm assuming we're waiting until 1.0.0 to "ensure" backwards compatibility between major versions?

rudyardrichter: This should be backwards compatible, since it's just adding an extra field to the tokens (and not removing any).

CHANGELOG.md
rudyardrichter: That's a good point, and not something we've discussed enough yet I think. For now, I figured no harm in starting it here.

Issue #24, opened 2017-12-22
list refresh token for users missing


philloooo:
`GET /credentials/cdis/` is not handled
https://github.com/uc-cdis/fence/blob/10d894983fa7c550dcdfbd73f772b71fafbc169a/fence/blueprints/storage_creds.py#L66

PR #35, opened 2018-01-09
fix(cdis_refresh_token): Fix mising of refresh_token for cdis backend

giangbui:
Added new table 'refresh_token' to DB
- insert new refresh token whenever new token is generated
- query the table to obtain refresh token

@thanh-nguyen-dang , @rudyardrichter , @philloooo

### DIFF
--- a/fence/blueprints/storage_creds.py
+++ b/fence/blueprints/storage_creds.py
@@ -4,12 +4,11 @@ from fence.data_model.models import GoogleProxyGroup
 import flask
 from flask import current_app as capp
 from flask import g, request, jsonify
+from ..resources.storage import get_endpoints_descriptions
 from flask import abort
 from flask_sqlalchemy_session import current_session
 
 from cirrus import GoogleCloudManager
-from ..resources.storage import get_endpoints_descriptions
-
 from fence.resources.storage.cdis_jwt import create_refresh_token,\
     revoke_refresh_token, create_access_token
 
@@ -92,7 +91,12 @@ def list_keypairs(provider):
     .. code-block:: JavaScript
         cdis
         {
-            "jti": "e9d58890-99b0-44f0-88bd-3ebc370b1329"
+            "jtis":
+            [
+                {
+                    "jti": "e9d58890-99b0-44f0-88bd-3ebc370b1329"
+                }
+            ]
         }
         non-cdis
         {
--- a/fence/blueprints/storage_creds.py
+++ b/fence/blueprints/storage_creds.py
@@ -90,7 +90,11 @@ def list_keypairs(provider):
     other:
 
     .. code-block:: JavaScript
-
+        cdis
+        {
+            "jti": "e9d58890-99b0-44f0-88bd-3ebc370b1329"
+        }
+        non-cdis
         {
             "access_keys":
             [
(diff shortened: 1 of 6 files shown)


giangbui:
resolve #24

tests/conftest.py
```diff
@@ -138,12 +138,12 @@ def app(request):
     root_dir = os.path.dirname(os.path.realpath(__file__))
     app_init(fence.app, test_settings, root_dir=root_dir)
 
-    def fin():
-        for tbl in reversed(Base.metadata.sorted_tables):
-            fence.app.db.engine.execute(tbl.delete())
-        mocker.unmock_functions()
+    # def fin():
```
rudyardrichter: Why is this removed?

fence/blueprints/storage_creds.py
```diff
@@ -66,6 +68,10 @@ def list_credentials(backend):
     if backend != 'cdis':
         result = capp.storage_manager.list_keypairs(backend, g.user)
         return jsonify(dict(access_keys=result))
+    if backend == 'cdis':
+        with capp.db.session as session:
+            tokens = session.query(RefreshToken.token).filter_by(userid=g.user.id).all()
+            return jsonify(dict(token=tokens))
```
rudyardrichter: this would look cleaner with the conditionals reordered as:
```python
if backend == 'cdis':
    with capp.db.session as session:
        # ...
else:
    result = capp.storage_manager.list_keypairs(backend, g.user)
    # ...
```

fence/data_model/models.py
```diff
@@ -68,6 +68,18 @@ def validate_scopes(self, scopes):
...
+    __tablename__ = "refresh_token"
+
+    token = Column(String, primary_key=True)
+    userid = Column(Integer)
+
+    def delete(self):
+        with capp.db.session as session:
+            session.delete(self)
+            session.commit()
+            return self
```
rudyardrichter: What is the `return self` for?

fence/blueprints/storage_creds.py
giangbui: fixed!

fence/data_model/models.py
```diff
@@ -68,6 +70,18 @@ def validate_scopes(self, scopes):
         scopes = scopes[0].split(',')
         return all(scope in self._scopes for scope in scopes)
 
+class RefreshToken(Base):
+    __tablename__ = "refresh_token"
+
+    token = Column(String, primary_key=True)
+    userid = Column(Integer)
+    created_time = Column(DateTime, default=datetime.datetime.utcnow().isoformat)
```
rudyardrichter: Rather than a `created_time` it would be better to get the `exp` (expiration) field from the token so that cleaning out old entries from the table can just check that column against the current unix time.

fence/data_model/models.py
```diff
@@ -68,6 +70,18 @@ def validate_scopes(self, scopes):
         scopes = scopes[0].split(',')
         return all(scope in self._scopes for scope in scopes)
 
+class RefreshToken(Base):
+    __tablename__ = "refresh_token"
+
+    token = Column(String, primary_key=True)
```
rudyardrichter: Similarly to the `exp` thing, the tokens have a `jti` (JWT ID) field which would make for a better primary key than just using the entire token.

giangbui:
Codacy comments the UserRefreshToken is unused but it is used in storage_creds.py. Probably, it is codacy bug.

fence/blueprints/storage_creds.py
```diff
@@ -69,7 +71,12 @@ def list_keypairs(provider):
         }
 
     '''
-    if provider != 'cdis':
+    if provider == 'cdis':
```
thanh-nguyen-dang: should add an example of return results so that people use it can know about the format.

PR #36, opened 2018-01-09
feat(dbgap): add dbgap syncing script

philloooo:
add script to sync from dbgap to both database and storage backends
depends on https://github.com/uc-cdis/cdisutils-test/pull/5

### DIFF
--- a/fence/blueprints/storage_creds.py
+++ b/fence/blueprints/storage_creds.py
@@ -2,6 +2,9 @@ from ..auth import login_required
 import flask
 from flask import current_app as capp
 from flask import g, request, jsonify
+from flask_sqlalchemy_session import current_session
+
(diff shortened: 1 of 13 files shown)


fence/blueprints/storage_creds.py
```diff
@@ -84,30 +94,29 @@ def create_credential(backend):
            Accept: application/json
 
     .. code-block:: JavaScript
-        cdis:
-        {
-            "token": "token_value"
-        }
-        non-cdis:
+
```
thanh-nguyen-dang: cdis needs different return format, so, keep this script

philloooo:
merging and ignoring codacy because this is admin syncing script that uses subprocess that codacy doesn't like

PR #37, opened 2018-01-10
feat(session): replace with jwt

Avantol13:


fence/jwt/token.py
```diff
@@ -25,12 +23,58 @@ def issued_and_expiration_times(seconds_to_expire):
...
         Tuple[int, int]: (issued, expired) times in unix time
     """
-    now = datetime.now()
+    now = datetime.utcnow()
     iat = int(now.strftime('%s'))
     exp = int((now + timedelta(seconds=seconds_to_expire)).strftime('%s'))
     return (iat, exp)
 
 
+def generate_signed_session_token(kid, private_key, expires_in,
```
rudyardrichter: [style nitpick](https://github.com/uc-cdis/rfcs/blob/master/text/0003-style-guide.md#breaking-long-lines)

fence/resources/storage/cdis_jwt.py
```diff
@@ -2,6 +2,13 @@
 from flask import jsonify
 
 
+def create_session_token(keypair, expires_in, username=None,
```
rudyardrichter: same as previous style nitpick

fence/__init__.py
```diff
@@ -14,9 +13,10 @@
 from .blueprints.oauth2 import blueprint as oauth2
```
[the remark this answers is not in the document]
Avantol13: Fixed in 86da04b

fence/jwt/token.py
Avantol13: Fixed in 86da04b

fence/resources/storage/cdis_jwt.py
Avantol13: Fixed in 86da04b

fence/resources/user/user_session.py
```diff
@@ -0,0 +1,229 @@
...
+
+def _create_access_token_cookie(app, response, user):
+    keypair = app.keypairs[0]
+    scopes = USER_ALLOWED_SCOPES
+    access_token = generate_signed_access_token(
+        keypair.kid, keypair.private_key, user,
+        app.config.get('ACCESS_TOKEN_LIFETIME').seconds, scopes,
+        client_id=user.id
+    )
+    timeout = datetime.utcnow() + app.config.get('ACCESS_TOKEN_LIFETIME')
```
philloooo: should use the same expiration time for both cookie and exp in jwt so we don't encounter a case that the cookie is not expired but the jwt expired and we failed to generate a new access_token

fence/jwt/token.py
```diff
@@ -25,12 +29,58 @@ def issued_and_expiration_times(seconds_to_expire):
...
+        'aud': ['session'],
+        'sub': username or '',
+        'iss': flask.current_app.config.get('HOST_NAME'),
+        'iat': iat,
+        'exp': exp,
+        'jti': str(uuid.uuid4()),
+        'context': context,
+    }
+    flask.current_app.logger.info(
+        'issuing JWT session token\n' + json.dumps(claims, indent=4)
```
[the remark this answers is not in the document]
rudyardrichter: This one shows up elsewhere, actually (access & refresh in fence). I had it at info-level for the claims json and debug for the encoded token string, but they could be changed to both be debug.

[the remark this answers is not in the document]
rudyardrichter: Maybe we can change it to have an info log with something short ("issued JWT with id [`jti`] to [`sub`]") and put the whole token on debug?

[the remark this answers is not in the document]
philloooo: Yeah that sounds good. I am fine as long as we don’t log.info the whole thing

fence/resources/user/user_session.py
```diff
@@ -0,0 +1,229 @@
...
+        )
+
+    def get_updated_token(self, app):
+        if self._encoded_jwt:
+            timeout = app.config.get('SESSION_TIMEOUT')
+
+            token = create_session_token(
+                current_app.keypairs[0],
+                timeout.seconds,
+                session_started=self.get("session_started", None),
```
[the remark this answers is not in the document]
Avantol13: this logic actually lives in `generate_signed_session_token`. It will add the `session_started` time to the context as either the given time (to keep sliding until max lifetime) or if `None` is provided (e.g. first token) it will use the `iat` (issued at time) field for the JWT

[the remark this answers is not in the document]
Avantol13: I'll add a comment to clarify here

fence/resources/user/user_session.py
Avantol13: Ah, yup. Missed that. I'm looking though and I can't think of an elegant solution for this.

The JWT expiration (`exp`) is set by `generate_signed_access_token` which takes an `expires_in` field which it adds to current time to get `exp`.

That function has to have that field since the `expires_in` comes from the request in oauth. So, I see two viable options:

1) Make a new function to generate a signed access token by providing the exact `exp` (which will duplicate some code)
2) Modify `generate_signed_access_token` to take an optional parameter to override the calculated `exp` with a given time

Both solutions don't seem very great, one duplicates code and the other essentially ignores a given argument in favor of another. I also didn't include the possible option of timing how long it takes to generate the token and adjusting the timeout of the cookie to compensate :P 

I'm leaning towards option 2 (add some sort of optional `override_exp` argument) unless a more clever idea surfaces. I'll pursue that for now

Avantol13: Ah, yup. Missed that. I'm looking though and I can't think of an elegant solution for this.

The JWT expiration (`exp`) is set by `generate_signed_access_token` which takes an `expires_in` field which it adds to current time to get `exp`.

That function has to have that field since the `expires_in` comes from the request in oauth. So, I see two viable options:

1) Make a new function to generate a signed access token by providing the exact `exp` (which will duplicate some code)
2) Modify `generate_signed_access_token` to take an optional parameter to override the calculated `exp` with a given time

Both solutions don't seem very great, one duplicates code and the other essentially ignores a given argument in favor of another. I also didn't include the possible option of timing how long it takes to generate the token and adjusting the timeout of the cookie to compensate :P 

I'm leaning towards option 2 (add some sort of optional `override_exp` argument) unless a more clever idea surfaces. I'll pursue that for now

fence/jwt/token.py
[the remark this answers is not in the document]
Avantol13: fixed in f9a3003

fence/resources/user/user_session.py
```diff
@@ -0,0 +1,229 @@
...
+from fence.jwt.token import generate_signed_access_token
+from fence.jwt.token import USER_ALLOWED_SCOPES
+from fence.resources.storage.cdis_jwt import create_session_token
+from fence.errors import Unauthorized
+from fence import auth
+
+
+class UserSession(SessionMixin):
+    def __init__(self, jwt):
+        self._encoded_jwt = jwt
```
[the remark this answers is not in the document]
Avantol13: fixed in 26f252c

fence/resources/user/user_session.py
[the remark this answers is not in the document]
Avantol13: added in 26f252c

fence/resources/user/user_session.py
Avantol13: option 2 implemented in 1346652

Avantol13:
bah, need to fix hostname, it broke when I merged. Will fix later today

fence/resources/user/user_session.py
```diff
@@ -0,0 +1,236 @@
...
+        clear current session
+        """
+        self._encoded_token = None
+        self.session_token = {"context": {}}
+
+    def clear_if_expired(self, app):
+        if self._encoded_token:
+            lifetime = app.config.get('SESSION_LIFETIME')
+
+            now = int(datetime.utcnow().strftime('%s'))
```
rudyardrichter: It seems `utcnow`, misleadingly, will actually produce the wrong timestamp (what the timestamp *would* be after the timezone difference):
```
In [1]: import time

In [2]: from datetime import datetime

In [3]: print('UNIX time: {}'.format(time.time()))
   ...: now = datetime.now()
   ...: print('datetime.now() == {}'.format(int(now.strftime('%s'))))
   ...: utcnow = datetime.utcnow()
   ...: print('datetime.utcnow() == {}'.format(int(utcnow.strftime('%s'))))
   ...:
UNIX time: 1516817378.49
datetime.now() == 1516817378
datetime.utcnow() == 1516838978
```

I think we should move to just using `int(time.time())` for this

fence/resources/user/user_session.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def __len__(self):
+        return len(self.session_token)
+
+
+class UserSessionInterface(SessionInterface):
+
+    def __init__(self):
+        super(UserSessionInterface, self).__init__()
+        self.access_token = None
```
philloooo: the sessioninterface is instantiated on the app level, but access_token should be per request?

philloooo: can store access_token in `flask.g` as that's request scoped

fence/resources/user/user_session.py
Avantol13: wow, yeah, I'll update this

fence/resources/user/user_session.py
Avantol13: mm, good point. I stored it in `flask.g` instead in 8f425a2

PR #40, opened 2018-01-17
feat/data

thanh-nguyen-dang:
add data endpoint and unit-test
@philloooo @rudyardrichter


fence/blueprints/data.py
```diff
@@ -0,0 +1,91 @@
...
+    authorized_dbgaps = []
+    for key, values in list_auth_ids.items():
+        if ('read-storage' in values):
+            authorized_dbgaps.append(key)
+    return authorized_dbgaps
+
+
+def check_authorization(doc):
+    metadata = doc['metadata']
+    set_dbgaps = set(metadata['acls'].split(','))
```
philloooo: need to filter by action/privilege
upload_data should only be allowed by user who has 'write-storage' privilege
download_data should only be allowed by user who has 'read-storage' privilege

philloooo:
1. update swagger doc, current one only has GET /data/{uuid}
2. allow specify `location` to be aws or gcs for download
3. add user_id and username, oauth client_id(if that's provided in the jwt token), to [presigned url query parameters](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html) so that our s3 logs can tell us who the presigned url belongs to,

fence/blueprints/data.py
```diff
@@ -0,0 +1,91 @@
...
+        raise NotFound("Can't find a location for the data")
+    else:
+        raise UnavailableError(res.text)
+
+
+blueprint = flask.Blueprint('data', __name__)
+
+
+@blueprint.route('/download/<file_id>', methods=['GET'])
+@hash_secret
```
philloooo: don't need this decorator?

fence/resources/aws/boto_manager.py
```diff
@@ -0,0 +1,122 @@
...
+            ClientMethod=method,
+            Params={
+                'Bucket': bucket,
+                'Key': key
+            } if method == 'get_object' else {
+                'Bucket': bucket,
+                'Key': key,
+                'ServerSideEncryption': 'AES256'
+            },
+            ExpiresIn=3600*24
```
philloooo: let's default to 30 min and max to 1 day, allow to be provded as a parameter

fence/resources/__init__.py
```diff
@@ -0,0 +1,4 @@
+
+def translate_action(service, action):
```
philloooo: dead code?

fence/resources/actions.json
```diff
@@ -0,0 +1,10 @@
+{
```
philloooo: why not just define them as global variable in data.py

fence/blueprints/data.py
```diff
@@ -0,0 +1,102 @@
...
+def upload_file(file_id):
+    '''
+    Get a presigned url to upload a file given by file_id.
+    '''
+    return get_file('upload', file_id)
+
+
+def return_link(action, urls):
+    protocol = request.args.get('protocol', None)
+    expired_in_mins = request.args.get('expired_in_mins', None)
```
philloooo: let's call it `expires` and use seconds

fence/blueprints/data.py
```diff
@@ -0,0 +1,118 @@
+import flask
+import jwt
+import json
+import requests
+from ..errors import UnavailableError, NotFound, Unauthorized, NotSupported
+from flask import current_app as capp
+from flask import jsonify, request
+from urlparse import urlparse
+
+action_dict = {
```
philloooo: let's uppercase them

fence/blueprints/data.py
```diff
@@ -0,0 +1,118 @@
...
+
+@blueprint.route('/upload/<file_id>', methods=['GET'])
+def upload_file(file_id):
+    '''
+    Get a presigned url to upload a file given by file_id.
+    '''
+    return get_file('upload', file_id)
+
+
+SUPPORTED_PROTOCOLS = ['s3', 'http']
```
philloooo: move this to top

fence/blueprints/data.py
```diff
@@ -0,0 +1,118 @@
...
+
+@blueprint.route('/download/<file_id>', methods=['GET'])
+def download_file(file_id):
+    '''
+    Get a presigned url to download a file given by file_id.
+    '''
+    return get_file('download', file_id)
+
+
+@blueprint.route('/upload/<file_id>', methods=['GET'])
```
philloooo: endpoint should be protected by `@login_required({'data'})` ?

fence/__init__.py
```diff
@@ -41,6 +43,7 @@ def app_config(app, settings='fence.settings', root_dir=None):
     if root_dir is None:
         root_dir = os.path.dirname(
                 os.path.dirname(os.path.realpath(__file__)))
+    app.boto = BotoManager(app.config['AWS'])
```
philloooo: should allow fence to work if there is no aws backend
if AWS is not in config, botomanager is not created and the blueprint is not registered

fence/blueprints/data.py
```diff
@@ -0,0 +1,118 @@
...
+
+def get_index_document(file_id):
+    indexd_server = (
+        capp.config.get('INDEXD') or
+        capp.config['HOSTNAME'] + '/index')
+    url = indexd_server + '/index/'
+    try:
+        res = requests.get(url + file_id)
+    except Exception as e:
+        capp.logger.exception("fail to reach indexd: {0}".format(e))
```
philloooo: let's log the url + file_id too

fence/blueprints/data.py
```diff
@@ -0,0 +1,118 @@
...
+        raise UnavailableError(
+            "Fail to reach id service to find data location")
+    if res.status_code == 200:
+        return res.json()
+    elif res.status_code == 404:
+        capp.logger.exception("fail to reach indexd: {0}".format(res.text))
```
philloooo: log the url+file_id, change error message to `'indexd can't find {} : {}'.format(url+file_id, res.text)`

fence/blueprints/data.py
```diff
@@ -0,0 +1,117 @@
...
+        capp.config['HOSTNAME'] + '/index')
+    url = indexd_server + '/index/'
+    try:
+        res = requests.get(url + file_id)
+    except Exception as e:
+        capp.logger.exception("fail to reach indexd: {0}".format(e))
+        raise UnavailableError(
+            "Fail to reach id service to find data location")
+    if res.status_code == 200:
+        return res.json()
```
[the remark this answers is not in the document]
zflamig: We should also maybe check that the json conforms to what we expect from indexd so that it has urls and metadata keys. To prevent from returning non-descriptive 500 errors.

zflamig:
I think we should include a list of buckets and associated credentials in the config too instead of just 1 set. That way we can support presigning for buckets we have access too that are in different accounts? This would also give us the ability to turn on/off access to specific buckets on the whole.

fence/blueprints/data.py
```diff
@@ -0,0 +1,117 @@
...
+        raise NotSupported("The specified protocol is not supported")
+    for url in urls:
+        location = urlparse(url)
+        if check_protocol(protocol, location.scheme):
+            return resolve_url(location, protocol, expired_in, action)
+    raise NotFound("Can't find a location for the data")
+
+
+def get_file(action, file_id):
+    doc = json.loads(get_index_document(file_id))
```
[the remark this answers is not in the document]
thanh-nguyen-dang: `json.loads()` deserialize s (a str or unicode instance containing a JSON document) to a Python object using this conversion table.

We still need it.

philloooo:
merging - codacy having a false negative report

PR #54, opened 2018-01-23
feat(id-token): initial addition of ID token

Avantol13:
Trimmed unnecessary access/refresh token fields

Used bare bones implementation of IDToken from Authlib and patched extra stuff on top. Minimal testing, should be expanded further once all the OIDC stuff gets merged together

### DIFF
--- a/fence/jwt/token.py
+++ b/fence/jwt/token.py
@@ -23,7 +23,7 @@ class UnsignedIDToken(AuthlibIDToken):
     def __init__(self, token):
         super(UnsignedIDToken, self).__init__(token)
 
-    def get_signed_token(self, kid, private_key):
+    def get_signed_and_encoded_token(self, kid, private_key):
         """
         Return a signed ID token by using the private key and kid provided
 
@@ -32,7 +32,7 @@ class UnsignedIDToken(AuthlibIDToken):
             private_key (str): RSA private key to sign and encode the JWT with
 
         Returns:
-            str: encoded JWT ID token signed with ``private_key``
+            str: UTF-8 encoded JWT ID token signed with ``private_key``
         """
         headers = {'kid': kid}
         token = jwt.encode(self.token, private_key, headers=headers, algorithm='RS256')
@@ -44,7 +44,8 @@ class UnsignedIDToken(AuthlibIDToken):
         Validate that the token hasn't expired
 
         Args:
-            now (int): number of seconds from 1970-01-01T0:0:0Z as measured in UTC until the date/time
+            now (int): number of seconds from 1970-01-01T0:0:0Z as measured in
+                       UTC until the date/time
 
         Raises:
             IDTokenError: token has expired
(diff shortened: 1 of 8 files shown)


fence/jwt/token.py
```diff
@@ -30,28 +155,37 @@ def issued_and_expiration_times(seconds_to_expire):
     return (iat, exp)
 
 
-def generate_signed_refresh_token(kid, private_key, user, expires_in, scopes):
+def generate_signed_id_token(kid, private_key, user, expires_in, scopes, client_id,
```
rudyardrichter: Is `scopes` necessary?

tests/oauth2/test_id_token.py
```diff
@@ -0,0 +1,169 @@
...
+        max_age=max_age, nonce=nonce)
+
+    unsigned_token.validate(issuers=[issuer], client_id=client_id, max_age=max_age, nonce=nonce)
+
+    assert not unsigned_token.token.get("nonce")
+
+
+def test_expired_id_token(app):
+    """
+    Create a token and then validate it and make sure there are no exceptions
```
rudyardrichter: Looks like this docstring was copied and not updated

tests/oauth2/test_id_token.py
```diff
@@ -0,0 +1,169 @@
...
+    issuer = app.config.get('HOST_NAME')
+    keypair = app.keypairs[0]
+    scopes = []
+    client_id = "client_12345"
+    user = User(username='test', is_admin=False)
+    expires_in = 0
+    nonce = None
+    max_age = None
+
+    with pytest.raises(Exception):
```
rudyardrichter: Could this be more specific?

tests/oauth2/test_id_token.py
```diff
@@ -0,0 +1,169 @@
...
+    keypair = app.keypairs[0]
+    scopes = []
+    client_id = "client_12345"
+    user = User(username='test', is_admin=False)
+    expires_in = 2592000
+    nonce = None
+    max_age = 1
+    now = int(time.time()) - 10
+
+    with pytest.raises(Exception):
```
rudyardrichter: Also could make this more specific

fence/jwt/token.py
```diff
@@ -1,17 +1,142 @@
...
 
-
-from fence.jwt import blacklist, errors, keys
+from fence.jwt import blacklist
+from fence.jwt import errors
+from fence.jwt import keys
 from fence.jwt.blacklist import BlacklistedToken
 
 
+class UnsignedIDToken(AuthlibIDToken):
```
rudyardrichter: `UnsignedIDToken` can inherit specifically from `authlib.specs.oidc.CodeIDToken`, since we only support code flow

fence/jwt/token.py
```diff
@@ -1,17 +1,142 @@
...
+        """
+        # Patch authlib to actually check max_age against auth_time and handle
+        super(UnsignedIDToken, self).validate_auth_time(max_age)
+        if max_age:
+            age = int(time.time()) - self.auth_time
+            if max_age < age:
+                # FIXME: OP MUST attempt to actively re-authenticate the End-User
+                raise IDTokenError('too old. age since auth_time is greater than max_age')
+
+    def validate(self, client_id=None, issuers=None, max_age=None, nonce=None):
```
rudyardrichter: Looks like this is  from authlib, but I think we should take `issuer` (not plural), no need for this to be a list

fence/jwt/token.py
```diff
@@ -60,18 +194,61 @@ def generate_signed_refresh_token(kid, private_key, user, expires_in, scopes):
...
+    # Only include if provided, used to associate a client session with an ID
+    # token. If present in Auth Request from client, should set same val
+    # in ID token
+    if nonce:
+        claims['nonce'] = nonce
+
+    token = UnsignedIDToken(claims)
+
+    # validate before signing
+    token.validate(issuers=[issuer], client_id=client_id, max_age=max_age, nonce=nonce)
```
rudyardrichter: `issuers` -> `issuer`

fence/jwt/token.py
```diff
@@ -30,28 +155,37 @@ def issued_and_expiration_times(seconds_to_expire):
...
     Return:
-        str: encoded JWT refresh token signed with ``private_key``
+        str: encoded JWT ID token signed with ``private_key``
     """
-    headers = {'kid': kid}
     iat, exp = issued_and_expiration_times(expires_in)
+    issuer = flask.current_app.config.get('HOST_NAME')
+
+    # by default, just include the client_id
+    audiences = audiences or [client_id]
```
rudyardrichter: because audiences "`MUST contain the OAuth 2.0 client_id of the Relying Party as an audience value`", seems like a good idea to add a `if client_id not in audiences` and raise an exception if it isn't (or just add it), in case the provided value is missing it

fence/jwt/token.py
Avantol13: done in 16dc23d

fence/jwt/token.py
Avantol13: done in b5ff38f

fence/jwt/token.py
Avantol13: nope, removed in b5ff38f

fence/jwt/token.py
Avantol13: done in b5ff38f

tests/oauth2/test_id_token.py
Avantol13: yup, copy paste mistake. fixed in b5ff38f

tests/oauth2/test_id_token.py
Avantol13: It could, but the tests would then be coupled with authlib (since the error lives there). But I do get that it's very ambigious as to what kind of exception we're testing for.. and we're going to be pretty coupled to Authlib for a while so I changed it to the specific error authlib throws. b5ff38f

tests/oauth2/test_id_token.py
Avantol13: done in b5ff38f

fence/jwt/token.py
Avantol13: fixed in b5ff38f

PR #61, opened 2018-01-24
fix(list): fix list api add exp

thanh-nguyen-dang:
- add `exp` to return value of list and parameter to delete
- change `exp` from DateTime to BigInt for `user_refresh_token` table
r @philloooo @rudyardrichter

### DIFF
--- a/fence/blueprints/oauth2.py
+++ b/fence/blueprints/oauth2.py
@@ -175,9 +175,10 @@ def signed_refresh_token_generator(kid, private_key, **kwargs):
         Return:
             str: encoded JWT signed with ``private_key``
         """
-        return token.generate_signed_refresh_token(kid, private_key, get_user(request),
-                                                   request.expires_in, request.scopes,
-                                                   client_id=request.body.get('client_id'))
+        refresh_token, _ = token.generate_signed_refresh_token(kid, private_key, get_user(request),
+                                                                  request.expires_in, request.scopes,
+                                                                  client_id=request.body.get('client_id'))
+        return refresh_token
     return generate_signed_refresh_token_from_request
 
 
--- a/fence/blueprints/storage_creds.py
+++ b/fence/blueprints/storage_creds.py
@@ -1,7 +1,9 @@
+import flask
+import json
+
 from fence.auth import login_required
 from fence.data_model.models import GoogleServiceAccount
 from fence.data_model.models import GoogleProxyGroup
-import flask
 from flask import current_app as capp
 from flask import g, request, jsonify
 from ..resources.storage import get_endpoints_descriptions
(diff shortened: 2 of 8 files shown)


fence/blueprints/storage_creds.py
```diff
@@ -13,6 +15,7 @@
     revoke_refresh_token, create_access_token
 
 from fence.data_model.models import UserRefreshToken
+import pdb
```
rudyardrichter: leftover pdb

rudyardrichter: leftover pdb

fence/blueprints/oauth2.py
```diff
@@ -175,9 +175,10 @@ def generate_signed_refresh_token_from_request(request):
         Return:
             str: encoded JWT signed with ``private_key``
         """
-        return token.generate_signed_refresh_token(kid, private_key, get_user(request),
-                                                   request.expires_in, request.scopes,
-                                                   client_id=request.body.get('client_id'))
+        refresh_token, jti, _ = token.generate_signed_refresh_token(kid, private_key, get_user(request),
```
rudyardrichter: is the `jti` meant to be used for something? (like is there meant to be a db insert here maybe?)

fence/blueprints/storage_creds.py
```diff
@@ -93,8 +96,8 @@ def list_keypairs(provider):
         {
             "jtis":
             [
-               {"jti": "e9d58890-99b0-44f0-88bd-3ebc370b1329"},
-               {"jti": "e9d58890-99b0-44f0-88bd-3ebc370b132a"}
+               {"jti": "e9d58890-99b0-44f0-88bd-3ebc370b1329", "exp": "2018-02-23 08:15:10"},
```
rudyardrichter: [`UserRefreshToken.expires`](https://github.com/uc-cdis/fence/blob/fix/list/fence/data_model/models.py#L85) will be an integer, right? so the `exp` field here will be an int?

fence/blueprints/storage_creds.py
```diff
@@ -300,7 +302,11 @@ def delete_keypair(provider, access_key):
 
     '''
     if provider == 'cdis':
-        status = revoke_refresh_token(access_key)
+        params = json.loads(request.data)
+        exp = params.get('exp')
+        if exp is None:
+            abort(404, 'No expires time passed')
```
rudyardrichter: Seems like this should just be a 400

fence/blueprints/oauth2.py
thanh-nguyen-dang: it is used to include into return response

fence/jwt/token.py
```diff
@@ -68,7 +69,7 @@ def generate_signed_refresh_token(kid, private_key, user, expires_in,
     )
     token = jwt.encode(claims, private_key, headers=headers, algorithm='RS256')
     token = oauthlib.common.to_unicode(token, 'UTF-8')
-    return token
+    return token, jti, exp
```
rudyardrichter: Instead of returning the jti and the exp why not just return all the `claims`? Reduces the chance this will have to get changed again

PR #63, opened 2018-01-25
fix(list): fix to use request.data

thanh-nguyen-dang:
@rudyardrichter @philloooo

### DIFF
--- a/fence/blueprints/storage_creds.py
+++ b/fence/blueprints/storage_creds.py
@@ -16,6 +16,7 @@ from fence.resources.storage.cdis_jwt import create_refresh_token,\
     revoke_refresh_token, create_access_token
 
 from fence.data_model.models import UserRefreshToken
+from fence.errors import NotSupported
 
 blueprint = flask.Blueprint('credentials', __name__)
 
@@ -191,7 +192,10 @@ def create_keypairs(provider):
     '''
     client_id = getattr(g, 'client_id', None)
     if provider == 'cdis':
-        scopes = request.form.getlist('scopes')
+        if flask.request.headers['Content-Type'] == 'application/x-www-form-urlencoded':
+            scopes = request.form.getlist("scopes")
+        else:
+            scopes = json.loads(request.data)["scopes"]
         if not isinstance(scopes, list):
             scopes = scopes.split(',')
         token, jti = create_refresh_token(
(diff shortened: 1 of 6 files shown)


fence/blueprints/data.py
```diff
@@ -26,6 +26,7 @@ def get_index_document(file_id):
             capp.config['HOSTNAME'] + '/index')
     url = indexd_server + '/index/'
     try:
+        print(url + file_id)
```
rudyardrichter: leftover prints in this file

fence/blueprints/storage_creds.py
```diff
@@ -191,7 +192,7 @@ def create_keypairs(provider):
     '''
     client_id = getattr(g, 'client_id', None)
     if provider == 'cdis':
-        scopes = request.form.getlist('scopes')
+        scopes = json.loads(request.data)["scopes"]
```
rudyardrichter: Is there a reason to *not* support form data? Could we instead do
```python
if flask.request.headers['Content-Type'] == 'application/json':
    # use request.data
elif flask.request.headers['Content-Type'] == 'application/x-www-form-urlencoded':
    # use request.form
```

tests/test_cdis.py
```diff
@@ -1,5 +1,5 @@
 from .utils import cdis as utils
-
+import json
```
rudyardrichter: imports out of order

fence/resources/storage/cdis_jwt.py
```diff
@@ -2,7 +2,9 @@
 
 from flask import current_app as cur_app
 from ...jwt import token, errors, blacklist
-from flask import jsonify
+from fence.errors import Unauthorized
```
rudyardrichter: imports out of order

fence/resources/storage/cdis_jwt.py
```diff
@@ -24,12 +26,15 @@ def create_session_token(
...
 
 
-def create_access_token(user, keypair, refresh_token, expires_in, client_id):
+def create_access_token(keypair, refresh_token, expires_in, client_id):
     try:
         decoded_jwt = token.validate_refresh_token(refresh_token)
-        scopes = decoded_jwt['access_aud']
+        scopes = decoded_jwt["access_aud"]
+        username = decoded_jwt["context"]["user"]["name"]
+        user = current_session.query(
```
rudyardrichter: Should use `fence.auth.get_current_user` instead

PR #76, opened 2018-01-31
Feat/oidc tests auth

Avantol13:
MVP for auth endpoint

There are quite a few tests I wrote and then ended up skipping since they're for optional params that are not being handled right now.

Current behavior for endpoint if user not logged in: redirect to default login (added some cfg vars for setting what the default login is) and then back to authorize


tests/utils/oauth2.py
```diff
@@ -21,7 +21,7 @@ def path_for_authorize(params=None):
     return '/oauth2/authorize' + make_query_string(params)
 
 
-def post_authorize(client, oauth_client, data=None):
+def get_authorize(client, oauth_client, data=None, confirm=None):
```
rudyardrichter: `confirm=False`?

tests/utils/oauth2.py
```diff
@@ -36,12 +36,50 @@ def post_authorize(client, oauth_client, data=None):
         'client_id': oauth_client.client_id,
         'redirect_uri': oauth_client.url,
         'response_type': 'code',
-        'scope': 'openid',
+        'scope': 'openid user',
+        'state': fence.utils.random_str(10)
+    }
+    default_data.update(data)
+    data = default_data
+    if confirm is not None:
```
rudyardrichter: `data['confirm'] = 'yes' if confirm else 'no'` if you're feeling terse

tests/utils/oauth2.py
```diff
@@ -36,12 +36,50 @@ def post_authorize(client, oauth_client, data=None):
...
+        'client_id': oauth_client.client_id,
+        'redirect_uri': oauth_client.url,
+        'response_type': 'code',
+        'scope': 'openid user',
         'state': fence.utils.random_str(10),
-        'confirm': 'yes',
     }
     default_data.update(data)
     data = default_data
+    if confirm is not None:
```
rudyardrichter: same as above

fence/blueprints/oauth2.py
```diff
@@ -16,40 +16,196 @@
...
+        **kwargs: additional keyword arguments
+    """
+    need_authentication = False
+    try:
+        user = get_current_user()
+    except Unauthorized:
+        need_authentication = True
+
+    if need_authentication or not user:
+        login_url = (
```
rudyardrichter: There's actually a utility function in `authlib` that makes this nicer:
```python
from authlib.common.urls import add_params_to_uri

params = {flask.current_app.config.get('DEFAULT_LOGIN_URL_REDIRECT_PARAM'): flask.request.url}
login_url = add_params_to_uri(flask.current_app.config.get('DEFAULT_LOGIN_URL', params)
```

tests/utils/oauth2.py
rudyardrichter: not sure if there should be differences between `confirm=None` and `confirm=False` so leaving this alone for now

PR #77, opened 2018-01-31
Feat/tt 300 awg group endpoint

nullet:


fence/blueprints/admin.py
```diff
@@ -22,3 +23,9 @@ def update_user(username):
     if resource not in ['compute', 'storage']:
         raise UserError('Resource {} is invalid'.format(resource))
     return update_user_resource(username, resource)
+
+
+@blueprint.route('/groups/<id>', methods=['GET'])
+@login_required({'admin'})
+def get_stupid_group(id):
```
jbarno: Lol. Maybe simple is a better adjective (given its open source)

fence/resources/group/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+    return group
+
+def get_group_info(group, session):
+    info = {
+        'group_id': group.id,
+        'group_name': group.name,
+        'lead_id': group.lead_id
+    }
+    return jsonify(info)
+
```
jesuspguofc: I like that you separated that function, now phoung will be able to use it to get all groups.
I would leave the jsonify on the admin endpoint, tho, so we can reuse the parsing part.

Good job!

fence/resources/group/__init__.py
```diff
@@ -0,0 +1,22 @@
+from flask import current_app as capp
+from flask import jsonify, g
+
+from fence.data_model.models import ResearchGroup
+
+def find_group(group_id, session):
```
jesuspguofc: I know this is a pain but it would be awesome to have some doc strings. ;)

fence/blueprints/admin.py
```diff
@@ -22,3 +23,9 @@ def update_user(username):
...
+    return get_info_by_group_id(id)
```
jesuspguofc: As mentioned below, this is where I was thinking of the jsonifying part

fence/resources/group/__init__.py
phuongph: In function: `def get_group_info(group, session):` `session` is not used. I guess you can safely remove it.

PR #90, opened 2018-02-05
docs(readme): update info for OIDC and include some initial sequence …

Avantol13:
Could use some extended details per step in each diagram but I think this is a good start, let me know if I'm totally misunderstanding something


README.md
```diff
@@ -1,9 +1,20 @@
 # Fence
+> AuthN and AuthZ service used primarily in the Gen3 Data Commons Software Stack
 
 [![Build Status](https://travis-ci.org/uc-cdis/fence.svg?branch=master)](https://travis-ci.org/uc-cdis/fence)
 [![Codacy Quality Badge](https://api.codacy.com/project/badge/Grade/1cb2ec9cc64049488d140f44027c4422)](https://www.codacy.com/app/uc-cdis/fence?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=uc-cdis/fence&amp;utm_campaign=Badge_Grade)
 [![Codacy Coverage Badge](https://api.codacy.com/project/badge/Coverage/1cb2ec9cc64049488d140f44027c4422)](https://www.codacy.com/app/uc-cdis/fence?utm_source=github.com&utm_medium=referral&utm_content=uc-cdis/fence&utm_campaign=Badge_Coverage)
 
+A `fence` separates protected resources from the outside world and allows
```
rudyardrichter: glad some gen3-themed description was added 😄

docs/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+Steps:
+- Follow link for diagram you want to change
+- Update diagram
+- Save as PNG using button on sequencediagram.org
+- Get permanent URL to updated diagram by using button on sequencediagram.org
+- Update image in this folder to new PNG (keep same name)
+- Update this README with new updated URL for diagram
+
+## client_registration.png
```
rudyardrichter: I think these would be somewhat more readable as:
```markdown
- [`client_registration.png`](https://sequencediagram.org/index.html?……)
- [`openid_connect_flow.png`](https://sequencediagram.org/index.html?……)
```
etc.

Avantol13:
diagrams cleaned up in b14dbe9

docs/README.md
Avantol13: I had something like this to start out but in practice, when you want to paste a new link after updating, it was easier to have it on it's own line so you don't have to worry about putting a `)` back on. It's a minor thing, I agree it's more markdown-y to do it your way.

Anyway, if you insist, I will change it :sweat_smile:

PR #111, opened 2018-02-19
feat(user): add user_id, username to presigned_url

thanh-nguyen-dang:
r @philloooo @rudyardrichter 
this pr depend on the pr below
https://github.com/uc-cdis/cdis-python-utils/pull/24

### DIFF
--- a/fence/blueprints/data.py
+++ b/fence/blueprints/data.py
@@ -1,22 +1,15 @@
-from urlparse import urlparse
-
 import flask
-from flask import current_app
 import requests
-
+from urlparse import urlparse
 from fence.auth import login_required
-from fence.errors import (
-    UnavailableError,
-    NotFound,
-    Unauthorized,
-    NotSupported,
-    InternalError,
-)
+from cdispyutils.hmac4 import generate_aws_presigned_url
+from cdispyutils.config import get_value
+from ..errors import UnavailableError, NotFound, Unauthorized, NotSupported, InternalError
 
 ACTION_DICT = {
     "s3": {
-        "upload": "put_object",
-        "download": "get_object"
+        "upload": "PUT",
+        "download": "GET"
     },
     "http": {
         "upload": "put_object",
@@ -29,20 +22,20 @@ SUPPORTED_PROTOCOLS = ['s3', 'http']
 
 def get_index_document(file_id):
     indexd_server = (
-            current_app.config.get('INDEXD') or
-            current_app.config['BASE_URL'] + '/index')
+            flask.current_app.config.get('INDEXD') or
+            flask.current_app.config['BASE_URL'] + '/index')
     url = indexd_server + '/index/'
     try:
         res = requests.get(url + file_id)
     except Exception as e:
-        current_app.logger.error("failed to reach indexd at {0}: {1}".format(url + file_id, e))
+        flask.current_app.logger.error("failed to reach indexd at {0}: {1}".format(url + file_id, e))
         raise UnavailableError(
             "Fail to reach id service to find data location")
     if res.status_code == 200:
         try:
             json_response = res.json()
             if 'urls' not in json_response or 'metadata' not in json_response:
-                current_app.logger.error(
+                flask.current_app.logger.error(
                     'URLs and metadata are not included in response from indexd: {}'.format(url + file_id)
                 )
                 raise InternalError('URLs and metadata not found')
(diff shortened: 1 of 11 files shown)


fence/blueprints/data.py
```diff
@@ -91,26 +88,35 @@ def check_protocol(protocol, scheme):
...
-        credential_key = buckets[location.netloc]
-        url = current_app.boto.presigned_url(
-            location.netloc,
-            location.path.strip('/'),
-            expires,
-            aws_creds[credential_key],
-            ACTION_DICT[protocol][action],
-        )
+        credential_key = capp.config['S3_BUCKETS'][location.netloc]['cred']
+        region = capp.config['S3_BUCKETS'][location.netloc]['region']
```
philloooo: can we get the region from boto instead of needing devops to manually configure it? also good so that we know the creds work for getting the bucket.

fence/blueprints/data.py
```diff
@@ -1,22 +1,19 @@
...
+from urlparse import urlparse
 from fence.auth import login_required
-from fence.errors import (
-    UnavailableError,
-    NotFound,
-    Unauthorized,
-    NotSupported,
-    InternalError,
-)
+from flask import current_app as capp
```
philloooo: why repeated import?
also flask is already imported, just use flask.current_app and flask.jsonify?

fence/blueprints/data.py
thanh-nguyen-dang: we cannot get region of a bucket. This one does not work and return error
https://docs.aws.amazon.com/cli/latest/reference/s3api/get-bucket-location.html

fence/blueprints/data.py
```diff
@@ -91,26 +88,35 @@ def check_protocol(protocol, scheme):
...
-        url = current_app.boto.presigned_url(
-            location.netloc,
-            location.path.strip('/'),
-            expires,
-            aws_creds[credential_key],
-            ACTION_DICT[protocol][action],
-        )
+        credential_key = capp.config['S3_BUCKETS'][location.netloc]['cred']
+        region = capp.config['S3_BUCKETS'][location.netloc]['region']
+        config = capp.config['AWS_CREDENTIALS'][credential_key]
```
philloooo: do `capp.config['AWS_CREDENTIALS'].get(credential_key)` too and check if it's null and raise error

fence/blueprints/data.py
```diff
@@ -91,26 +88,35 @@ def check_protocol(protocol, scheme):
...
+        http_url = 'https://{}.s3.amazonaws.com/{}'.format(location.netloc, location.path.strip('/'))
+        if 'aws_access_key_id' not in config:
+            raise Unauthorized('credential is not configured correctly')
+        else:
+            aws_access_key_id = config['aws_access_key_id']
+            aws_secret_key = config['aws_secret_access_key']
```
philloooo: `config.get` for both and check if either one is null

fence/blueprints/data.py
philloooo: works for me ??

fence/oidc/endpoints.py
```diff
@@ -59,6 +59,13 @@ def create_revocation_response(self):
         Return:
             Tuple[int, dict, dict]: (status_code, body, headers)
         """
+        headers = [
```
philloooo: why do we need any change on oidc for this issue?

fence/local_settings.example.py
```diff
@@ -69,13 +69,16 @@
 os.environ["GOOGLE_IDENTITY_DOMAIN"] = ""
 os.environ["GOOGLE_CLOUD_IDENTITY_ADMIN_EMAIL"] = ""
 
+AWS = {
```
philloooo: what's this for??

fence/resources/aws/boto_manager.py
```diff
@@ -40,6 +40,15 @@ def presigned_url(self, bucket, key, expires, config, method='get_object'):
         )
         return url
 
+    def get_bucket_region(self, bucket, config):
+        if config.has_key('aws_access_key_id'):
+            self.s3_client = client('s3', **config)
+        response = self.s3_client.get_bucket_location(Bucket=bucket)
```
philloooo: need more error handling here, could fail if
1. fail to reach aws
2. our credential don't have access to the bucket

PR #125, opened 2018-02-22
feat(default-login): add endpoint for default login and provider information

### LINKED ISSUE #120: add default login endpoint
`/login` should provide high level information like:
```
{
"default_provider": {
    "name": "NIH Login",
    "url": "BASE_URL/login/shib"},
"providers": [
    {"name": "NIH Login",
    "url": "BASE_URL/login/shib"}
]
}
```
So that a Commons's portal can know what login options to support
blocks https://github.com/uc-cdis/data-portal/issues/228

rudyardrichter:
Resolves #120.


fence/local_settings.example.py
```diff
@@ -110,13 +110,22 @@
...
 ENABLED_IDENTITY_PROVIDERS = {
-    'fence',
-    'google',
-    'shibboleth',
+    # ID for which of the providers to default to.
+    'default': 'google',
+    # Information for identity providers.
+    'providers': {
+        'fence': {
+            'name': 'Fence Multi-Tenant OAuth',
```
philloooo: hmmm not sure if this name makes sense for the portal to be displayed...

philloooo: oh I guess it's just example settings, nvm

rudyardrichter: Ok—what would be a better name?

PR #141, opened 2018-03-01
fix(default-login): fix URL joining

rudyardrichter:
fix URLs in default `/login` endpoint; should not be using `urljoin`


rudyardrichter:
@zflamig my concern there is that if the first part _doesn't_ have a slash at the end that doesn't work either:
```python
>>> urljoin('https://fence.net/user', 'login/google')
'https://fence.net/login/google'
```

zflamig:
Cool, I prefer being explicit when what you want is simple and not relying on magic functions anyways.

PR #142, opened 2018-03-01
Feat/google proxy groups

Avantol13:
- this also contains changes for userdatamodel merge
- https://github.com/uc-cdis/fence/pull/129

- db changes and migrations added to support proxy groups
- new `fence-create` script arg `google-init` to create proxy groups for all users who don't have one
- tested migration against old fence database
- tested google-init script with GCP and Cloud Identity to ensure service account and google group creation


Avantol13:
very old, already done in other prs

PR #144, opened 2018-03-03
feat/remove-user: support deleting users using `fence-create` script

### LINKED ISSUE #110: support removing users using fence-create
support `fence-create delete --user <username>`

rudyardrichter:
Resolves #110.

Also, refactors `bin/fence-create` to move the actual functions into the fence module itself, which allows for testing.


rudyardrichter:
NOTE: the fence part of this PR should be ready, however, because the `User` model in userdatamodel is not set up to cascade delete for its foreign keys in `Group` (for example) this won't work yet:
```
sqlalchemy.exc.IntegrityError: (IntegrityError) update or delete on table "User" violates foreign key constraint "user_group_user_id_fkey" on table "user_group"
DETAIL:  Key (id)=(33) is still referenced from table "user_group".
 'DELETE FROM "User" WHERE "User".id = %(id)s' {'id': 33}
```

tests/scripting/test_fence-create.py
```diff
@@ -0,0 +1,23 @@
...
+from fence.scripting.fence_create import (
+    delete_users,
+)
+
+
+def test_delete_users(app, db_session, example_usernames, example_users):
+    """
+    Test the functionality of ``delete_users``.
+    """
+    for user in example_users:
```
philloooo: can you test user deletion with user that's assigned Access privilege to projects

rudyardrichter: Added in 03002e5; fixed tests bug in fa621ad

PR #147, opened 2018-03-05
fix(slash): strict slash = false

thanh-nguyen-dang:
r @rudyardrichter @philloooo 
enable both '/' and '' in the endpoint

### DIFF
--- a/fence/blueprints/login/__init__.py
+++ b/fence/blueprints/login/__init__.py
@@ -75,7 +75,6 @@ def make_login_blueprint(app):
                 'name': idps[idp_id]['name'],
                 'url': absolute_login_url(idp_id),
             }
-
         try:
             all_provider_info = [
                 provider_info(idp_id) for idp_id in idps.keys()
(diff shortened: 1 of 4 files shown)


fence/blueprints/login/__init__.py
```diff
@@ -53,7 +53,7 @@ def make_login_blueprint(app):
     blueprint = flask.Blueprint('login', __name__)
     blueprint_api = Api(blueprint)
 
-    @blueprint.route('', methods=['GET'])
+    @blueprint.route('/', methods=['GET'], strict_slashes=False)
```
rudyardrichter: The fence app should already have this:

https://github.com/uc-cdis/fence/blob/c993cf9ef206856fdf571b1353f45d22fad70ff7/fence/__init__.py#L129

So this change shouldn't be necessary?

run.py
```diff
@@ -13,6 +13,7 @@
         get_client)
     patcher.start()
 
+app_register_blueprints(app)
```
rudyardrichter: `app.url_map.strict_slashes = False` has to go before this, so `app_sessions` should be called first, and `app_register_blueprints` after that.

Issue #11, opened 2017-12-04
modules to setup project group and proxy group after dbgap syncing


philloooo:

Avantol13:
needed to modify database tables to support this. pulled userdatamodel tables into fence to easily support this issue. Aditionally, ran into bugs in cirrus that require patching it.

So far I've spent ~12 hrs on this and ~6 on the userdatamodel move

PR #152, opened 2018-03-06
feat(google-proxy-group): script and fixes for creating proxy groups …

Avantol13:
…for users

- Relies on changes in cirrus
https://github.com/uc-cdis/cirrus/pull/5
- Also relies on changes in userdatamodel
https://github.com/uc-cdis/userdatamodel/pull/35

### DIFF
--- a/bin/fence-create
+++ b/bin/fence-create
@@ -245,11 +245,10 @@ def google_init(db):
                 primary_service_account = response["primary_service_account"]
                 user.google_proxy_group_id = group["id"]
 
-                # Add user's primary service account to database with
-                # "client_id" as user.id
+                # Add user's primary service account to database
                 service_account = GoogleServiceAccount(
                     google_unique_id=primary_service_account["uniqueId"],
-                    client_id=user.id,
+                    client_id=None,
                     user_id=user.id,
                     email=primary_service_account["email"]
                 )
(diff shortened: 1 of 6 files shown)


bin/fence-create
```diff
@@ -225,6 +228,42 @@ def assign_group_to_user(s, user, group_name, group_data):
...
+            s.query(User).filter(User.google_proxy_group == None)
+        )
+
+        for user in users_without_proxy:
+            with GoogleCloudManager() as g_mgr:
+                response = g_mgr.create_proxy_group_for_user(
+                    user.id, user.username)
+
+                group = response["group"]
+                primary_service_account = response["primary_service_account"]
```
philloooo: what's this primary service account?

bin/fence-create
```diff
@@ -257,22 +296,51 @@ if __name__ == '__main__':
...
+                urls=args.urls, auto_approve=not args.external)
+        else:
+            delete_client_action(DB, args.client)
+
     elif args.action == 'sync-dbgap':
         sync_dbgap(args.__dict__['projects'])
+
+    elif args.action == 'google-init':
+        db = (
+            "postgresql://{username}:{password}@{host}/{db}"
```
philloooo: why is this different from FENCE_DB?

bin/fence-create
Avantol13: This is a service account that gets added to the group when its created and is like a "pet" service account for the user. If we want to act on behalf of the user themselves in Google context, we'll use this service account.

That being said, because we're not creating VM's or anything, this is actually not getting used... it could allow the user themselves to hit creds endpoints to get temp keys? idk if we want to do that though.

The idea was that in addition to client service accounts we'd have one for the user themselves but maybe we don't really need this?

bin/fence-create
Avantol13: ah, this was actually an artifact of when I was moving all the userdatamodel stuff into fence. I was running into issues with importing local_settings... I think it was a different issue though. I fixed this in 95b3ebd

bin/fence-create
philloooo: I see, it makes sense to allow users hitting that endpoint too to claim using dcf is not dependent on going through any cloud resources

Avantol13:
resolves #11

Issue #50, opened 2018-01-18
monitoring to remove expired secret key for google service account


philloooo:

Avantol13:
23 hrs (required lots of changes/bug fixes to cirrus, updated endpoints to use tokens to remove db hits)

Avantol13:
resolves #158

PR #158, opened 2018-03-12
feat(google): temporary key expiration handling

Avantol13:
- Script to manage expired google account keys
- refactoring in fence-create for style and Python3 compatibility
- ensure tokens have 'azp' (authorized party field) to determine the client we're talking to
    - I think this got lost at some point during the oidc merging
- make storage cred endpoints use information from token 
- fix tests to send token in header so storage creds endpoint tests pass

Relies on some fixes in cirrus for service account id validation:
https://github.com/uc-cdis/cirrus/pull/8

### DIFF
--- a/fence/blueprints/storage_creds.py
+++ b/fence/blueprints/storage_creds.py
@@ -4,7 +4,8 @@ from cirrus import GoogleCloudManager
 from flask_sqlalchemy_session import current_session
 import flask
 
-from fence.auth import login_required
+from fence.auth import require_auth_header
+from fence.auth import current_token
 from fence.errors import NotSupported, UserError
 from fence.jwt.blacklist import blacklist_token
 from fence.jwt.token import USER_ALLOWED_SCOPES
@@ -32,7 +33,7 @@ ALL_RESOURCES = {
 
 
 @blueprint.route('/', methods=['GET'])
-@login_required({'credentials'})
+@require_auth_header({'credentials'})
 def list_sources():
     """
     List different resources user can have credentials
@@ -59,7 +60,7 @@ def list_sources():
 
 
 @blueprint.route('/<provider>/', methods=['GET'])
-@login_required({'credentials'})
+@require_auth_header({'credentials'})
 def list_keypairs(provider):
     """
     List access keys for user
@@ -128,10 +129,13 @@ def list_keypairs(provider):
                     [{'jti': item.jti, 'exp': item.expires} for item in tokens]}
     elif provider == 'google':
         with GoogleCloudManager() as g_cloud_manager:
-            service_account = _get_google_service_account_for_client(g_cloud_manager)
+            client_id = current_token["azp"] or None
+            service_account = _get_google_service_account_for_client(
+                g_cloud_manager, client_id)
 
             if service_account:
-                keys = g_cloud_manager.get_service_account_keys_info(service_account.google_unique_id)
+                keys = g_cloud_manager.get_service_account_keys_info(
+                    service_account.google_unique_id)
                 result = {'access_keys': keys}
             else:
                 result = {'access_keys': []}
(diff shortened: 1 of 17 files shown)


fence/blueprints/storage_creds.py
```diff
@@ -421,31 +478,30 @@ def _create_google_service_account_for_client(g_cloud_manager):
         fence.models.GoogleServiceAccount: New service account
     """
     # create service account, add to db
-    proxy_group_id = (
+    # TODO eventually proxy group id should be in the token
```
philloooo: when will that be token?

fence/resources/storage/cdis_jwt.py
```diff
@@ -44,7 +44,7 @@ def create_api_key(user, keypair, expires_in, scopes, client_id):
     with flask.current_app.db.session as session:
         session.add(
             UserRefreshToken(
-                jti=claims['jti'], userid=user.id, expires=claims['exp']
```
philloooo: can we rename the arg to userid for both `create_api_key` and `generate_api_key` for clarity?

fence/blueprints/storage_creds.py
Avantol13: I should probably do it as a part of this task actually, that would make the most sense. I'll work on it

fence/resources/storage/cdis_jwt.py
Avantol13: yeah, fixed in 662f14d3b78ceec1303929c9266f543be4bb8387

fence/blueprints/storage_creds.py
Avantol13: added in 57f2d90, fixed tests in  d040fde

fence/jwt/token.py
```diff
@@ -364,8 +367,12 @@ def generate_signed_access_token(
                 'name': user.username,
                 'is_admin': user.is_admin,
                 'projects': dict(user.project_access),
+                'google': {
+                    'proxy_group': user.google_proxy_group_id,
+                }
             },
         },
+        'azp': client_id or ''
```
philloooo: should this be client_id or user_id? your google code is assuming that the azp should be user_id and then give user's service account's key?

fence/blueprints/storage_creds.py
```diff
@@ -116,12 +119,15 @@ def list_keypairs(provider):
         }
 
     """
+    client_id = current_token["azp"] or None
```
philloooo: this will raise exception for previous version of token, do `current_token.get("azp")`?

fence/blueprints/storage_creds.py
```diff
@@ -232,17 +249,28 @@ def create_keypairs(provider):
...
         api_key, claims = create_api_key(
-            flask.g.user, flask.current_app.keypairs[0], expires_in, scope,
+            user_id, flask.current_app.keypairs[0], expires_in, scope,
             client_id
         )
         return flask.jsonify(dict(key_id=claims['jti'], api_key=api_key))
     elif provider == 'google':
         with GoogleCloudManager() as g_cloud:
-            key = _get_google_access_key(g_cloud)
+            client_id = current_token["azp"] or None
```
philloooo: same as above

fence/blueprints/storage_creds.py
```diff
@@ -322,28 +352,49 @@ def delete_keypair(provider, access_key):
         blacklist_token(jti, api_key.expires)
     elif provider == 'google':
         with GoogleCloudManager() as g_cloud:
-            service_account = _get_google_service_account_for_client(g_cloud)
+            client_id = current_token["azp"] or None
```
philloooo: same as above

fence/jwt/token.py
Avantol13: leaving as just client_id, will change endpoints to check for user's service account if cliend_id isn't there

fence/blueprints/storage_creds.py
Avantol13: ah, yup, good point. Thanks for catching that. fixed this and other occurances

Avantol13:
resolves #50

PR #166, opened 2018-03-16
feat(expire): centralize all expiration time

thanh-nguyen-dang:
Add all TTL settings into local_setting
r @rudyardrichter @philloooo

### DIFF
--- a/fence/local_settings.example.py
+++ b/fence/local_settings.example.py
@@ -132,4 +132,14 @@ ENABLED_IDENTITY_PROVIDERS = {
 
 APP_NAME = ''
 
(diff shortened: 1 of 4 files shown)


fence/local_settings.example.py
```diff
@@ -131,5 +131,3 @@
 }
 
 APP_NAME = ''
-
-MAX_PRESIGNED_URL_TTL = 7200
```
philloooo: we still want to be able to customize this ttl for some commons, can you move local_settings import in settings below all the TTL definitions?

thanh-nguyen-dang: I moved it back

PR #173, opened 2018-03-21
refactor/db-calls: reduce DB usage

### LINKED ISSUE #66: PXD-391 ⁃ eliminate database calls
for endpoints that do not require to get latest user information, don't call database to get user. Just get needed info from jwt token

rudyardrichter:
Resolves #66.

Notable changes:
- Removes the original `get_current_user()` usage; instead, uses `current_user` proxy which will do at most 1 lookup for user per request and store the user info from the database as a dictionary
- Changes endpoints to use `require_auth` (instead of `login_required`) and use the `current_token` proxy where possible to get necessary information from the token instead of database
- Changes `MOCK_AUTH` to be an entire fake token instead of a boolean, to support "stateless" mocked auth

Changes left:
- Change storage manager setup to use token (place that only needs username)
- Appease codacy


fence/user.py
```diff
@@ -1,14 +1,78 @@
...
+    else:
+        mock_auth_token = flask.current_app.config.get('MOCK_AUTH')
+        if mock_auth_token:
+            username = mock_auth_token['context']['user']['name']
+    return username
+
+
+def _get_user(username):
+    """
+    Look up the user with the given username, and return a dictionary
```
philloooo: does this change the behavior of fence? before `login_user` will automatically create a new user if it doesn't exist before, so when you login to our commons the first time, it will work

rudyardrichter: That's a good point, will change this.

rudyardrichter: This was fixed in f3e74befd7f421f796e5ff28baf4222d25ea1715

PR #176, opened 2018-03-27
Feat/google linking

Avantol13:
- Allow linking of a user's personal Google account to their User identity 
- GET /link/google will run user through oauth with google to link their Google account
    - redirect and other needed information is saved in the session before the oauth flow with Google
    - after "linking", the endpoint will also add the Google account to their Proxy Group so it has access to the same things as the user. This addition is temporary, by default the expiration is 24 hours
    - Once authN and adding to proxy group is done, fence will redirect back to client (with errors if anything bad happened)
    - error handling is done similar to oauth2 where there exist query params in the url (`error` and `error_description`)
- PATCH /link/google will extend the expiration of a user's linked Google account to 24 hours from current time
- DELETE /link/google will remove the account linkage permanently

NOTE:
- needs https://github.com/uc-cdis/cirrus/pull/10

resolves fence#28


fence/blueprints/link.py
```diff
@@ -0,0 +1,376 @@
...
+    Linking a google account will add it to user's proxy group for a
+    configurable amount of time. During this time, the google account
+    will have permissions the same buckets the user does.
+
+    Extending access
+    """
+
+    @require_auth_header({'user'})
+    def get(self):
+        # TODO should this be allowed?
```
Avantol13: Should we allow GET /link/google to create linkage?

fence/blueprints/link.py
```diff
@@ -0,0 +1,376 @@
...
+            error = _get_error_params(
+                'g_acnt_link_error',
+                'User already has a linked Google account.')
+            flask.redirect_url = provided_redirect + error
+
+        return flask.redirect(flask.redirect_url)
+
+    @staticmethod
+    def _extend_account_expiration():
+        provided_redirect = flask.request.args.get('redirect')
```
Avantol13: On second thought, do we really need this redirect param? I think we can just return status with json response if errors?

fence/blueprints/link.py
Avantol13: Yes and don't allow POST. issues with redirection.

Avantol13: Thanks future Alex

fence/resources/user/user_session.py
```diff
@@ -109,8 +110,8 @@ def get(self, key, *args):
         """
         return self.session_token["context"].get(key, *args)
 
-    def pop(self, key):
-        return self.session_token['context'].pop(key)
+    def pop(self, key, *args, **kwargs):
```
Avantol13: this was due to an issue with providing a default pop value if I remember correctly

fence/blueprints/link.py
```diff
@@ -0,0 +1,482 @@
...
+                }
+                return error_message, 400
+
+            if g_account_access:
+                current_session.delete(g_account_access)
+                current_session.commit()
+
+            current_session.delete(g_account)
+            current_session.commit()
+        else:
```
rudyardrichter: I would maybe reverse the logic here and do this first, towards the top of the function: 
```python
if not g_account:
    error_message = {
    ...
```
and then the normal flow (currently in the `if` part) can go after that and be de-dented.

fence/blueprints/link.py
```diff
@@ -0,0 +1,482 @@
...
+        'Linking Google account {} to user with id {}.'.format(
+            google_email, user_id))
+    current_session.commit()
+    return user_google_account
+
+
+def _get_error_params(error, description):
+    params = ''
+    if error:
+        params += (
```
rudyardrichter: If this is meant to go in a URL then use `urllib`:
```python
args = {'error': error, 'error_description': description}
params = '?' + urllib.urlencode(params)
```

fence/blueprints/link.py
```diff
@@ -0,0 +1,482 @@
...
+    error = ''
+    error_description = ''
+
+    user_google_account = (
+        current_session.query(UserGoogleAccount)
+        .filter(UserGoogleAccount.email == google_email).first()
+    )
+
+    if not user_google_account:
+        if _already_authed is True:
```
rudyardrichter: `if _already_authed:`

fence/blueprints/link.py
```diff
@@ -0,0 +1,482 @@
...
+    NOTE: _already_authed=True means that AUTHENTICATION SHOULD HAVE
+         ALREADY OCCURED. That means that we've already verified that the
+         google_email belongs to the given user_id.
+
+    For just updating expiration, you can provide _already_authed=False, which
+    means that this will NOT allow the creation of a new UserGoogleAccount and
+    only update one that already exists.
+
+    Args:
+        user_id (TYPE): Description
```
rudyardrichter: Looks like some placeholders got left here

fence/blueprints/link.py
```diff
@@ -0,0 +1,482 @@
...
+    error_description = ''
+
+    user_google_account = (
+        current_session.query(UserGoogleAccount)
+        .filter(UserGoogleAccount.email == google_email).first()
+    )
+
+    if not user_google_account:
+        if _already_authed is True:
+            if user_id is None:
```
rudyardrichter: could also do `if not user_id` if other false-y values are not accepted

fence/blueprints/link.py
```diff
@@ -0,0 +1,482 @@
...
+    means that this will NOT allow the creation of a new UserGoogleAccount and
+    only update one that already exists.
+
+    Args:
+        user_id (TYPE): Description
+        google_email (TYPE): Description
+        proxy_group (TYPE): Description
+        _already_authed (bool, optional): Description
+    """
+    error = ''
```
rudyardrichter: Maybe it would make more sense to leave these as `None` in the event that there's no error?

fence/blueprints/link.py
```diff
@@ -0,0 +1,482 @@
...
+    WARNING: This assumes that provided arguments represent valid information.
+             This BLINDLY adds without verification. Do verification
+             before this.
+
+    Specifically, this ASSUMES that the proxy group provided belongs to the
+    given user and that the user has ALREADY authenticated to prove the
+    provided google_email is also their's.
+
+    Args:
+        user_id (TYPE): Description
```
rudyardrichter: Placeholders

fence/blueprints/link.py
```diff
@@ -0,0 +1,482 @@
...
+        NotFound: Description
+        Unauthorized: Description
+    """
+    user_google_account = (
+        current_session.query(UserGoogleAccount)
+        .filter(UserGoogleAccount.email == google_email).first()
+    )
+
+    if not user_google_account:
+        if _allow_new is True:
```
rudyardrichter: `if _allow_new:`

fence/blueprints/link.py
```diff
@@ -0,0 +1,482 @@
...
+            user_google_account.user_id,
+            proxy_group_id,
+            expiration)
+    )
+
+    current_session.commit()
+
+
+def get_default_google_account_expiration():
+    now = int(time.time())
```
rudyardrichter: :100:

fence/blueprints/link.py
```diff
@@ -0,0 +1,482 @@
...
+    Return user's linked google account's email.
+    """
+    google_email = get_users_linked_google_email_from_token()
+    if not google_email:
+        # hit db to check for google_email if it's not in token.
+        # this will catch cases where the linking happened during the life
+        # of an access token and the same access token is used here (e.g.
+        # account exists but a new token hasn't been generated with the linkage
+        # info yet)
+        google_email = get_users_linked_google_email_from_db(user_id)
```
rudyardrichter: It looks like this is the only place that `get_users_linked_google_email_from_db` is used; IMO it could just be inlined here, since it's not that long

fence/jwt/token.py
```diff
@@ -1,12 +1,12 @@
+import flask
 import json
+import jwt
```
rudyardrichter: nitpick: break up standard library (`json`, `time, `uuid`) and then 3rd-party imports (``flask`, `jwt`, with `authlib`)

fence/models.py
```diff
@@ -179,7 +179,7 @@ class GoogleServiceAccount(Base):
     # case we're ever juggling mult. projects)
     google_unique_id = Column(
         String,
-        unique=True,
+        unique=True,  # TODO this should be False...
```
rudyardrichter: reading this left me simultaneously very amused and confused

fence/models.py
```diff
@@ -204,12 +204,64 @@ class GoogleServiceAccount(Base):
             'google_service_accounts', cascade='all, delete-orphan')
     )
 
+    # TODO project_id
```
rudyardrichter: TODO?

fence/oidc/jwt_generator.py
```diff
@@ -7,7 +7,10 @@
     generate_signed_id_token,
     generate_signed_refresh_token,
 )
-from fence.models import AuthorizationCode, User
+from fence.models import AuthorizationCode
```
rudyardrichter: can put this and the next line on the same line

fence/resources/user/user_session.py
rudyardrichter: It seems like pop just has 2 arguments? ([docs here](https://docs.python.org/2/library/stdtypes.html#dict.pop)) so `*args, **kwargs` shouldn't be necessary

fence/utils.py
```diff
@@ -9,7 +9,9 @@
 import flask
 from userdatamodel.driver import SQLAlchemyDriver
 from werkzeug.datastructures import ImmutableMultiDict
+from flask_sqlalchemy_session import current_session
 
+from fence.models import UserGoogleAccount
```
rudyardrichter: can put this on same line as next

tests/conftest.py
```diff
@@ -235,6 +235,30 @@ def app(kid, rsa_private_key, rsa_public_key):
     mocker = Mocker()
     mocker.mock_functions()
     root_dir = os.path.dirname(os.path.realpath(__file__))
+    fence.app.config['OPENID_CONNECT'] = {
```
rudyardrichter: [this bit of config](https://github.com/uc-cdis/fence/blob/1854459e44b3d60e152b2cef8a239b54da310ba7/tests/test_settings.py#L108-L114) is in `test_settings.py` already, so it seemingly shouldn't be necessary? though I expect this got put here because of some problems related to that...

tests/link/test_link.py
```diff
@@ -0,0 +1,539 @@
...
+    Test the link endpoint for setting session details (this will be
+    needed by the return endpoint).
+    """
+    encoded_credentials_jwt = encoded_creds_jwt['jwt']
+    user_id = encoded_creds_jwt['user_id']
+    proxy_group_id = encoded_creds_jwt['proxy_group_id']
+
+    redirect = 'http://localhost'
+    r = client.get(
+        '/link/google?redirect=' + redirect,
```
rudyardrichter: I think you can just give the client a `query_string` argument?
```python
client.get(
    '/link/google',
    headers=...,
    query_string={'redirect': redirect},
)
```

rudyardrichter: (same idea in a few other places in the tests)

rudyardrichter:
For the codacy complaints—you could try changing the config to be exactly this:
```
exclude_paths:
  - 'tests/**'
```
and then it might stop complaining about the asserts. I am entirely to blame for the current broken state of the `.codacy.yml`, so my apologies...

fence/blueprints/link.py
Avantol13: Yeah I did this since if I don't check against the error before casting it to a str I'll get '' instead of 'None'. I guess it really doesn't matter much since I `if not error` the result and both are false-y. eh, I'll change it

fence/blueprints/link.py
Avantol13: whoops. yup, I'll fix those

fence/models.py
Avantol13: lol, I goofed up when I first created it and realized that here. need to fix it and this was a forced reminder since I didn't want to write the migration at the time. I'll go back and do that

fence/models.py
Avantol13: **sigh** yeah, I need to remember to add this for future Google stuff, not needed for this pr. I know I'm littering but I'll clean it up soon. Maybe I'll just add this when I fix the above db issue

fence/oidc/jwt_generator.py
Avantol13: ohh, I misread the style guide. https://github.com/uc-cdis/rfcs/blob/master/text/0003-style-guide.md#imports *Different* libraries on separate lines. My bad

Avantol13: I've been doing this everywhere... :sweat_smile:

tests/conftest.py
Avantol13: oh, I think this was a merge issue, it wasn't there when I first started this I think? idk, I'll remove it and make sure tests still work

tests/link/test_link.py
Avantol13: oh, nice! I didn't know about that

fence/blueprints/link.py
Avantol13: Yeah, there are a few functions here that should really live somewhere else in the long run (these will be used elsewhere when more google functionality gets added). I just didn't have a great place to put them for this pr :P

Avantol13:
addressed most comments in b6a4937 and 33c07fd . Am going to work on db changes for those TODO's

rudyardrichter:
(approving pending the db TODOs)

fence/models.py
```diff
@@ -411,11 +429,124 @@ def drop_foreign_key_constraint_if_exist(
...
+    if column_name in table.c:
+        with driver.session as session:
+            session.execute(
+                "ALTER TABLE \"{}\" ALTER COLUMN \"{}\" SET NOT NULL;"
+                .format(table_name, column_name)
+            )
+            session.commit()
+
+
+def _add_google_project_id(driver, md):
```
rudyardrichter: This could use a brief docstring to say what this does

Avantol13: :+1:

PR #180, opened 2018-03-30
chore(docs): update swagger for google endpoints

Avantol13:
- Add POST `/link/google`
- Add details to `/link/google`
- More restful interpretation of service account registration 
    - `/google/service_accounts` and `/google/service_account/{id}`
    - GET (list service accounts for user)
    - POST (new service account)
    - Then for specific service accounts: GET, PATCH (change dataset access), DELETE

P.S. I've been using https://editor.swagger.io/ to visualize

### DIFF
--- a/openapi/swagger.yaml
+++ b/openapi/swagger.yaml
@@ -743,6 +743,27 @@ paths:
         '400':
           description: failure, registering service account would result in errors
 
+  '/google/service_account/monitor':
+    get:
+      tags:
+        - google
+      summary:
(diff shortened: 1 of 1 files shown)


openapi/swagger.yaml
```diff
@@ -583,7 +585,7 @@ paths:
           description: Couldn't find key, unable to delete given key
 
   '/link/google':
-    get:
+    post:
```
philloooo: what's the difference between get and post?

PR #190, opened 2018-04-13
Feat/signed urls

Avantol13:
- Modify upload/download endpoints to generate signed urls for Google documents (e.g. gs://blahblahblah)
- Refactor upload/download and signed url endpoints to more clearly support different protocols in an OO model
- Modify previous test suite to run the same tests against s3 and gs locations
     - additionally, test when acl information is in index document as `acl` field AND when acl is in `metadata["acls"]` field

Details on implementation can be found here: https://docs.google.com/presentation/d/1lZjKcKqxsjXAceOzM-vj0rPs2dm3fT4V2EaApXjMb_M/edit?usp=sharing

Details on expiration times for various resources in fence here:
https://docs.google.com/spreadsheets/d/1fyOos1WFjddjPVVBA-i6wrCmUjAB4NipaKq4Re9rgCY/edit?usp=sharing

Depends on https://github.com/uc-cdis/cirrus/pull/11


fence/models.py
```diff
@@ -273,6 +268,30 @@ def delete(self):
...
+
+    service_account_id = Column(
+        Integer,
+        ForeignKey(GoogleServiceAccount.id),
+        nullable=False
+    )
+
+    expires = Column(BigInteger)
+
+    private_key = Column(JSONB)
```
philloooo: let's encrypt the private key

Avantol13: done in a5eabf2

fence/blueprints/data.py
```diff
@@ -1,49 +1,43 @@
...
+from fence.auth import set_current_token
+from fence.auth import validate_request
+from fence.auth import current_token
 from cdispyutils.hmac4 import generate_aws_presigned_url
 from cdispyutils.config import get_value
-from ..errors import UnavailableError, NotFound, Unauthorized, NotSupported, InternalError
+
+from fence.resources.google.utils import (
+    get_or_create_users_primary_google_service_account_key,
+    create_users_primary_google_service_account_key
```
philloooo: shouldn't `get_or_create_users_primary_google_service_account_key` be a superset of the other one and you don't need to use both?

fence/blueprints/data.py
```diff
@@ -1,49 +1,43 @@
...
 ACTION_DICT = {
-    "s3": {
-        "upload": "PUT",
-        "download": "GET"
-    }
+    's3': {
+        'upload': 'PUT',
+        'download': 'GET'
+    },
+    'http': {
```
philloooo: pretty should the http block is deadcode and should be removed.

fence/blueprints/data.py
```diff
@@ -66,97 +60,300 @@ def upload_file(file_id):
...
+        self.file_id = file_id
+        self.index_document = self._get_index_document()
+        self.metadata = self.index_document.get('metadata', {})
+        self.set_acls = self._get_acls()
+        self.indexed_file_locations = (
+            self._get_indexed_file_locations(
+                self.index_document.get('urls', []))
+        )
+
+        if check_public(self.set_acls):
```
philloooo: `self.public = check_public(self.set_acls)` ?

fence/blueprints/data.py
```diff
@@ -66,97 +60,300 @@ def upload_file(file_id):
...
+        except Exception as e:
+            flask.current_app.logger.error(
+                "failed to reach indexd at {0}: {1}"
+                .format(url + self.file_id, e))
+            raise UnavailableError(
+                "Fail to reach id service to find data location")
+        if res.status_code == 200:
+            try:
+                json_response = res.json()
+                if 'urls' not in json_response or 'metadata' not in json_response:
```
philloooo: why is metaata required?

fence/blueprints/data.py
Avantol13: there's an edge case: if the key already exists and its expiration is before the desired expiration of the signed url, we need to keep it but create another key that will last long enough. here's the code where that happens: https://github.com/uc-cdis/fence/blob/feat/signed_urls/fence/blueprints/data.py#L329-L349

fence/blueprints/data.py
```diff
@@ -66,97 +60,300 @@ def upload_file(file_id):
...
+
+        return self._get_signed_url(protocol, action, expires_in)
+
+    def _get_signed_url(self, protocol, action, expires_in):
+        signed_url = None
+
+        if protocol:
+            for file_location in self.indexed_file_locations:
+
+                # allow https in they specific http
```
philloooo: fix syntax for comment?

fence/blueprints/data.py
```diff
@@ -66,97 +60,300 @@ def upload_file(file_id):
...
+
+        if protocol == 's3':
+            return S3IndexedFileLocation(url)
+        elif protocol == 'gs':
+            return GoogleStorageIndexedFileLocation(url)
+        else:
+            return IndexedFileLocation(url)
+
+
+class IndexedFileLocation(object):
```
philloooo: let's docstring this to explain this is the parent class and catch all non-aws/gs cases

fence/blueprints/data.py
```diff
@@ -66,97 +60,300 @@ def upload_file(file_id):
...
+        super(GoogleStorageIndexedFileLocation, self).__init__(url)
 
-def get_file(action, file_id):
-    doc = get_index_document(file_id)
-    if 'acl' in doc:
-        set_acls = set(doc['acl'])
+    def get_signed_url(self, action, expires_in, public_data=False):
+        resource_path = self.parsed_url.path.strip('/')
+        network_location = (
+            self.parsed_url.netloc.strip('/') or 'storage.googleapis.com'
```
philloooo: won't `parsed_url.netloc.strip('/')` be the bucket name?

fence/resources/google/utils.py
```diff
@@ -0,0 +1,265 @@
...
+    # Note that client_id is None, which is how we store the user's SA
+    user_google_service_account = (
+        get_google_service_account_for_client(client_id=None, user_id=user_id)
+    )
+
+    if not user_google_service_account:
+        raise InternalError(
+            'User {} does not have a primary service account. '
+            'Unable to get primary service account key. This service account '
+            'should automatically get created when the user\'s proxy group '
```
philloooo: when does this happen?

Avantol13: it shouldn't... but figured this would be a more useful error if it does? idk

fence/blueprints/data.py
Avantol13: I agree, will remove

fence/blueprints/data.py
Avantol13: much better, will change it :)

fence/blueprints/data.py
Avantol13: idk, this code was written by @thanh-nguyen-dang I just refactored it here...

fence/resources/google/utils.py
```diff
@@ -0,0 +1,265 @@
...
+    """
+    key = {}
+    service_account = get_google_service_account_for_client(client_id, user_id)
+
+    if not service_account:
+        if client_id:
+            service_account = _create_google_service_account_for_client(
+                client_id, user_id, proxy_group_id)
+        else:
+            # error about requiring client id in azp field of token
```
philloooo: client_id is None for direct user request?

fence/blueprints/data.py
Avantol13: sorry for throwing him under the bus :P I'm just not 100% certain why the check is there

fence/blueprints/data.py
Avantol13: will do

fence/resources/google/utils.py
```diff
@@ -0,0 +1,265 @@
...
+            'Unable to get primary service account key. This service account '
+            'should automatically get created when the user\'s proxy group '
+            'was created.'.format(user_id)
+        )
+
+    return _create_users_primary_google_service_account_key(
+        user_id, proxy_group_id, expires)
+
+
+def create_google_access_key_for_client(client_id, user_id, proxy_group_id):
```
philloooo: question about naming covention for functions/methods:
I feel it's unnecessary to have `_for_client` in the function name? all the arguments means `for` you'd want to name it `create_google_access_key_for_user_client_proxy_group` to let the function name describe the whole purpose?
What do you think about just make them create_google_access_key, get_google_access_key etc...

fence/blueprints/data.py
Avantol13: ah, shoot. yeah, I think I was testing with incorrect paths like `gs://storage.cloud.google.com/bucket_name/file.txt` but it should just be `gs://bucket_name/file.txt`. I will fix this

fence/resources/google/utils.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+def add_custom_service_account_key_expiration(
+        key_id, service_account_id, expires):
+    """
+    Add db entry of user service account key and its custom expiration.
+    """
+    sa_key = GoogleServiceAccountKey(
+        key_id=key_id,
+        service_account_id=service_account_id,
```
philloooo: does the SA id have the user & oidc client id embeded so we can trace back to who this key belongs to?

fence/resources/google/utils.py
philloooo: even shorter I would prefer it to be:
```
from fence.resources import google
google.create_service_account(...)
```
...

fence/blueprints/data.py
```diff
@@ -66,97 +60,300 @@ def upload_file(file_id):
...
+                    raise InternalError('URLs and metadata not found')
+                return res.json()
+            except Exception as e:
+                flask.current_app.logger.error(
+                    'indexd response missing JSON field {}'
+                    .format(url + self.file_id))
+                raise InternalError('internal error from indexd: {}'.format(e))
+        elif res.status_code == 404:
+            flask.current_app.logger.error(
+                'indexd did not find find {}; {}'
```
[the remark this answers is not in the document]
Avantol13: In what way?

fence/blueprints/data.py
Avantol13: just remove the empty line above?

fence/resources/google/utils.py
philloooo: we should probably update the swagger doc to explain any dependencies on hitting this presigned url endpoint

fence/blueprints/data.py
philloooo: yeah, I think we should remove the check, it's probably when the acl is still in the metadata. Now all commons is using the acls as separate field and having a metadata field is not required for presigend url to work

fence/blueprints/storage_creds/google.py
```diff
@@ -93,15 +96,45 @@ def post(self):
...
+        """
+        Add the service account creds for the user into our db. Actual
+        Oauth Client SAs are handled separately. This function assigns
+        the same expiration to the user's generated key but the mechanism
+        for expiration uses our db instead of checking google directly.
+
+        See fence-create scripting functions for expiration logic.
+
+        The reason for this difference is due to the fact that fence itself
+        uses the user's primary service account for url signing (in addition
```
philloooo: the url signing part seems to be adding the record to db in `_create_users_primary_google_service_account_key` already, so this code is not used for that endpoint?

fence/resources/google/utils.py
Avantol13: Yeah, the functions are already in `google.utils`, so I'll remove `google` from the names and try to decrease the verbosity

fence/resources/google/utils.py
Avantol13: yeah, but this is a more nuanced case where the primary service account doesn't exist. the error here isn't very useful or correct. I think this is another situation where an internal error should be raised. Ideally, we'll never get here because the proxy group and primary sa are created at the same time. I'll update the error though

fence/resources/google/utils.py
Avantol13: SA emails are guaranteed to have at least the user_id. For user sa's, the email will also have as much of the user's username as will fit (google has restrictions on the SA emails). For client sa's, the email will also have as much of the client_id that will fit. Since we always have user_id we can trace to whatever else we need in fence db.

fence/blueprints/storage_creds/google.py
Avantol13: no, this is not for url signing it's for when a user wants to generate temporary credentials for themself

fence/blueprints/data.py
Avantol13: okay, I'll remove that check

fence/resources/google/utils.py
Avantol13: yeah, and all the google stuff (slides I've made and whatnot) should probably eventually make it's way to the docs folder in fence in some fashion.

Avantol13: all the google endpoints have some setup dependancies in regards to running the init scripts and monitoring scripts.. so it might be best to consolidate all the information in some setup guide instead of in the swagger doc. I did update the docs with "gs" support in there as an example though

philloooo: sounds good, that makes sense, I previously thought you need to hit some other endpoint to 'activate google' before you hit this

fence/settings.py
```diff
@@ -55,6 +55,11 @@
 #: The maximum session lifetime in seconds.
 SESSION_LIFETIME = 28800
 
+#: ``GOOGLE_SERVICE_ACCOUNT_KEY_FOR_URL_SIGNING_EXPIRES_IN: int``
+#: The number of seconds the user's Google service account key used for
+#: url signing will last before being expired/rotated
```
philloooo: can you comment on how many days this equals too :p

requirements.txt
```diff
@@ -34,5 +34,5 @@ pyyaml
 -e git+https://github.com/uc-cdis/userdatamodel.git@1.1.4#egg=userdatamodel
 -e git+https://github.com/uc-cdis/datamodelutils.git@0.3.0#egg=datamodelutils
 -e git+https://github.com/uc-cdis/cdis-python-utils.git@0.2.8#egg=cdispyutils
--e git+https://github.com/uc-cdis/cirrus.git@0.0.4#egg=cirrus-0.0.4
+-e git+https://github.com/uc-cdis/cirrus.git@081dc0d6afe063f415098789d14c9fe2404c8db3#egg=cirrus
```
philloooo: can you create a new tag? (should be minor version bump)

PR #200, opened 2018-04-30
feat(oidc-config): add endpoint returning OIDC configuration 

### LINKED ISSUE #87: /.well-known/openid-configuration endpoint
also have jwt key endpoint in the response

rudyardrichter:
Resolves #87.

### DIFF
--- a/fence/blueprints/well_known.py
+++ b/fence/blueprints/well_known.py
@@ -1,10 +1,16 @@
 """
 This blueprint defines the endpoints under ``.well-known/``, which includes:
-- TODO: OIDC provider configuration
+- OIDC provider configuration
 - JWK endpoint ``/jwks``
 """
 
 import flask
+
+from fence.jwt.token import (
+    SESSION_ALLOWED_SCOPES,
+    USER_ALLOWED_SCOPES,
+    CLIENT_ALLOWED_SCOPES,
+)
 
 
 blueprint = flask.Blueprint('.well-known', __name__)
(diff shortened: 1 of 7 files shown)


fence/blueprints/well_known.py
```diff
@@ -22,3 +28,92 @@ def jwks():
...
+    authorization_endpoint = path_to(flask.url_for('oauth2.authorize'))
+    token_endpoint = path_to(flask.url_for('oauth2.get_token'))
+    userinfo_endpoint = path_to(flask.url_for('user.user_info'))
+    registration_endpoint = None  # not yet supported
+
+    # List all the scopes allowed in OAuth2 requests.
+    scopes_supported = list(set(USER_ALLOWED_SCOPES + CLIENT_ALLOWED_SCOPES))
+
+    # List of all the claims which fence MAY set in the ID token.
+    claims_supported = [
```
Avantol13: is this only spec-defined claims? If not, what about `pur`?

fence/blueprints/well_known.py
```diff
@@ -22,3 +28,92 @@ def jwks():
...
+        'userinfo_endpoint': userinfo_endpoint,
+        'registration_endpoint': registration_endpoint,
+        'jwks_uri': jwks_uri,
+        'scopes_supported': scopes_supported,
+        'response_types_supported': ['openid', 'code', 'token'],
+        'response_modes_supported': [],
+        'grant_types_supported': ['authorization'],
+        'subject_types_supported': subject_types_supported,
+        'id_token_signing_alg_values_supported': ['RS256'],
+        'id_token_encryption_alg_values_supported': [],
```
Avantol13: does this signify None or that we didn't specify? Spec says these few fields are optional, but doesn't that mean we should just not include it in the response at all?

tests/test_settings.py
```diff
@@ -12,7 +12,7 @@
 
 
 # WARNING: the test database is cleared every run
-DB = 'postgresql://postgres:postgres@localhost:5432/fence_test_tmp'
+DB = 'postgresql://postgres@localhost:5432/fence_test_tmp'
 
 MOCK_AUTH = True
```
Avantol13: maybe add a unit test that hits this new endpoint and checks for required fields?

fence/blueprints/well_known.py
rudyardrichter: this one seems funny—the docs say these are claims that the provider "MAY" have values for, and does not need to be exhaustive. since `pur` is for internal use, I figured it wasn't necessary, but we could certainly include it.

fence/blueprints/well_known.py
rudyardrichter: The empty list means we don't support token encryption.

fence/blueprints/well_known.py
Avantol13: ahh, if it's optional to include it then perhaps you're right, we don't need to include it

fence/blueprints/well_known.py
Avantol13: gotcha, makes sense. thanks for clarifying

PR #201, opened 2018-05-01
feat(logout): clear session

philloooo:


fence/utils.py
```diff
@@ -156,3 +156,9 @@ def strip(s):
     if isinstance(s, str):
         return s.strip()
     return s
+
+def clear_cookies(response):
+
+    # Set all cookies to empty and expired.
```
rudyardrichter: Could just make this the docstring.

fence/auth.py
```diff
@@ -59,7 +60,11 @@ def logout(next_url=None):
     if flask.session.get('provider') == IdentityProvider.itrust:
         next_url = flask.current_app.config['ITRUST_GLOBAL_LOGOUT'] + next_url
     flask.session.clear()
-    return next_url
+    redirect_response = flask.make_response(
+        flask.redirect(next_url)
+    )
+    clear_cookies(redirect_response)
+    return redirect_response
```
rudyardrichter: I'm sort of surprised that this doesn't break any tests (like [here](https://github.com/uc-cdis/fence/blob/2326d9e338cf37bfc4796717476c5fe22a9555a0/tests/test_logout.py#L4-L7)), since `build_redirect_url` is now returning a `flask.Response` instead of a `str` right?

PR #203, opened 2018-05-01
refactor(kid): change fence keypair scheme

rudyardrichter:
This changes the way that keypairs must be stored, and the way that fence loads them in at initialization.

The new method (also documented in the READEME) is that keypairs should be stored in subdirectories of the `keys` directory, with the subdirectory named as the ISO 8601 timestamp of when the keypair is created, like this:
```
keys/
    2018-05-01T20:22:07Z/
        jwt_public_key.pem
        jwt_private_key.pem
```


rudyardrichter:
For context: the directly JWT-related changes are contained in the `Keypair` class and `fence/jwt/keys.py`, but the changes to the keypairs required refactoring the JWT generation stuff used for `fence-create`. Also required refactoring some tests accordingly.

bin/fence-create
```diff
@@ -82,7 +84,8 @@ def parse_arguments():
     )
 
     token_create = subparsers.add_parser('token-create')
-    token_create.add_argument('--kid')
+    token_create.add_argument('--kid', required=True)
+    token_create.add_argument('--prv-key-path', required=True)
```
frickjack: Why are these arguments required?
Fence-create should just pick the newest keys by default if a key is not specified.
Otherwise please patch:
     https://github.com/uc-cdis/gen3-qa/blob/master/run-tests.sh

tests/keys/2018-05-01T21:29:02Z/jwt_private_key.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
frickjack: Can we rig up the test with a 'setup' function that generates test keys rather than check in this file?  Somebody is bound to us it.

rudyardrichter: fence is already using the test key fixtures from authutils—I wanted to specifically test that the functionality of fence loading keypairs from a directory like this would work. If avoiding committing any keys is better, I can take out the tests.

bin/fence-create
rudyardrichter: Good point, will change

rudyardrichter: Fixed in 316eee2

PR #207, opened 2018-05-03
Feat/google buckets

Avantol13:
* Added db tables for Google bucket administration
* fence-create script for bucket creation 
* fence-create script for linking a bucket to a "Project"
     * for the case where the bucket already exists
* Updates and patches for other Google items discovered when deploying to qa env

Relies on https://github.com/uc-cdis/cirrus/pull/15

NOTE: https://github.com/uc-cdis/fence/issues/212 will handle giving users access after dbgap/user sync (e.g. adding users to the Google access groups that get created here)


fence/settings.py
```diff
@@ -23,7 +23,13 @@
         import imp
         imp.load_source('local_settings', '/var/www/local_settings.py')
```
Avantol13: @rudyardrichter Is there a reason this isn't already `var/www/fence/local_settings.py`? I was having issues since cloud automation puts it there by default

rudyardrichter: yep, this should be `/var/www/fence/local_settings.py`. my bad.

fence/scripting/fence_create.py
```diff
@@ -617,3 +631,230 @@ def create_user_access_token(DB, BASE_URL, ROOT_DIR, kid, username, scopes, expi
...
+                name=name,
+                storage_class=storage_class,
+                requester_pays=requester_pays,
+                google_project_id=google_project_id,
+                public=public,
+                project_auth_id=project_auth_id,
+                access_logs_bucket=access_logs_bucket)
+        )
+
+        if not public:
```
philloooo: let's create one read group and one write group?

Avantol13: added support for that in 7dbfa7f, now creates read and write groups by default

openapi/swagger.yaml
```diff
@@ -0,0 +1,1177 @@
+---
```
philloooo: Can you merge this into the master’s openapis/swagger.yaml an remove this file?

requirements.txt
```diff
@@ -34,5 +33,5 @@ pyyaml
 -e git+https://github.com/uc-cdis/userdatamodel.git@1.1.4#egg=userdatamodel
 -e git+https://github.com/uc-cdis/datamodelutils.git@0.3.0#egg=datamodelutils
 -e git+https://github.com/uc-cdis/cdis-python-utils.git@0.2.8#egg=cdispyutils
--e git+https://github.com/uc-cdis/cirrus.git@0.0.5#egg=cirrus-0.0.5
--e git+https://github.com/uc-cdis/authutils.git@2.0.1#egg=authutils-2.0.1
+-e git+https://github.com/uc-cdis/cirrus.git@0.0.6#egg=cirrus-0.0.6
+-e git+https://github.com/uc-cdis/authutils.git@2.0.1#egg=authutils-2.0.1authutils-2.0.1
```
philloooo: Duplicated ?

PR #209, opened 2018-05-08
Feat/awg integration

jesuspguofc:


fence/resources/admin/admin_token.py
```diff
@@ -0,0 +1,9 @@
+from fence.resources.user.user_session import create_long_access_token
+import fence.resources.user as us
+import jwt
+
+
+def get_long_lived_token(current_app, current_session,  token):
+    decoded_token = jwt.decode(token, verify=False)
```
rudyardrichter: seeing this line makes me really anxious! token validation should happen at the endpoint level, and then if you need to get information out of the token use `fence.auth.current_token`, which should be a dictionary of the already validated and decoded token (as long as the endpoint calling this function has required auth already). 

(maybe I'm missing a very good reason for doing it this way but in general, the above procedure is what should always happen)

fence/resources/admin/admin_token.py
```diff
@@ -0,0 +1,9 @@
...
+    user = us.get_user(current_session, decoded_token['context']['user']['name'])
+    return create_long_access_token(current_app, user)
```
rudyardrichter: this function should probably not exist; whatever needs to use this should be calling `create_long_access_token`, and should be called from an endpoint that already checked auth

consequently, remove this file

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,216 @@
+from fence.resources import (
```
rudyardrichter: imports out of order

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,216 @@
...
+from flask import current_app as capp
+from fence.models import User, Group
+from fence.errors import NotFound
+import json
+from fence.errors import UserError
+
+
+
+
+#### USERS ####
```
rudyardrichter: unnecessary

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,216 @@
...
+    response = udm.create_user_by_username_project(
+        current_session,
+        usr,
+        project
+    )
+    if response["result"] == "success":
+        proj = response["project"]
+        priv = response["privileges"]
+        cloud_providers = udm.get_cloud_providers_from_project(current_session, proj.id)
+        response = []
```
rudyardrichter: this is confusing that the function can either return `response: dict` if the result is not a success, or it can return a list if it is, it looks like?

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,216 @@
...
+    """
+    Create a user name for the specific project.
+    Returns a dictionary.
+    """
+    response = udm.create_user_by_username_project(
+        current_session,
+        usr,
+        project
+    )
+    if response["result"] == "success":
```
rudyardrichter: looks like it would be shorter to do it the opposite way: if result is not successful, return early; rest of the function should flow normally (and de-dented to the base level within the function)

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,216 @@
...
+                           " created for a bucket in the project {0}")
+                    response.append(msg.format(proj.name))
+                except:
+                    msg = ("Error user access not"
+                           " created for project {0} and bucket {2}")
+                    response.append(
+                        msg.format(proj.name, bucket["name"]))
+    return response
+
+def get_user_info(current_session, username):
```
rudyardrichter: re-defined function

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,216 @@
...
+    if new_name.upper() in user_list and not username.upper() == new_name.upper():
+        raise UserError(("Error: user with a name with the same combination/order "
+                         "of characters already exists. Please remove this other user"
+                         " or modify the new one. Contact us in case of doubt"))
+    usr.email = email or usr.email
+    if role:
+        is_admin = True if role == 'admin' else False
+    else:
+        is_admin = usr.is_admin
+    usr.is_admin = is_admin
```
rudyardrichter:
```python
if role:
    usr.is_admin = role == 'admin'
```
delete the `else` part and then this line

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,216 @@
...
+def delete_user(current_session, username):
+    """
+    Remove a user from both the userdatamodel
+    and the assciated storage for that project/bucket.
+    Returns a dictionary.
+    """
+    response = us.delete_user(current_session, username)
+    if response["result"] == "success":
+        # commented until we figure out where did the providers go
+        """
```
rudyardrichter: is this a TODO?

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,216 @@
...
+            connect_user_to_project(current_session, usr, project)
+        return responses
+
+
+def remove_user_from_groups(current_session, username, groups=[]):
+    usr = us.get_user(current_session, username)
+    user_groups = us.get_user_groups(current_session, username)['groups']
+    groups_to_keep =  [x for x in user_groups if x not in groups]
+    # thanks stackoverflow
+    projects_to_keep =  [ item for sublist in
```
rudyardrichter: awkward formatting; see previous comments

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,216 @@
...
+        responses.append(gp.connect_user_to_group(current_session, usr, grp))
+        projects = gp.get_group_projects(current_session, groupname)
+        projects_data = [ pj.get_project(current_session, project).auth_id for project in projects]
+        projects_list = [{"auth_id": auth_id, "privilege": ["read"]} for auth_id in projects_data]
+        for project in projects_list:
+            connect_user_to_project(current_session, usr, project)
+        return responses
+
+
+def remove_user_from_groups(current_session, username, groups=[]):
```
rudyardrichter: risky default argument

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,216 @@
...
+    usr = us.get_user(current_session, username)
+    user_groups = us.get_user_groups(current_session, username)['groups']
+    groups_to_keep =  [x for x in user_groups if x not in groups]
+    # thanks stackoverflow
+    projects_to_keep =  [ item for sublist in
+                          [ gp.get_group_projects(current_session, x) for x in groups_to_keep ]
+                          for item in sublist]
+    projects_to_keep = set(projects_to_keep)
```
rudyardrichter: can just use a set comprehension instead of converting:
```python
projects_to_keep = {
    ...
}
```

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,216 @@
...
+    projects_to_remove = [item for sublist in
+                          [ gp.get_group_projects(current_session, x) for x in groups]
+                          for item in sublist if item not in projects_to_keep]
+    projects_to_remove = set(projects_to_remove)
```
rudyardrichter: same here, use set comprehension

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,216 @@
...
+    for project in projects_to_remove:
+        remove_user_from_project(current_session, usr, project)
+    return {"result": responses}
+
+
+def disconnect_user_from_group(current_session, usr, groupname):
+    grp = gp.get_group(current_session, groupname)
+    if not grp:
+        return {"warning": ("Group {0} doesn't exist".format(group))}
+    else:
```
rudyardrichter: remove else, de-dent til end

fence/resources/group/__init__.py
```diff
@@ -0,0 +1,109 @@
+from fence.resources import userdatamodel as udm
+from fence.errors import UserError, NotFound
+
+def get_group(current_session, groupname):
```
rudyardrichter: re-defined function

in this case, just do
```python
from fences.resources.userdatamodel import get_group
```

fence/resources/group/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+    else:
+        projects = get_group_projects(current_session, groupname)
+        return {"name": group.name, "description": group.description, "projects": projects}
+
+
+def delete_group(current_session, groupname):
+    group = udm.get_group(current_session, groupname)
+    if not group:
+        raise UserError("Group doesn't exist")
+    else:
```
rudyardrichter: remove else & de-dent

fence/resources/project/__init__.py
```diff
@@ -0,0 +1,26 @@
+import fence.resources.userdatamodel as udm
```
rudyardrichter: this file only redefines functions; probably want to remove

fence/resources/provider/__init__.py
```diff
@@ -0,0 +1,10 @@
+from fence.resources import userdatamodel as udm
```
rudyardrichter: file is only redefinitions

fence/resources/user/__init__.py
```diff
@@ -30,25 +30,27 @@ def update_user_resource(username, resource):
                 server=flask.current_app.config['EMAIL_SERVER'])
         return get_user_info(user, session)
 
+def delete_user(current_session, username):
```
rudyardrichter: re-defined function

fence/resources/user/__init__.py
```diff
@@ -94,15 +98,21 @@ def send_mail(send_from, send_to, subject, text, server, certificates=None):
...
-    )
+    user = udm.get_user_accesses()
     if not user:
         raise InternalError(
             'Error: %s user does not exist'
             % flask.g.user.username
         )
     return user
+
+def get_user_groups(current_session, username):
```
rudyardrichter: re-defined function

fence/resources/user/user_session.py
```diff
@@ -261,3 +261,23 @@ def _create_access_token_cookie(app, session, response, user):
...
 
     return response
+
+
+
+def create_long_access_token(app, user):
+    keypair = app.keypairs[0]
+    scopes = SESSION_ALLOWED_SCOPES
+
+    now = datetime.now()
```
rudyardrichter: use `time.time()`, avoid `datetime`

fence/resources/user/user_session.py
```diff
@@ -261,3 +261,23 @@ def _create_access_token_cookie(app, session, response, user):
...
+    expiration = int(
```
rudyardrichter: the expiration can just get passed as `int` I think

rudyardrichter:
The review page glitched out so I "committed" it because I didn't want to lose anything, still reviewing

fence/resources/user/user_session.py
```diff
@@ -261,3 +261,23 @@ def _create_access_token_cookie(app, session, response, user):
     )
 
     return response
+
+
+
+def create_long_access_token(app, user):
+    keypair = app.keypairs[0]
+    scopes = SESSION_ALLOWED_SCOPES
```
rudyardrichter: this will always grant _all_ the scopes, including admin. maybe that is what you want. just a warning.

fence/resources/userdatamodel/__init__.py
```diff
@@ -5,127 +5,8 @@
...
-from fence.models import (
-    AccessPrivilege,
-    Bucket,
-    CloudProvider,
-    Project,
-    ProjectToBucket,
-    StorageAccess,
-    User,
-)
+from userdatamodel_project import *
```
rudyardrichter: I would definitely use explicit imports here

fence/resources/user/user_session.py
```diff
@@ -261,3 +261,23 @@ def _create_access_token_cookie(app, session, response, user):
     )
 
     return response
+
+
+
+def create_long_access_token(app, user):
```
rudyardrichter: does this function get used anywhere? I don't see it used anywhere but am suspicious of github collapsing large files on this review page

fence/resources/userdatamodel/userdatamodel_group.py
```diff
@@ -0,0 +1,87 @@
...
+        Group.name == groupname).first()
+    if group:
+        links = current_session.query(UserToGroup).filter(
+            UserToGroup.group_id == group.id)
+        for link in links:
+            current_session.delete(link)
+            current_session.flush()
+
+def get_group(current_session, groupname):
+    return current_session.query(Group).filter(
```
rudyardrichter: can use `filter_by(name=groupname)`, a bit shorter

I think this can just be one line?

fence/resources/userdatamodel/userdatamodel_group.py
```diff
@@ -0,0 +1,87 @@
...
+        Group.name == groupname).first()
+
+
+def get_user_group_access_privilege(current_session, user, group):
+    return current_session.query(UserToGroup).filter(
```
rudyardrichter: use `filter_by`, and only once
```python
return current_session.query(UserToGroup).filter_by(user_id=user.id, group_id=group.id).first()
```

for breaking long lines in this format, use this style:
```python
return (
    current_session
    .query(UserToGroup)
    .filter_by(user_id=user.id, group_id=group.id)
    .first()
)
```

fence/resources/userdatamodel/userdatamodel_group.py
```diff
@@ -0,0 +1,87 @@
...
+        UserToGroup.user_id == user.id).filter(
+            UserToGroup.group_id == group.id).first()
+
+def get_project_group_access_privilege(current_session, project, group):
```
rudyardrichter: same comments as previous

fence/resources/userdatamodel/userdatamodel_group.py
```diff
@@ -0,0 +1,87 @@
...
+    return current_session.query(AccessPrivilege).filter(
+        AccessPrivilege.project_id == project.id).filter(
+            AccessPrivilege.group_id == group.id).first()
+
+
+def get_all_groups(current_session):
```
rudyardrichter: seems not worth making this its own function

fence/resources/userdatamodel/userdatamodel_group.py
```diff
@@ -0,0 +1,87 @@
...
+    projects = []
+    for project in projects_to_group:
+        new_project = current_session.query(Project).filter(
+            Project.id == project.project_id).first()
+        if new_project:
+            projects.append(new_project.name)
+    return projects
+
+
+def get_empty_group():
```
rudyardrichter: this and the next two seem like they definitely don't need to be their own functions

fence/resources/userdatamodel/userdatamodel_project.py
```diff
@@ -0,0 +1,282 @@
...
+)
+
+from fence.errors import (
+    NotFound,
+    UserError,
+)
+
+
+
+def get_project(current_session, projectname):
```
rudyardrichter: use `filter_by`, should fit one line

fence/resources/userdatamodel/userdatamodel_project.py
```diff
@@ -0,0 +1,282 @@
...
+            Bucket).filter(Bucket.id == row.bucket_id).first()
+        info['associated buckets'].append(bucket.name)
+    return info
+
+
+def get_all_projects(current_session):
+    projects = current_session.query(Project).all()
+    projects_info = []
+    for project in projects:
+        projects_info.append(get_project_info(current_session, project.name))
```
rudyardrichter: use list comprehension

fence/resources/userdatamodel/userdatamodel_project.py
```diff
@@ -0,0 +1,282 @@
...
+    Delete the project from the database
+    The project should have no buckets in use
+    """
+    proj = current_session.query(
+        Project).filter(Project.name == project_name).first()
+    if proj:
+        buckets = current_session.query(
+            ProjectToBucket).filter(
+                ProjectToBucket.project_id == proj.id).first()
+        if not buckets:
```
rudyardrichter: i would reverse this: check if there are buckets and return early, otherwise just proceed as normal (and de-dent)

same with `if proj:`, do `if not proj` first

fence/resources/userdatamodel/userdatamodel_project.py
```diff
@@ -0,0 +1,282 @@
...
+        buck = current_session.query(Bucket).filter(
+            Bucket.id == bucket.bucket_id,
+            Bucket.provider_id == provider_id).first()
+        if buck:
+            response['buckets'].append(buck)
+    return response
+
+
+def get_user_project_access_privilege(current_session, user, project):
+    return current_session.query(
```
rudyardrichter: use a single `filter_by`, see above comments for formatting

fence/resources/userdatamodel/userdatamodel_provider.py
```diff
@@ -0,0 +1,89 @@
...
+    ongoing relationships
+    """
+    provider = current_session.query(
+        CloudProvider).filter(CloudProvider.name == provider_name).first()
+    if provider:
+        projects = current_session.query(
+            StorageAccess).filter(
+                StorageAccess.provider_id == provider.id).first()
+        if projects:
+            msg = "".join(
```
rudyardrichter: prefer something like this:
```python
msg = (
    "provider name {provider} in use in projects."
    " Please ..."
)
msg = msg.format(provider=provider_name)
```

fence/resources/userdatamodel/userdatamodel_provider.py
```diff
@@ -0,0 +1,89 @@
...
+                ["provider name ",
+                 provider_name,
+                 (" in use in projects."
+                  " Please remove these references and retry")])
+            raise UserError(msg)
+        else:
```
rudyardrichter: again with the return early from errors style here

fence/resources/userdatamodel/userdatamodel_user.py
```diff
@@ -0,0 +1,122 @@
...
+
+from fence.errors import (
+    NotFound,
+    UserError,
+)
+
+from flask_sqlalchemy_session import current_session as curr_sess
+
+
+def get_user(current_session, username):
```
rudyardrichter: strikes me as unnecessary

fence/resources/userdatamodel/userdatamodel_user.py
```diff
@@ -0,0 +1,122 @@
...
+    
+
+
+def delete_user(current_session, username):
+    """
+    Delete the user with the given username
+    """
+    user = current_session.query(
+        User).filter(User.username == username).first()
+    if user:
```
rudyardrichter: check error first, de-dent normal logic

fence/resources/userdatamodel/userdatamodel_user.py
```diff
@@ -0,0 +1,122 @@
...
+        msg = "".join(["user name ", username, " not found"])
+        raise NotFound(msg)
+
+def create_user_by_username_project(current_session, new_user, proj):
+    """
+    Create a user for a specific project
+    """
+    project = current_session.query(Project).filter(
+        Project.auth_id == proj['auth_id']).first()
+    if not project:
```
rudyardrichter: good example here for the error checking

fence/resources/userdatamodel/userdatamodel_user.py
```diff
@@ -0,0 +1,122 @@
...
+        current_session.add(priv)
+        current_session.flush()
+
+    return {"user": new_user,
+            "result": "success",
+            "project": project,
+            "privileges": priv}
+
+
+def get_all_users(current_session):
```
rudyardrichter: don't need a whole function for this probably

tests/admin/test_admin_groups.py
```diff
@@ -0,0 +1,153 @@
...
+    user = db_session.query(User).filter(User.username == 'awg_user').first()
+    user_projects = [ db_session.query(Project).filter(Project.id == item.project_id).first().name 
+                      for item in db_session.query(AccessPrivilege).filter(AccessPrivilege.user_id == user.id).all()
+                      if item.project_id != None ]
+    expected_projects = ['test_project_1', 'test_project_2']
+    expected_projects.sort()
+    user_projects.sort()
+    assert expected_projects == user_projects
+    adm.add_projects_to_group(db_session, 'test_group_1', ['test_project_6', 'test_project_5'])
+    user_projects = [ db_session.query(Project).filter(Project.id == item.project_id).first().name
```
rudyardrichter: fix list comprehension formatting

fence/resources/admin/admin_token.py
jesuspguofc: yeah no worries, that entire file should be gone hehehe sorry, it was from before Phillis told me about the api key thing. I removed the endpoint but I forgot all the trialing crap that I had around.

fence/blueprints/admin.py
```diff
@@ -1,24 +1,478 @@
...
+    role = request.get_json().get('role',None)
+    email = request.get_json().get('email',None)
+    current_app.logger.debug(
+        ("update_user:\n\tname: {0}\n\t"
+        "role: {1}\n\temail: {2}").format(
+            username, role, email))        
+    response = jsonify(adm.update_user(current_session, username, role, email, name))
+    return response
+
+@blueprint.route('/user/<username>', methods=['DELETE'])
```
rudyardrichter: If time allows, these endpoints should be documented in `fence/openapis/swagger.yaml`

fence/resources/userdatamodel/userdatamodel_project.py
```diff
@@ -0,0 +1,282 @@
...
+    else:
+        return {"result": "error, project not found"}
+
+
+def delete_bucket_on_project(current_session, project_name, bucket_name):
+    """
+    Remove a bucket and its relationship to a project
+    """
+    bucket = current_session.query(
+        Bucket).filter(Bucket.name == bucket_name).first()
```
rudyardrichter: could use `filter_by` here and maybe one line (same for a few places after here)

tests/admin/test_admin_groups.py
```diff
@@ -0,0 +1,153 @@
...
+    assert expected_projects == user_projects
+
+    adm.delete_group(db_session, 'test_group_2')
+    user_projects = [ db_session.query(Project).filter(Project.id == item.project_id).first().name 
+                      for item in db_session.query(AccessPrivilege).filter(AccessPrivilege.user_id == user.id).all()
+                      if item.project_id != None ]
+    expected_projects = ['test_project_1']
+    expected_projects.sort()
+    user_projects.sort()
+    assert expected_projects == user_projects
```
rudyardrichter: instead of sorting, probably want to convert to sets and then check equality on those

tests/admin/test_admin_groups.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def test_get_group_projects(db_session, awg_groups):
+    group_projects = adm.get_group_projects(db_session, 'test_group_4')
+    group_projects.sort()
+    expected_projects = ['test_project_7', 'test_project_6']
+    expected_projects.sort()
+    assert expected_projects == group_projects
```
rudyardrichter: same here, check sets equal

tests/admin/test_admin_groups.py
```diff
@@ -0,0 +1,153 @@
...
+
+def test_getprojects_from_inexistent_group(db_session, awg_groups):
+    with pytest.raises(NotFound):
+        group_projects = adm.get_group_projects(db_session, 'test_group_XXX')
+
+
+def test_remove_project_to_group(db_session, awg_groups):
+    group = db_session.query(Group).filter(Group.name == 'test_group_4').first()
+    group_projects = [ db_session.query(Project).filter(Project.id == item.project_id).first().name 
+                       for item in db_session.query(AccessPrivilege).filter(AccessPrivilege.group_id == group.id).all()]
```
rudyardrichter: shorter with `filter_by`

also formatting, change to be same as other list comprehensions

tests/admin/test_admin_groups.py
```diff
@@ -0,0 +1,153 @@
...
+
+    group_projects = [ db_session.query(Project).filter(Project.id == item.project_id).first().name 
+                       for item in db_session.query(AccessPrivilege).filter(AccessPrivilege.group_id == group.id).all()]
+    expected_projects = []
+    assert expected_projects == group_projects
+
+
+def test_remove_project_to_group_updates_user(db_session, awg_users, awg_groups):
+    user = db_session.query(User).filter(User.username == 'awg_user').first()
+    user_projects = [ db_session.query(Project).filter(Project.id == item.project_id).first().name
```
rudyardrichter: list comprehension formatting (also below here)

fence/__init__.py
```diff
@@ -106,9 +106,11 @@ def root():
     @app.route('/logout')
     def logout_endpoint():
         root = app.config.get('APPLICATION_ROOT', '')
-        next_url = build_redirect_url(app.config.get('ROOT_URL', ''), flask.request.args.get('next', root))
+        request_next = flask.request.args.get('next', root)
+        next_url = request_next if request_next.startswith('https') or request_next.startswith('http') else build_redirect_url(app.config.get('ROOT_URL', ''), request_next)
```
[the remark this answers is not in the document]
jesuspguofc: Done

fence/blueprints/admin.py
```diff
@@ -1,24 +1,478 @@
...
+
+#### CLOUD PROVIDER ####
+
+
+@blueprint.route('/cloud_provider/<providername>', methods=['GET'])
+@login_required({'admin'})
+@admin_required
+def get_cloud_provider(providername):
+    """
+    Retriev the information related to a cloud provider
```
jbarno: comment spelling

fence/resources/admin/admin_groups.py
```diff
@@ -0,0 +1,164 @@
...
+    Deletes a group
+    """
+    projects_to_purge =  gp.get_group_projects(current_session,
+                                              groupname)
+    remove_projects_from_group(current_session, groupname,
+                               projects_to_purge)
+    gp.clear_users_in_group(current_session, groupname)
+    gp.clear_projects_in_group(current_session, groupname)
+    gp.delete_group(current_session, groupname)
+    return {'result': 'success'}
```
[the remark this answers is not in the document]
jbarno: Is this all handled within a single transaction? Feels like if one of these didn't succeed we could get into a bad state quickly.

fence/resources/admin/admin_groups.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def update_group(current_session, groupname, description, new_name):
+    gp.update_group(current_session, groupname, description, new_name)
+    name = new_name or groupname
+    return gp.get_group_info(current_session, name)
+
+
+def get_group_info(current_session, groupname):
```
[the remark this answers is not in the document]
jbarno: Agreed I think if the blueprints were split into groups/cloud_providers/users under admin I think this wouldn't be here.

fence/auth.py
```diff
@@ -55,18 +55,20 @@ def logout(next_url=None):
...
     flask.session.clear()
     redirect_response = flask.make_response(
-        flask.redirect(next_url)
+        flask.redirect(itrust_next_url or next_url)
     )
     clear_cookies(redirect_response)
     return redirect_response
 
 
+
```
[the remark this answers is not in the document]
rudyardrichter: I have seen too many wonky merge strategies even _without_ those spaces...

fence/resources/admin/admin_groups.py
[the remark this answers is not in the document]
rudyardrichter: Right, I wondered about that too, which just requires being careful with committing changes to the DB; here at least I was thinking like having a `GroupUpdateResult` class (or whatever) that can be successful or unsuccessful and translate itself into JSON. Though, you could even have that be the level at which you finalize your changes, after everything has worked correctly

fence/blueprints/admin.py
```diff
@@ -1,24 +1,478 @@
...
+
+from fence.user import get_current_user
+from fence.resources import admin as adm
+from flask_sqlalchemy_session import current_session
+import jwt
+
+blueprint = Blueprint('admin', __name__)
+
+
+#### USERS ####
```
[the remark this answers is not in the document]
rudyardrichter: yeah, usually when I see these sort of top-level "section" comments my instinct is "break into multiple files"

fence/resources/userdatamodel/userdatamodel_project.py
```diff
@@ -0,0 +1,282 @@
...
+    new_project = Project(name=name, auth_id=auth_id)
+    current_session.add(new_project)
+    current_session.flush()
+    for storage in storage_accesses:
+        provider = current_session.query(
+            CloudProvider).filter(CloudProvider.name == storage).first()
+        if provider:
+            new_storage_access = StorageAccess(
+                provider_id=provider.id, project_id=new_project.id)
+            current_session.add(new_storage_access)
```
jbarno: Do you need to flush the storage_access's too?

fence/resources/userdatamodel/userdatamodel_project.py
```diff
@@ -0,0 +1,282 @@
...
+                       Bucket.provider_id == provider.id).first()
+    if not bucket:
+        bucket = Bucket(name=bucket_name, provider_id=provider.id)
+        current_session.add(bucket)
+        current_session.flush()
+        proj_to_bucket = ProjectToBucket(
+            project_id=project.id, bucket_id=bucket.id, privilege=['owner'])
+        current_session.add(proj_to_bucket)
+        # Find the users that need to be deleted
+        users_in_project = current_session.query(AccessPrivilege).filter(
```
jbarno: The users don't need to be deleted, they need to be added to this acl for this bucket though right? I think I saw a function above for getting all users on a project.

( I may not be understanding this)

fence/resources/userdatamodel/userdatamodel_project.py
```diff
@@ -0,0 +1,282 @@
...
+                                CloudProvider.id == storage.provider_id).first()
+                        usr = current_session.query(
+                            User).filter(User.id == access.user_id).first()
+                        users_to_remove.append((provider,usr))
+                        current_session.delete(usr)
+                current_session.delete(access)
+            for storage in storage_access:
+                current_session.delete(storage)
+            current_session.delete(proj)
+            return {"result": "success", "users_to_remove": users_to_remove}
```
jbarno: Don't you do the user deletion in line 176?

fence/resources/userdatamodel/userdatamodel_project.py
```diff
@@ -0,0 +1,282 @@
...
+        msg = "".join(["Bucket name ", bucket_name, " not found"])
+        raise NotFound(msg)
+    provider = current_session.query(
+        CloudProvider).filter(
+            CloudProvider.id == bucket.provider_id).first()
+    project = current_session.query(
+        Project).filter(Project.name == project_name).first()
+    if not project:
+        msg = "".join(["Project name ", project_name, " not found"])
+        raise NotFound(msg)
```
jbarno: These getters and checks look like they could live in their own "private" functions and be used a few times in this file.

fence/resources/userdatamodel/userdatamodel_project.py
```diff
@@ -0,0 +1,282 @@
...
+    project = current_session.query(
+        Project).filter(Project.name == project_name).first()
+    if not project:
+        msg = ''.join(["Project name ", project_name, " not found"])
+        raise NotFound(msg)
+    buckets = current_session.query(
+        ProjectToBucket).filter(ProjectToBucket.project_id == project.id)
+    response = {"buckets": []}
+    for bucket in buckets:
+        buck = current_session.query(
```
jbarno: is `buck` a copy of `bucket` here?

fence/resources/userdatamodel/userdatamodel_project.py
```diff
@@ -0,0 +1,282 @@
...
+
+def get_buckets_by_project_cloud_provider(current_session, project_id, provider_id):
+    """
+    List all the buckets assigned to a project
+    """
+    buckets = current_session.query(ProjectToBucket).filter(
+        ProjectToBucket.project_id == project_id)
+    response = {"buckets": []}
+    for bucket in buckets:
+        buck = current_session.query(Bucket).filter(
```
jbarno: Couldn't the provider_id be queried in the original query that gets you buckets?

fence/resources/userdatamodel/userdatamodel_user.py
```diff
@@ -0,0 +1,122 @@
...
+    user = current_session.query(
+        User).filter(User.username == username).first()
+    if user:
+        accesses = current_session.query(
+            AccessPrivilege).filter(AccessPrivilege.user_id == user.id)
+        groups = current_session.query(
+                UserToGroup).filter(
+                    UserToGroup.user_id == user.id).all()
+        for row in groups:
+            current_session.delete(row)
```
jbarno: Wait :bangbang:  this is deleting groups isn't it?

tests/conftest.py
```diff
@@ -288,7 +288,7 @@ def protected_endpoint(methods=['GET']):
 @pytest.fixture(scope='function')
 @mock_s3
 @mock_sts
-def user_client(app, request, db_session):
+def user_client(db_session):
```
jbarno: reading a file per test feels heavy handed.

jbarno: (I know you didn't change it :) )

tests/conftest.py
```diff
@@ -297,16 +297,30 @@ def user_client(app, request, db_session):
...
 
+@pytest.fixture(scope='function')
+def awg_users(db_session):
+    awg_usr = dict(json.loads(
+        utils.read_file('resources/awg_user.json')
+    ))
+    user_id, username = utils.create_awg_user(awg_usr, db_session)
+
+@pytest.fixture(scope='function')
+def awg_groups(db_session):
```
jbarno: splitting these two fixtures into 4 could help. Two session scoped that loads the dict (so you're only loading from a file once per test run). Then the other two take the db_session and the above session-scoped fixture to add users/ groups. 

Files are reads are "expensive", tests should be fast.

tests/utils/__init__.py
```diff
@@ -50,8 +50,116 @@ def create_user(users, db_session, is_admin=False):
                 s.add(ap)
             else:
                 ap.privilege = privilege
+
+    return user.id, user.username
+
+
+def create_awg_user(users, db_session):
```
jbarno: These feel like fixtures.

tests/conftest.py
jesuspguofc: We can review it offline (so I can understand the ramifications/impact of such action) and shoot an improvement ticket for this.

fence/auth.py
```diff
@@ -181,3 +184,16 @@ def get_user_from_claims(claims):
...
+def admin_required(f):
+    """
+    Require user to be an admin user. 
+    """
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        if not flask.g.user:
+            raise Unauthorized("Require login")
+        if flask.g.user.is_admin is not True:
```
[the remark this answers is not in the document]
rudyardrichter: if it isn't initialized then it should blow up anyways I think, the only case this is different is that if `is_admin == "True"` (or some truthy value) then `is_admin is not True == True` but `not is_admin == False`. It should be a boolean though, so I suggested the more idiomatic way to paint the bike shed...

fence/resources/admin/admin_groups.py
```diff
@@ -0,0 +1,164 @@
...
+from fence.errors import NotFound
+import json
+from fence.errors import UserError
+
+
+
+#### GROUPS ####
+
+
+def create_group(current_session, groupname, description):
```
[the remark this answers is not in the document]
rudyardrichter: I see what you mean with this—fair point. Can keep it this way

fence/__init__.py
```diff
@@ -192,9 +195,14 @@ def check_csrf():
         return
```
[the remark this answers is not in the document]
jbarno: I added it to the issue I had originally opened.

tests/admin/test_admin_groups.py
```diff
@@ -0,0 +1,135 @@
...
+                      for item in db_session.query(AccessPrivilege).filter_by(user_id=user.id).all()
+                      if item.project_id != None}
+    expected_projects = {'test_project_1'}
+    assert expected_projects == user_projects
+
+
+def test_get_group_projects(db_session, awg_groups):
+    group_projects = set(adm.get_group_projects(db_session, 'test_group_4'))
+    expected_projects = {'test_project_7', 'test_project_6'}
+    assert expected_projects == group_projects
```
jbarno: comparing set to dict?

tests/admin/test_admin_projects.py
```diff
@@ -0,0 +1,61 @@
...
+        for project in projects
+    }
+    expected = {
+        'test_project_1': {
+            'auth_id': u'phs_project_1',
+            'associated buckets': [],
+            'description': None,
+            'name': u'test_project_1'
+        },
+        'test_project_2': {
```
jbarno: feels brittle, but I get it.

tests/admin/test_admin_users.py
```diff
@@ -0,0 +1,134 @@
...
+    assert user.email == "insert_user@fake.com"
+
+
+def test_delete_user(db_session, awg_users):
+    user = db_session.query(User).filter(User.username == "awg_user").first()
+    assert user != None
+    user_id = user.id
+    user_access = db_session.query(AccessPrivilege).filter(AccessPrivilege.user_id == user_id).all()
+    assert user_access != []
+    user_groups = db_session.query(UserToGroup).filter(UserToGroup.user_id == user_id). all()
```
jbarno: strange space on the .all()

tests/conftest.py
jbarno: I got a good idea how to do it slickly. 

Also do you use auth_client fixture? Because it will not respect the existing app config value of MOCK_AUTH.

fence/blueprints/admin.py
```diff
@@ -1,24 +1,478 @@
...
+def create_user():
+    """
+    Create a user on the userdatamodel database
+    Returns a json object
+    """
+    username = request.get_json().get('name',None)
+    role = request.get_json().get('role',None)
+    email = request.get_json().get('email',None)
+    current_app.logger.debug(
+        ("create_user:\n\tname: {0}\n\t"
```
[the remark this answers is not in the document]
jesuspguofc: I know, I will remove those pesky logs altogether... :P

fence/blueprints/admin.py
jesuspguofc: There is a version of swagger that has them, but hey should probably be updated before merging.

fence/blueprints/admin.py
```diff
@@ -1,24 +1,478 @@
...
+    response = jsonify(adm.add_user_to_groups(current_session, username, groups=groups))
+    return response
+
+
+@blueprint.route('/user/<username>/groups', methods=['DELETE'])
+@login_required({'admin'})
+@admin_required
+def remove_user_from_groups(username):
+    """
+    Create a user to group relationship in the database
```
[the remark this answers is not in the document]
jesuspguofc: Semantics....

fence/blueprints/admin.py
```diff
@@ -1,24 +1,478 @@
...
+    response = jsonify(adm.remove_user_from_groups(current_session, username, groups=groups))
+    return response
+
+
+@blueprint.route('/user/<username>/projects', methods=['DELETE'])
+@login_required({'admin'})
+@admin_required
+def remove_user_from_projects(username):
+    """
+    Create a user to group relationship in the database
```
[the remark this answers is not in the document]
jesuspguofc: white lie?

fence/blueprints/admin.py
```diff
@@ -1,24 +1,478 @@
...
+@blueprint.route('/projects/<projectname>', methods=['GET'])
+@login_required({'admin'})
+@admin_required
+def get_project(projectname):
+    """
+    Get the information related to a project
+    from the userdatamodel database
+    Returns a json object
+    """
+    current_app.logger.debug("get_project")
```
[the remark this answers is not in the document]
jesuspguofc: That's a good idea. TBH I don't think there is time right now, but it should definitely be opened as a ticket, and expanded. Logging is pretty lacking in this code, and that's definitely my bad

fence/blueprints/admin.py
```diff
@@ -1,24 +1,478 @@
...
+    Create a user to group relationship in the database
+    Returns a json object
+    """
+    groups = request.get_json().get('groups', [])
+    response = jsonify(adm.add_project_to_groups(current_session, username, groups=groups))
+    return response
+
+@blueprint.route('/projects/<projectname>/bucket/<bucketname>', methods=['POST'])
+@login_required({'admin'})
+@admin_required
```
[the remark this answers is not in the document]
jesuspguofc: There are a couple of exceptions around where you need to be logged in but not an admin. like to get your own permissions.

fence/blueprints/admin.py
jesuspguofc: Trying to save bytes here

fence/resources/admin/admin_groups.py
```diff
@@ -0,0 +1,161 @@
...
+                au.connect_user_to_project(current_session,
+                                           us.get_user(current_session,
+                                                       user.username),
+                                           project_info)
+        except NotFound:
+            pass
+    return {"success": "users {0} connected to project {1}".format(
+        [user.username for user in users], project)}
+
+def add_projects_to_group(current_session, groupname, projects=[]):
```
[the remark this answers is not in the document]
jesuspguofc: Thanks, changing!

fence/resources/admin/admin_groups.py
```diff
@@ -0,0 +1,164 @@
...
+    if projectname not in projects_to_keep:
+        try:
+            us.remove_user_from_project(current_session,
+                                        us.get_user(current_session, username),
+                                        pj.get_project(current_session, projectname))
+        except NotFound as e:
+            # somehow the user was not linked to that project
+            pass
+
+def remove_projects_from_group(current_session, groupname, projects=[]):
```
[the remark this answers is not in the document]
jesuspguofc: Changed them, thanks for pointing them out!!

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,216 @@
...
+    group as gp,
+    project as pj,
+    user as us,
+    userdatamodel as udm
+)
+from flask import current_app as capp
+
+
+
+__all__ = ['connect_user_to_project', 'get_user_info', 'get_all_users',
```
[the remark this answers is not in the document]
jesuspguofc: Would that show only the public ones of the private/protected ones too?

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,216 @@
...
+    if role:
+        is_admin = True if role == 'admin' else False
+    else:
+        is_admin = usr.is_admin
+    usr.is_admin = is_admin
+    usr.username = new_name or usr.username
+    return us.get_user_info(current_session, usr.username)
+
+
+def add_user_to_projects(current_session, username, projects=[]):
```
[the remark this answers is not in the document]
jesuspguofc: thank you!

fence/resources/user/__init__.py
```diff
@@ -68,7 +72,7 @@ def get_user_info(user, session):
         info['certificates_uploaded'] = [
             c.name for c in user.application.certificates_uploaded]
         info['message'] = user.application.message
-    return flask.jsonify(info)
+    return info
 
 
 def send_mail(send_from, send_to, subject, text, server, certificates=None):
```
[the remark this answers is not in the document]
jesuspguofc: Yeah, I can't tell what this is for but the ways of fence are inscrutable, so I'd rather not change it....

fence/resources/userdatamodel/userdatamodel_project.py
jesuspguofc: I removed most of them, one was needed and destroyed the universe for a while. I kept that one and removed the others.

fence/resources/userdatamodel/userdatamodel_project.py
```diff
@@ -0,0 +1,283 @@
...
+        if num == 1:
+            for storage in storage_access:
+                provider = current_session.query(
+                    CloudProvider).filter(
+                        CloudProvider.id == storage.provider_id).first()
+                usr = current_session.query(
+                    User).filter(User.id == access.user_id).first()
+                users_to_remove.append((provider,usr))
+                current_session.delete(usr)
+        current_session.delete(access)
```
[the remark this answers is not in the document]
jesuspguofc: The link between user and project goes in the Access Privilege table

fence/resources/userdatamodel/userdatamodel_user.py
```diff
@@ -0,0 +1,125 @@
...
+    Group,
+    UserToGroup,
+)
+
+from fence.errors import (
+    NotFound,
+    UserError,
+)
+
+from flask_sqlalchemy_session import current_session as curr_sess
```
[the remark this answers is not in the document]
jesuspguofc: Good catch! There used to be a function that would call one of the user related ones from I different place without using the current_session parameter, but seems like I fixed that and forgot to remove this.

tests/admin/test_admin_groups.py
```diff
@@ -0,0 +1,135 @@
...
+from fence.errors import UserError, NotFound
+
+
+def test_get_group(db_session, awg_users):
+    info = adm.get_group_info(db_session, "test_group_2")
+    assert info['name'] == 'test_group_2'
+    assert info['description'] == 'the second test group'
+    expected_projects = ['test_project_1', 'test_project_2']
+    expected_projects.sort()
+    info['projects'].sort()
```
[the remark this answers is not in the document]
jesuspguofc: Sure! :)
I changed them to sets to make more slick, as per Rudy's comment

tests/admin/test_admin_groups.py
jesuspguofc: Nopes, that's also a set

tests/admin/test_admin_projects.py
jesuspguofc: yeah, it kills me to have all that data in the test itself, but if felt like using the original file to get that data was cheating...

tests/admin/test_admin_users.py
jesuspguofc: We all() need our space

tests/conftest.py
jesuspguofc: that seems troubling, I'd like to review that with you

fence/resources/admin/admin_users.py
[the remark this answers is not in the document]
rudyardrichter: `inspect` is not the right tool here, since `__all__` is just the alternative to named imports to make `import *` not as bad

fence/resources/admin/admin_projects.py
```diff
@@ -0,0 +1,103 @@
...
+    return pj.get_all_projects(current_session)
+
+def create_project(current_session, projectname, authid, storageaccesses):
+    """
+    Create a project with the specified auth_id and
+    storage access.
+    Returns a dictionary.
+    """
+    if pj.create_project(current_session, projectname, authid, storageaccesses):
+        return {'result': 'success'}
```
rudyardrichter: no return in this function if `create_project` returns false-y

fence/resources/admin/admin_projects.py
```diff
@@ -0,0 +1,103 @@
+from fence.resources import project as pj
```
rudyardrichter: Can we do the same thing in this file to preserve the interface?
```python
from fence.resources.project import (
    list_buckets_on_project, get_all_projects,
)
```
etc

fence/resources/admin/admin_providers.py
```diff
@@ -0,0 +1,55 @@
+from fence.resources import (
```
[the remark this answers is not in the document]
rudyardrichter: Could just use re-exports here.

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,219 @@
...
+            raise e
+    for project in projects_to_remove:
+        remove_user_from_project(current_session, usr, project)
+    return {"result": responses}
+
+
+def disconnect_user_from_group(current_session, usr, groupname):
+    grp = gp.get_group(current_session, groupname)
+    if not grp:
+        return {"warning": ("Group {0} doesn't exist".format(group))}
```
rudyardrichter: `group` undefined

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    Create a user for all the projects or groups in the list.
+    If the user already exists, to avoid unadvertedly changing it, we suggest update
+    Returns a dictionary.
+    """
+    try:
+        usr = us.get_user(current_session, username)
+        raise UserError(("Error: user already exist. If this is not a"
+               " mistake, please, retry using update"))
+        return msg
```
rudyardrichter: `msg` undefined

fence/resources/admin/admin_users.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+def disconnect_user_from_group(current_session, usr, groupname):
+    grp = gp.get_group(current_session, groupname)
+    if not grp:
+        return {"warning": ("Group {0} doesn't exist".format(group))}
+
+    response = gp.remove_user_from_group(current_session, usr, grp)
+    projects = gp.get_group_projects(current_session, groupname)
+    projects_data = [ pj.get_project(current_session, project).auth_id for project in projects]
```
rudyardrichter: This variable doesn't get used

fence/blueprints/admin.py
```diff
@@ -1,24 +1,478 @@
...
+@blueprint.route('/projects/<projectname>/groups', methods=['PUT'])
+@login_required({'admin'})
+@admin_required
+def add_project_to_groups(projectname):
+    """
+    Create a user to group relationship in the database
+    Returns a json object
+    """
+    groups = request.get_json().get('groups', [])
+    response = jsonify(adm.add_project_to_groups(current_session, username, groups=groups))
```
rudyardrichter: `username` not defined

fence/resources/admin/admin_projects.py
jesuspguofc: but then I need to change the names upstream so the don't clash...
me no likey...
why so much discrimination towards abbreviations, they are here to help!

fence/resources/admin/admin_users.py
jesuspguofc: true, thanks

fence/resources/admin/admin_users.py
jesuspguofc: true, thanks

fence/blueprints/admin.py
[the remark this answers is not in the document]
rudyardrichter: Added in 8c1d0e4

fence/resources/userdatamodel/userdatamodel_project.py
rudyardrichter: bump

PR #216, opened 2018-05-15
feat(caseinsensitive): Lookup usernames in lower case

zflamig:


rudyardrichter:
also, some tests are using not necessarily case-sensitive usernames but definitely not uniform case, like here: https://github.com/uc-cdis/fence/blob/db8a7bb762591880ea1a9122d62107676f0f401b/tests/dbgap_sync/conftest.py#L32

rudyardrichter:
also here, might want to change to  use uniform case: https://github.com/uc-cdis/fence/blob/2c8b92f9e2ff42bc2c477469af16169534d1e23a/fence/scripting/fence_create.py#L259

PR #228, opened 2018-05-21
fix(index-acls): check if index acl field is empty and fallback on me…

Avantol13:
…tadata acls


Avantol13:
@philloooo thinks it's unecessary. "don’t want to encourage continue using metadata acl if the latest indexd is deployed"

PR #233, opened 2018-05-23
fix(errors): improve/clarify error output, add tests for unsupported …

thanh-nguyen-dang:
@Avantol13 r?

### DIFF
--- a/fence/blueprints/data.py
+++ b/fence/blueprints/data.py
@@ -112,19 +112,19 @@ class IndexedFile(object):
                         or (protocol == 'http' and file_location.protocol == 'https')):
                     signed_url = file_location.get_signed_url(
                         action, expires_in, public_data=self.public)
+            if not signed_url:
+                raise NotFound(
+                    'File {} does not have a location with specified '
+                    'protocol {}.'.format(self.file_id, protocol))
 
         # no protocol specified, return first location as signed url
         elif len(self.indexed_file_locations) > 0:
             signed_url = self.indexed_file_locations[0].get_signed_url(
                 action, expires_in, public_data=self.public)
         else:
-            # will get caught below when signed_url is still None
-            pass
-
-        if not signed_url:
-            raise NotFound(
-                'File {} does not have a location with specified '
-                'protocol {}.'.format(self.file_id, protocol))
+            # at this point, they haven't specified a protocol and we don't
+            # have any actual locations, error out
+            raise NotFound('Can\'t find any file locations.')
 
         return signed_url
 
@@ -161,7 +161,9 @@ class IndexedFile(object):
                 'Not Found. indexd could not find {}'
                 '\nIndexd\'s response: {}'
                 .format(url + self.file_id, res.text))
-            raise NotFound("Can't find a location for the data")
+            raise NotFound(
+                "No indexed document found with id {}"
+                .format(self.file_id))
         else:
             raise UnavailableError(res.text)
 
--- a/fence/restful.py
+++ b/fence/restful.py
@@ -1,6 +1,7 @@
 import flask
 import flask_restful
 from authlib.specs.rfc6749.errors import OAuth2Error
+from werkzeug.exceptions import HTTPException
 
 from fence.errors import APIError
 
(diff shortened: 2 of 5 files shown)


fence/restful.py
```diff
@@ -22,6 +23,11 @@ def handle_error(error):
             return flask.jsonify(message=error.message), error.code
     elif isinstance(error, OAuth2Error):
         return flask.jsonify(error.get_body()), error.status_code
+    elif isinstance(error, HTTPException):
+        return (
+            flask.jsonify(message=error.__dict__.get('description')),
```
rudyardrichter: can use `getattr(error, 'description')`

wsgi.py
```diff
@@ -2,5 +2,5 @@
 from fence import app
 
 app_init(app)
-app.debug = False
+app.debug = True
```
rudyardrichter: We almost certainly don't want this on by default in `wsgi.py`, unless we're planning on making changes to the logging in kubernetes or something; this is too verbose to be usable ordinarily

fence/blueprints/data.py
```diff
@@ -161,7 +162,9 @@ def _get_index_document(self):
                 'Not Found. indexd could not find {}'
                 '\nIndexd\'s response: {}'
                 .format(url + self.file_id, res.text))
-            raise NotFound("Can't find a location for the data")
+            raise NotFound(
```
rudyardrichter: It looks like there's two `raise` in a row here?

fence/blueprints/data.py
```diff
@@ -118,8 +118,9 @@ def _get_signed_url(self, protocol, action, expires_in):
             signed_url = self.indexed_file_locations[0].get_signed_url(
                 action, expires_in, public_data=self.public)
         else:
-            # will get caught below when signed_url is still None
-            pass
+            # at this point, they haven't specified a protocol and we don't
```
rudyardrichter: Can just delete this `else` block; `signed_url == None` so it will hit the `NotFound` immediately below this

fence/blueprints/data.py
thanh-nguyen-dang: what is the second one?

fence/blueprints/data.py
thanh-nguyen-dang: the meaning is different,  we need another message when user does not input the protocol --> completely no location for that file_id

fence/blueprints/data.py
rudyardrichter: one on 162, and this one on 165

fence/restful.py
thanh-nguyen-dang: fixed it

wsgi.py
thanh-nguyen-dang: yeah, I forgot to set it back

fence/blueprints/data.py
thanh-nguyen-dang: the one on 162 is logger only

fence/blueprints/data.py
rudyardrichter: maybe then move the `if not signed_url` at 125 up to after 113, to put that check where it's logically relevant? and then leave this block here

thanh-nguyen-dang: yeah, make sense

fence/blueprints/data.py
rudyardrichter: oh, sorry...I, er, blame the github folding :P

fence/blueprints/data.py
```diff
@@ -112,19 +112,19 @@ def _get_signed_url(self, protocol, action, expires_in):
...
-            # will get caught below when signed_url is still None
-            pass
-
-        if not signed_url:
-            raise NotFound(
-                'File {} does not have a location with specified '
-                'protocol {}.'.format(self.file_id, protocol))
+            # at this point, they haven't specified a protocol and we don't
+            # have any actual locations, error out
+            raise NotFound('Can\'t find any file locations.')
```
rudyardrichter: :+1:

PR #237, opened 2018-05-29
feat(roles): add basic client interface to arborist

### LINKED ISSUE #189: PXD-277 ⁃ support CRUD roles and attach roles to users
no role hierachy, just allow CRUD roles and attach/detach roles to users

rudyardrichter:
To resolve #189.

~TODO:~
- [x] merge userdatamodel PR first; update version pin here.


fence/__init__.py
```diff
@@ -163,6 +171,10 @@ def app_sessions(app):
     if configured_fence:
         app.fence_client = OAuthClient(**app.config['OPENID_CONNECT']['fence'])
     app.session_interface = UserSessionInterface()
+    if 'ARBORIST' in app.config:
```
Avantol13: in the new configuration structure, you're guaranteed to have a value there (so long as it's set in the default config). that being said, you could just check truthiness here. e.g. `if app.config.get('ARBORIST'):` which would also handle the case where the default config has `None` set (meaning it requires user to add info). super convoluted comment, hopefully I'm making sense.

fence/blueprints/rbac.py
```diff
@@ -0,0 +1,183 @@
...
+        List[fence.model.Policy]: list of policy models
+
+    Raises:
+        - ValueError: if any of the policy IDs do not correspond to an existing
+          policy
+    """
+    policies = []
+    with flask.current_app.db.session as session:
+        for policy_id in policy_ids:
+            policy = session.query(Policy).filter_by(ID=policy_id).first()
```
Avantol13: `ID` -> `id` from change in userdatamodel

fence/blueprints/rbac.py
```diff
@@ -0,0 +1,183 @@
...
+
+from fence.auth import admin_login_required
+from fence.errors import NotFound, UserError
+from fence.models import Policy, User
+
+
+blueprint = flask.Blueprint('rbac', __name__)
+
+
+def _get_user(user_id):
```
Avantol13: this is minor (and pretty subjective) but I usually put internal/private/low-level functions below the public/high-level functions/endpoints that use them (lots of slashes there :P). to me it just makes sense reading top-down from high-level logic to lower-level logic. but eh, it's up to you

fence/blueprints/rbac.py
```diff
@@ -0,0 +1,183 @@
...
+        - Policies argument is there
+        - All the listed policies are valid
+            - Contain correct fields
+            - Actually exist in arborist
+
+    Args:
+        policy_ids (List[str]): list of policy IDs
+
+    Return:
+        List[str]: the same policy_ids, if they validated
```
Avantol13: Oh shoot. I think I've been doing docstring types wrong. is it really supposed to be `List[str]` and not `List(str)` ?

fence/blueprints/rbac.py
```diff
@@ -0,0 +1,183 @@
...
+
+    Raises:
+        UserError: if the policy ID list fails to validate
+    """
+    if not policy_ids:
+        raise UserError('JSON missing required value `policies`')
+    missing_policies = flask.current_app.arborist.policies_not_exist(
```
Avantol13: do we need a check here to see if current_app has arborist set up? what if we don't configure arborist and hit /user/<user_id>/policies/ ?

fence/blueprints/rbac.py
```diff
@@ -0,0 +1,183 @@
...
+
+@blueprint.route('/user/<user_id>/policies/', methods=['POST'])
+@admin_login_required
+def grant_policy_to_user(user_id):
+    """
+    Grant additional policies to a user.
+    """
+    policy_ids = _validate_policy_ids(flask.request.get_json().get('policies'))
+
+    with flask.current_app.db.session as session:
```
Avantol13: could use your _get_user helper here

fence/__init__.py
rudyardrichter: yep that makes sense, I'll update this

fence/blueprints/rbac.py
rudyardrichter: Interesting, I think I tend to define lowest-level stuff first and then work upwards, but I've never really had a strong opinion on this. If you prefer it this way then I'll defer to that :smile:

fence/blueprints/rbac.py
rudyardrichter: For the docstrings it doesn't _really_ matter, I just like to copy the same syntax from [type hinting from PEP 484](https://docs.python.org/3/library/typing.html).

fence/blueprints/rbac.py
rudyardrichter: If the app isn't configured for using arborist then the blueprint doesn't get loaded, so the endpoint shouldn't exist. [it gets registered here](https://github.com/uc-cdis/fence/pull/237/files#diff-feb6ac6f84a84b339e1cd1e1fc97e3daR91)

rudyardrichter: and I just noticed that that is in the wrong case :no_mouth: fixing...

fence/blueprints/rbac.py
rudyardrichter: I had to get rid of the `_get_user` elsewhere because of a bug with how I was using sqlalchemy, so I think this can be left alone

fence/rbac/client.py
```diff
@@ -0,0 +1,237 @@
...
+                "resources": [
+                    "/",
+                    "/a",
+                    "/a/b",
+                    "/x",
+                    "/x/y",
+                ]
+            }
+        """
+        return _request_get_json(requests.get(self._resource_url))
```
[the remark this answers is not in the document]
rudyardrichter: We're unlikely to have _that_ many resources (maybe tens or hundreds in typical usage right now)—but it almost sounds like this comment was referring to somewhere else?

fence/blueprints/rbac.py
Avantol13: okay, that makes sense. I'll try to remember to defer to that then

fence/rbac/client.py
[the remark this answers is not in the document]
Avantol13: I had assumed each node in the graph would be a resource, but that's not the case? like project/case/aliquot/.. etc.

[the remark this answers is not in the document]
Avantol13: even if we support that later, we may want to consider how we're handling this. But I may be misunderstanding how the resource structure is going to be used/look

fence/blueprints/rbac.py
Avantol13: No pressure to change the way you do it, like I said, pretty subjective

fence/rbac/client.py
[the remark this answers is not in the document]
rudyardrichter: In the extreme case there's support for that, but it's not likely that every node will become its own resource because in most cases the security won't be so fine-grained. For other use cases like DUOS we'll have to figure out something specific later on.

fence/blueprints/rbac.py
```diff
@@ -0,0 +1,181 @@
...
+
+from fence.auth import login_required
+from fence.errors import NotFound, UserError
+from fence.models import Policy, User
+
+
+blueprint = flask.Blueprint('rbac', __name__)
+
+
+@blueprint.route('/policy/', methods=['GET'])
```
Avantol13: dunno how we want to handle, but for service account endpoints I opted for plural everywhere so there's some consistency. e.g. `GET /policies/` and `GET /user/<id>/policies`. I guess it's up to you since this is slightly different, though to me plural makes more sense here, especially since the response itself uses `policies` in the response JSON

fence/blueprints/rbac.py
```diff
@@ -0,0 +1,181 @@
...
+    """
+    List the policies that this user has access to.
+
+    Output will be in the same format as the ``/policy/`` endpoint, but
+    only containing policies this user has access to.
+    """
+    return flask.jsonify({'policies': _get_user_policy_ids(user_id)})
+
+
+@blueprint.route('/user/<user_id>/policies/', methods=['POST'])
```
Avantol13: Would it make sense to use a flaskrestful class for this like some of the login stuff? I know that's a pretty big refactor and it's not a big deal, but seems like the route we've been going with new stuff

rudyardrichter: Hmm, to be honest I hadn't really thought about this. I think there's not much opportunity for code additional reuse but could be missing something

rudyardrichter:
first merge `feat/arborist-sync` into this branch

fence/blueprints/rbac.py
```diff
@@ -0,0 +1,181 @@
...
+        {
+            "policies": [
+                "policy-abc",
+                "policy-xyz"
+            ]
+        }
+    """
+    with flask.current_app.db.session as session:
+        policies = _list_all_policies(session)
+    return flask.jsonify({'policies': [policy.id for policy in policies]})
```
philloooo: should have create and delete endpoints for policy

fence/blueprints/rbac.py
```diff
@@ -0,0 +1,234 @@
...
+    with flask.current_app.db.session as session:
+        user = session.query(User).filter(User.id == user_id).first()
+        if not user:
+            raise NotFound('no user exists with ID: {}'.format(user_id))
+        return [policy.id for policy in user.policies]
+
+
+def _validate_policy_ids(policy_ids):
+    """
+    Check some user-inputted policy IDs which should correspond to roles in
```
philloooo: nitpick : should correspond to policies not roles in arborist?

fence/blueprints/rbac.py
```diff
@@ -0,0 +1,234 @@
...
+    arborist.
+
+    Check:
+        - Policies argument is there
+        - All the listed policies are valid
+            - Contain correct fields
```
philloooo: remove?

openapis/swagger.yaml
```diff
@@ -943,6 +945,206 @@ paths:
...
+      summary: Create a new policy
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                'id':
+                  type: string
+                'resources':
```
philloooo: remove resources and roles

rudyardrichter: Those would still be necessary since fence will just post the policy through to arborist. We could remove that interaction entirely and separate them (seems like the worst choice)? or make it so fence will only use things which exist in arborist already (maybe better)?

fence/blueprints/rbac.py
rudyardrichter: fixed

fence/blueprints/rbac.py
rudyardrichter: fixed

fence/blueprints/rbac.py
rudyardrichter: added in 3a22989

Issue #240, opened 2018-05-30
PXD-247 ⁃ As a developer, I want to add tables in the database for the Service Account Registration feature.


philloooo:
Please refer to the attachment.... 

* **UserServiceAccount Table**
  * user_id     int
  * id     int    PK
  * google_unique_id   Varchar   
  * email    Varchar
  * google_project_id    Varchar 

** 

* **ServiceAccountAccessPrivilege Table**

* 

  * id     int    PK
  * project_id    int
  * service_account_id    int 

** 

* **ServiceAccountToGoogleBucketAccessGroup Table**

  * id     int    PK
  * service_account    int   
  * expires     bigint
  * access_group    int

┆Attachments: <a href="https:&#x2F;&#x2F;jira.opensciencedatacloud.org&#x2F;secure&#x2F;attachment&#x2F;17133&#x2F;image.png">image.png</a>

PR #245, opened 2018-06-07
Added tables for the Service Account Registration feature. #240

benJrohrer:
- Added UserServiceAccount, ServiceAccountAccessPrivilege and ServiceAccountToGoogleBucketAccessGroup Tables. 
- Added unit test in test_datamodel.py to test relationships between/within new tables and existing tables (GoogleAccessBucketGroup and Project).


fence/models.py
```diff
@@ -346,6 +346,76 @@ class GoogleProxyGroupToGoogleBucketAccessGroup(Base):
...
+    email = Column(
+        String,
+        nullable=False)
+
+    google_project_id = Column(
+        String,
+        nullable=False)
+
+class ServiceAccountAccessPrivilege(Base):
+    __tablename__="service_account_access_privilege"
```
Avantol13: minor style inconsistency: need spaces on either side of `=`

fence/models.py
```diff
@@ -346,6 +346,76 @@ class GoogleProxyGroupToGoogleBucketAccessGroup(Base):
...
+        ForeignKey(UserServiceAccount.id),
+        nullable=False)
+
+    service_account = relationship(
+        'UserServiceAccount',
+        backref=backref(
+            'access_privileges', cascade='all, delete-orphan'))
+
+class ServiceAccountToGoogleBucketAccessGroup(Base):
+    __tablename__="service_account_to_google_bucket_access_group"
```
Avantol13: same style thing as above, spaces around `=`

tests/test_datamodel.py
```diff
@@ -19,3 +27,53 @@ def test_user_delete_cascade(db_session):
         db_session.query(Client).filter_by(client_id=client.client_id).count()
         == 0
     )
+
```
Avantol13: needs another newline here. should be 2 newlines between functions

tests/test_datamodel.py
```diff
@@ -19,3 +27,53 @@ def test_user_delete_cascade(db_session):
...
+        expires=0,
+        access_group_id=1)
+    db_session.add(project)
+    db_session.add(bucket)
+    db_session.add(user_sa)
+    db_session.add(sa_access_privilege)
+    db_session.add(gbag)
+    db_session.add(sa_to_gbag)
+    db_session.commit()
+    assert project.sa_access_privileges[0].__class__ == ServiceAccountAccessPrivilege
```
Avantol13: more style nitpicking: try not to go over 79 chars in a line. You can use parens
```
    assert (
        project.sa_access_privileges[0].__class__
        == ServiceAccountAccessPrivilege
    )
    
```

Avantol13: same thing for lines below this

fence/models.py
```diff
@@ -346,6 +346,76 @@ class GoogleProxyGroupToGoogleBucketAccessGroup(Base):
             'proxy_groups_with_access', cascade='all, delete-orphan')
     )
```
Avantol13: style nitpick: need another newline. 2 between classes/functions

fence/models.py
```diff
@@ -346,6 +346,76 @@ class GoogleProxyGroupToGoogleBucketAccessGroup(Base):
...
+        nullable=False)
+
+    service_account = relationship(
+        'UserServiceAccount',
+        backref=backref(
+            __backrefname__, cascade='all, delete-orphan'))
+
+    expires = Column(
+        BigInteger,
+        nullable=False)
```
Avantol13: let's allow this to be null in case we want a service account to always have access for some reason. I can't think of any reasons off the top of my head... but it looks like the rest of the `expires` fields allow null too. We're going to be maintaining the table so we can safely assume that when it needs to expire, we'll be adding that field

fence/models.py
```diff
@@ -346,6 +346,76 @@ class GoogleProxyGroupToGoogleBucketAccessGroup(Base):
...
+
+    service_account = relationship(
+        'UserServiceAccount',
+        backref=backref(
+            'access_privileges', cascade='all, delete-orphan'))
+
+class ServiceAccountToGoogleBucketAccessGroup(Base):
+    __tablename__="service_account_to_google_bucket_access_group"
+    __backrefname__="to_access_groups"
```
Avantol13: I wouldn't bother making a variable for this, just put the string in the `relationship` call. We usually don't use `__varname__` for locals either, since it's usually denoting some sort of internal Python var (or library var). in this case, `__tablename__` is required by sqlalchemy

fence/models.py
```diff
@@ -346,6 +346,76 @@ class GoogleProxyGroupToGoogleBucketAccessGroup(Base):
             'proxy_groups_with_access', cascade='all, delete-orphan')
     )
 
+class UserServiceAccount(Base):
+    __tablename__ = "user_service_account"
+    id = Column(Integer, primary_key=True)
+
+    google_unique_id = Column(
```
Avantol13: I'd add this note (it exists in a previous table too):

```
    # The uniqueId google provides to resources is ONLY unique within
    # the given project, so we shouldn't rely on that for a primary key (in
    # case we're ever juggling mult. projects)
```

tests/test_datamodel.py
```diff
@@ -19,3 +27,58 @@ def test_user_delete_cascade(db_session):
...
+        expires=0,
+        access_group_id=1)
+    db_session.add(project)
+    db_session.add(bucket)
+    db_session.add(user_sa)
+    db_session.add(sa_access_privilege)
+    db_session.add(gbag)
+    db_session.add(sa_to_gbag)
+    db_session.commit()
+    assert (project.sa_access_privileges[0].__class__
```
Avantol13: Another super minor PEP8 style thing, when breaking up lines with enclosing ()'s, the indentation can be done like:
```
    assert (user_sa.access_privileges[0].__class__
            == ServiceAccountAccessPrivilege)
```
or
```
    assert (
        user_sa.access_privileges[0].__class__
        == ServiceAccountAccessPrivilege)
```
I prefer the last version (it may even be in our style guide, I don't remember)

PR #249, opened 2018-06-11
PXD-641 Logout error handling

benJrohrer:
- error is not displayed if someone tries to logout when they are not logged in. WIP.


Avantol13:
need to update/remove unit test that checks for a 401. will always be 302 now

PR #250, opened 2018-06-12
Feat/refactor config

Avantol13:
- cleaned up fence's configuration (general refactoring and removing unecessary configs)
- moved to a new YAML file that should be located outside the repo
- addition of new command line tool to ease development with new configuration structure
- Fence searches configured directories for the new config.yaml file
- Backwards-compatibility with local_settings.py for now
- Fence updates some paths in the new config based on other values (a template-like syntax)
- Fence now ignores configuration variables it doesn't recognize
    - It loads in default then overlays the provided configuration over that, discarding any extra config not defined in the default configuration
- Updated tests


fence/settings.py
```diff
@@ -1,47 +1,24 @@
...
-# only receive traffic from internal (CDIS) clients, and can safely use HTTP.
-os.environ['AUTHLIB_INSECURE_TRANSPORT'] = 'true'
+#: ``CONFIG_SEARCH_FOLDERS: List(str)``
+#: Folders to look in for the *config.yaml for fence
+CONFIG_SEARCH_FOLDERS = [
+    '/var/www/fence',
+    '/etc/gen3/fence'
+]
 
+#: ``ACCESS_TOKEN_COOKIE_NAME: str``
```
rudyardrichter: Could we actually enforce typechecking of the config values on the app using something like `jsonschema`?

cfg_help.py
```diff
@@ -0,0 +1,129 @@
...
+        '-n', '--name', default='fence-config.yaml', help='configuration file name if you used something '
+        'other than "fence-config.yaml"')
+
+    args = parser.parse_args()
+
+    if args.action == 'create':
+        sys.stdout.write(create_config_file(args.name, args.config_path))
+    elif args.action == 'get':
+        sys.stdout.write(get_config_file(args.name))
+    else:
```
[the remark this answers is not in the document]
Avantol13: Will change to ValueErrur

cfg_help.py
```diff
@@ -0,0 +1,129 @@
...
+LOCAL_CONFIG_FOLDER = '/etc/gen3/fence'
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    subparsers = parser.add_subparsers(title='action', dest='action')
+
+    create = subparsers.add_parser('create')
+    create.add_argument(
+        '-n', '--name', default='fence-config.yaml', help='configuration file name if you want something '
```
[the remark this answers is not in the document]
Avantol13: yeah, I'll clean it up a bit

README.md
```diff
@@ -42,13 +42,35 @@ pip install -r requirements.txt
...
+The configuration file itself will live outside of this repo (to
+prevent accidentally checking in sensitive information like database passwords).
+
+To create a new configuration file from the default configuration:
+
+```bash
+python cfg_help.py create
+```
+
+Now you can get the path where the new configuration file was created with:
```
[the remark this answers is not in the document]
Avantol13: I'll add some more clarification

cfg_help.py
```diff
@@ -0,0 +1,129 @@
...
+    possible_configs = []
+    for folder in search_folders:
+        file_name = file_name or '*config.yaml'
+        config_path = os.path.join(folder, file_name)
+        possible_files = glob.glob(config_path)
+        possible_configs.extend(possible_files)
+
+    if len(possible_configs) == 1:
+        return possible_configs[0]
+    elif len(possible_configs) > 1:
```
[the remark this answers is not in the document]
Avantol13: ooh, I like that. but it may be confusing for someone looking back on it, it wouldn't be super explicit. ideally the compiled python code will calculate the length once for this if/elif block, but I'm not sure how smart the compilation is. I'll update to add a local var with the calculated length to maybe save that extra evaluation.

cfg_help.py
```diff
@@ -0,0 +1,129 @@
...
+        raise IOError(
+            'Config file {file_name} could not be found in the search '
+            'locations: {search_folders}. Run '
+            '"cfg_help.py create -n {file_name}" first.'
+            .format(file_name=file_name, search_folders=search_folders))
+
+    return config_path
+
+
+def get_config_path(search_folders, file_name=None):
```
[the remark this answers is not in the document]
Avantol13: ah, yup, good point. Habit because of weird behavior with mutables

fence/config-default.yaml
```diff
@@ -0,0 +1,206 @@
...
+# url where indexd microservice is running (for signed urls primarily)
+# NOTE: ONLY provide if indexd is NOT running in {{BASE_URL}}/index
+#
+# INDEXD: https://example.com/index
+
+#### Multi-tenant fence setup ####
+# Use when fence is *both* the client and an IDP.
+# configure this to the base URL for the provider fence
+# (i.e. ``BASE_URL`` in the provider fence config).
+# OIDC_ISSUER: 'http://localhost:8080/user
```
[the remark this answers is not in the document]
Avantol13: It was commented out like this before but I can check the logic and see what makes sense

fence/__init__.py
```diff
@@ -137,6 +327,12 @@ def app_sessions(app):
     app.db = SQLAlchemyDriver(app.config['DB'])
     migrate(app.db)
     session = flask_scoped_session(app.db.Session, app)  # noqa
+
```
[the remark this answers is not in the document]
Avantol13: yup, I'll move it

fence/__init__.py
```diff
@@ -35,23 +37,80 @@
 CORS(app=app, headers=['content-type', 'accept'], expose_headers='*')
 
 
-def app_config(app, settings='fence.settings', root_dir=None):
+def app_config(
```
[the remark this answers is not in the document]
Avantol13: yeah, this section could use some refactoring. I'll split out some logic and try to make it clearer

cfg_help.py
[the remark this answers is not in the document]
rudyardrichter: yeah not really important here, just an idiom to keep in mind. the standard python (cpython) has no JIT and pretty minimal optimization so as far as I know it will always re-evaluate, which I think is why the idiom exists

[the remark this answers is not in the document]
rudyardrichter: actually I'm not so sure about the re-evaluation here—would be interested to know more about how it handles repeated expressions like this. but `if len(x):` will compute the length, while `if x:` is cheaper

cfg_help.py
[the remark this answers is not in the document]
rudyardrichter: good habit for the defensive python programmer :)

fence/config-default.yaml
```diff
@@ -0,0 +1,206 @@
+---
+############################ Fence Configuration ##############################
+# This file contains configuration necessary for the fence microservice.
+#
+# This is initially configured for minimal local development.
+# Read comments above each of the configuration variables to how to use them.
+
+#### General App Settings ####
+APP_NAME: Gen3 Data Commons
```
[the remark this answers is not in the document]
Avantol13: *sigh* yeah, I think that would be awesome to do but then we couldn't just load the YAML file into the flask settings, we'd have to do some intermediery parsing to get just the value. I think it may make sense to do this change if we decide to add type checking / validation in the future but at this point, I don't know if it's worth the effort. But I'm slightly biased since I'd be the one having to change everything :P If your opinion about it is strong... we need a third party to settle the dispute :D

fence/settings.py
Avantol13: I had looked into/considered type checking as part of this but the scope got so bloated from other requirements that I figured the schema/type-checking could be added on later

Avantol13:
In order to make this compatible with fence-create and when fence is importable, we need a consistent configuration interface (a new config.py with singleton that pulls in config from some static source like this yaml file). Then we need to NOT use flask's config in the code and instead pull from this singleton

PlanXCyborg:

<details>

```diff
--- cfg_help.py
+++ blackened
@@ -43,78 +43,84 @@
 import os
 from shutil import copyfile
 import sys
 
 ROOT_DIR = os.path.dirname(os.path.realpath(__file__))
-LOCAL_CONFIG_FOLDER = '/etc/gen3/fence'
+LOCAL_CONFIG_FOLDER = "/etc/gen3/fence"
 
 
 def main():
     parser = argparse.ArgumentParser()
-    subparsers = parser.add_subparsers(title='action', dest='action')
+    subparsers = parser.add_subparsers(title="action", dest="action")
 
-    create = subparsers.add_parser('create')
+    create = subparsers.add_parser("create")
     create.add_argument(
-        '-n', '--name',
-        default='fence-config.yaml',
+        "-n",
+        "--name",
+        default="fence-config.yaml",
         help=(
-            'configuration file name if you want something '
-            'other than "fence-config.yaml"')
+            "configuration file name if you want something "
+            'other than "fence-config.yaml"'
+        ),
     )
     create.add_argument(
-        '--config_path',
+        "--config_path",
         help=(
-            'Full path to a yaml config file to create. '
-            'Will override/ignore name if provided.')
+            "Full path to a yaml config file to create. "
+            "Will override/ignore name if provided."
+        ),
     )
 
-    edit = subparsers.add_parser('get')
+    edit = subparsers.add_parser("get")
     edit.add_argument(
-        '-n', '--name',
-        default='fence-config.yaml',
+        "-n",
+        "--name",
+        default="fence-config.yaml",
         help=(
-            'configuration file name if you used something '
-            'other than "fence-config.yaml"')
+            "configuration file name if you used something "
+            'other than "fence-config.yaml"'
+        ),
     )
 
     args = parser.parse_args()
 
-    if args.action == 'create':
+    if args.action == "create":
         sys.stdout.write(create_config_file(args.name, args.config_path))
-    elif args.action == 'get':
+    elif args.action == "get":
         sys.stdout.write(get_config_file(args.name))
     else:
-        raise ValueError(
-            '{} is not a recognized action.'.format(args.actions))
+        raise ValueError("{} is not a recognized action.".format(args.actions))
 
 
 def create_config_file(file_name, full_path=None):
     config_path = full_path or os.path.join(LOCAL_CONFIG_FOLDER, file_name)
     dir_name = os.path.dirname(config_path)
     if dir_name and not os.path.exists(dir_name):
         os.makedirs(os.path.dirname(config_path))
 
-    copyfile(os.path.join(ROOT_DIR, 'fence/config-default.yaml'), config_path)
+    copyfile(os.path.join(ROOT_DIR, "fence/config-default.yaml"), config_path)
 
     return config_path
 
 
 def get_config_file(file_name):
     search_folders = [LOCAL_CONFIG_FOLDER]
     try:
         config_path = get_config_path(search_folders, file_name=file_name)
     except IOError:
         raise IOError(
-            'Config file {file_name} could not be found in the search '
-            'locations: {search_folders}. Run '
-            '"cfg_help.py create -n {file_name}" first.'
-            .format(file_name=file_name, search_folders=search_folders))
+            "Config file {file_name} could not be found in the search "
+            "locations: {search_folders}. Run "
+            '"cfg_help.py create -n {file_name}" first.'.format(
+                file_name=file_name, search_folders=search_folders
+            )
+        )
 
     return config_path
 
 
-def get_config_path(search_folders, file_name='*config.yaml'):
+def get_config_path(search_folders, file_name="*config.yaml"):
     """
     Return the path of a single configuration file ending in config.yaml
     from one of the search folders.
 
     NOTE: Will return the first match it finds. If multiple are found,
@@ -129,17 +135,20 @@
     possible_cfgs_count = len(possible_configs)
     if possible_cfgs_count == 1:
         return possible_configs[0]
     elif possible_cfgs_count > 1:
         raise IOError(
-            'Multiple config.yaml files found: {}. Please specify which '
-            'configuration to use with "python run.py -c some-config.yaml".'
-            .format(str(possible_configs)))
+            "Multiple config.yaml files found: {}. Please specify which "
+            'configuration to use with "python run.py -c some-config.yaml".'.format(
+                str(possible_configs)
+            )
+        )
     else:
         raise IOError(
-            'Could not find config.yaml. Searched in the following locations: '
-            '{}'.format(str(search_folders)))
+            "Could not find config.yaml. Searched in the following locations: "
+            "{}".format(str(search_folders))
+        )
 
 
-if __name__ == '__main__':
+if __name__ == "__main__":
     main()
 

--- fence/resources/openid/google_oauth2.py
+++ blackened
@@ -22,11 +22,11 @@
     def get_auth_url(self):
         return self.flow.step1_get_authorize_url()
 
     def get_user_id(self, code):
         try:
-            if self.HTTP_PROXY and self.HTTP_PROXY.get('host'):
+            if self.HTTP_PROXY and self.HTTP_PROXY.get("host"):
                 proxy = httplib2.ProxyInfo(
                     proxy_type=httplib2.socks.PROXY_TYPE_HTTP,
                     proxy_host=self.HTTP_PROXY["host"],
                     proxy_port=self.HTTP_PROXY["port"],
                     proxy_rdns=True,
```
</details></br>

cfg_help.py
```diff
@@ -0,0 +1,153 @@
...
+
+def get_config_path(search_folders, file_name="*config.yaml"):
+    """
+    Return the path of a single configuration file ending in config.yaml
+    from one of the search folders.
+
+    NOTE: Will return the first match it finds. If multiple are found,
+    this will error out.
+    """
+    possible_configs = []
```
rudyardrichter: I think the comprehension version is sufficiently concise:
```python
possible_configs = [
    glob.glob(os.path.join(folder, file_name))
    for folder in search_folders
]
```

cfg_help.py
```diff
@@ -0,0 +1,153 @@
...
+    for folder in search_folders:
+        config_path = os.path.join(folder, file_name)
+        possible_files = glob.glob(config_path)
+        possible_configs.extend(possible_files)
+
+    possible_cfgs_count = len(possible_configs)
```
rudyardrichter: fun facts, python collections in general cache their lengths. so you can just write `len(possible_configs)` everywhere

cfg_help.py
```diff
@@ -0,0 +1,153 @@
...
+    if possible_cfgs_count == 1:
```
rudyardrichter: some personal preference probably, but I would order these as

```python
if not possible_configs:
    # found 0
    raise IOError

if len(possible_configs) > 1:
    # found > 1
    raise IOError

return possible_configs[0]
```

fence/blueprints/storage_creds/__init__.py
```diff
@@ -68,7 +69,13 @@ def list_sources():
                 "/google", "access to Google Cloud storage"
             }
         """
-        services = flask.current_app.config.get("STORAGES", [])
+        services = set(
```
rudyardrichter: can use set comprehension:
```
services = {info.get("backend") for ...}
```

fence/config.py
```diff
@@ -0,0 +1,324 @@
...
+    def __str__(self):
+        return str(self._configs)
+
+    def update(self, *args, **kwargs):
+        """
+        update configuration properties
+
+        support passing dictionary or keyword args
+        """
+        self._configs.update(*args)
```
rudyardrichter: wait, how does this work? `_configs` is a dict right?

run.py
```diff
@@ -1,14 +1,25 @@
...
-    app,
-    app_config,
-    app_config_oauth,
-    app_register_blueprints,
-    app_sessions,
+from fence import app, app_config, app_register_blueprints, app_sessions
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument(
```
rudyardrichter: this is delightful :clap:

bin/fence-create
```diff
@@ -265,30 +265,56 @@ def main():
     # get database information
     sys.path.append(args.path)
 
+    # load the configuration
+    # FIXME? This import doesn't work if it's at the top of the file and you call
```
rudyardrichter: what is the import error? seems like this shouldn't be a problem

cfg_help.py
rudyardrichter: (however checking booleanness is still fastest)

cfg_help.py
Avantol13: ah, much better :+1: will change

cfg_help.py
Avantol13: forget sometimes that Python is a bit smarter than I expect :D

cfg_help.py
Avantol13: I like that more too, I'll change it

fence/config.py
Avantol13: fixed, inherit from dict instead

run.py
Avantol13: :smile:

cfg_help.py
```diff
@@ -0,0 +1,151 @@
...
+
+"""
+import argparse
+import glob
+import os
+from shutil import copyfile
+import sys
+
+ROOT_DIR = os.path.dirname(os.path.realpath(__file__))
+LOCAL_CONFIG_FOLDER = "/etc/gen3/fence"
```
philloooo: can we not use `/etc` for dev local path? something like `HOME/.gen3` that's under user dir would make more sense?

Avantol13: yeah, we could. I had read that /etc is a good place for application configuration and stuck with it. I can update to dump in HOME

Avantol13: fixed in 2bf8bb9

fence/config-default.yaml
```diff
@@ -0,0 +1,438 @@
...
+#   Sibboleth? Use: '{{BASE_URL}}/login/shib'
+DEFAULT_LOGIN_URL: '{{BASE_URL}}/login/google'
+
+# Which Identity Provider fence will/can use
+#
+# See ``fence/blueprints/login/__init__.py`` for which identity providers can
+# be loaded.
+#
+ENABLED_IDENTITY_PROVIDERS:
+  # ID for which of the providers to default to
```
rudyardrichter: this has to match one of the providers below...not sure if there is anything clever to do here to make sure that's the case? it would be cool though :smile:

fence/config.py
```diff
@@ -0,0 +1,355 @@
...
+        # these cfg use to be in settings.py so we need to make sure they gets defaulted
+        default_config = yaml_load(
+            open(
+                os.path.join(
+                    os.path.dirname(os.path.abspath(__file__)), "config-default.yaml"
+                )
+            )
+        )
+
+        self._set_default("APPLICATION_ROOT", default_config=default_config)
```
rudyardrichter: this chunk could be consolidated:
```python
defaults = [...]
for default in defaults:
    self._set_default(default, default_config=default_config)
```

rudyardrichter: or change `_set_default` to `_set_defaults` :stuck_out_tongue:

tests/multi_tenant/test_multi_tenant.py
```diff
@@ -20,6 +21,11 @@ def test_redirect_from_oauth(fence_client_app, oauth_client):
     Test that the ``/oauth2/authorize`` endpoint on the client redirects to the
     ``/login/fence`` endpoint, also on the client.
     """
+    config.update(OPENID_CONNECT=fence_client_app.config["OPENID_CONNECT"])
```
rudyardrichter: can consolidate these to `config.update({"OPENID_CONNECT": ..., "BASE_URL": ...})`

tests/multi_tenant/test_multi_tenant.py
```diff
@@ -55,6 +61,11 @@ def test_login(
     # app.
     monkeypatch.setattr("authutils.token.keys.refresh_jwt_public_keys", lambda: None)
 
+    config.update(OPENID_CONNECT=fence_client_app.config["OPENID_CONNECT"])
+    config.update(BASE_URL=fence_client_app.config["BASE_URL"])
+    config.update(MOCK_AUTH=fence_client_app.config["MOCK_AUTH"])
+    config.update(DEFAULT_LOGIN_URL=fence_client_app.config["DEFAULT_LOGIN_URL"])
```
rudyardrichter: is the idea here to overwrite the normal config so the client fence will work? it would make more sense to have a separate yaml file for the client fence to load a config from, I think

tests/multi_tenant/conftest.py
```diff
@@ -71,8 +72,14 @@ def fence_client_app(
...
     """
     root_dir = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+
     client_app = flask.Flask("client_app")
-    app_init(client_app, test_settings, root_dir=root_dir)
+    app_init(
+        client_app,
+        test_settings,
+        root_dir=root_dir,
+        config_path=os.path.join(root_dir, "test-fence-config.yaml"),
```
rudyardrichter: so here, the client fence should use its own config file

tests/test_settings.py
```diff
@@ -1,174 +1,3 @@
...
-the data sets associated to the service accounts, or do not \
-adhere to the security policies.
-
-    """,
-}
-
-GUN_MAIL = {}
+#: ``CONFIG_SEARCH_FOLDERS: List(str)``
+#: Folders to look in for the *config.yaml for fence
+CONFIG_SEARCH_FOLDERS = ["/var/www/fence", "/etc/gen3/fence"]
```
rudyardrichter: should this include `~/.gen3`?

fence/config.py
Avantol13: ah, good point. will consolidate

tests/multi_tenant/test_multi_tenant.py
Avantol13: the problem is that there are mixed places in the code where the config singleton is used vs the flask app config. in this testing case we really just want to make sure that the current config has the necessary items used by the susequent app. it's super confusing and we probably want/need a better way of doing this test with the new shared config. I didn't want to do two files since then you'd have to remember to update both whenever there's a config change. Probably also we should change it so that the test_config gets generated on the fly with values replced... so we don't have two copies of the config

tests/test_settings.py
Avantol13: ah, yup forgot this spot

fence/config-default.yaml
Avantol13: Made a story for it PXD-2302

cfg_help.py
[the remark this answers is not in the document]
rudyardrichter: for posterity—I was wrong here! turns out `len` is kind of special

fence/blueprints/storage_creds/__init__.py
rudyardrichter:
```suggestion
        services = {info.get("backend") for info in config["STORAGE_CREDENTIALS"].values() if info.get("backend")}
```

bin/fence-create
```diff
@@ -265,30 +265,56 @@ def main():
...
+        try:
+            from fence.settings import DB
+        except ImportError:
+            pass
 
     if os.environ.get("BASE_URL"):
         BASE_URL = os.environ["BASE_URL"]
     else:
-        from fence.settings import BASE_URL
+        BASE_URL = config.get("BASE_URL")
```
rudyardrichter:
```suggestion
        BASE_URL = os.environ.get("BASE_URL") or config.get("BASE_URL")
```
similarly for other variables

cfg_help.py
```diff
@@ -0,0 +1,153 @@
...
+    Return the path of a single configuration file ending in config.yaml
+    from one of the search folders.
+
+    NOTE: Will return the first match it finds. If multiple are found,
+    this will error out.
+    """
+    possible_configs = []
+
+    for folder in search_folders:
+        possible_configs.extend(glob.glob(os.path.join(folder, file_name)))
```
rudyardrichter:
```suggestion
        globber = lambda folder: glob.glob(os.path.join(folder, file_name))
        possible_configs = list(map(globber, search_folders))
```
or
```suggestion
        globber = lambda folder: glob.glob(os.path.join(folder, file_name))
        possible_configs = [globber(folder) for folder in search_folders]
```

tests/multi_tenant/test_multi_tenant.py
rudyardrichter: bump

Avantol13: whoops, forgot this spot, fixed

PR #251, opened 2018-06-13
PXD-598 Lazy creation of Google Proxy Groups

benJrohrer:


fence/blueprints/storage_creds/google.py
```diff
@@ -6,13 +6,14 @@
...
 from fence.resources.google.utils import (
-    add_custom_service_account_key_expiration
+    add_custom_service_account_key_expiration,
+    create_google_access_key,
+    get_service_account,
+    get_or_create_service_account,
+    get_or_create_proxy_group_id
 )
 
-
```
Avantol13: need this newline, two lines between classes and functions

fence/blueprints/storage_creds/google.py
```diff
@@ -52,16 +53,21 @@ def get(self):
         """
         client_id = current_token.get("azp") or None
         user_id = current_token["sub"]
+        username = (current_token
```
Avantol13:
```
        username = (
            current_token
            .get("context")
            .get("user")
            .get("name")
        )
```

fence/resources/google/utils.py
```diff
@@ -276,3 +279,128 @@ def create_service_account(client_id, user_id, username, proxy_group_id):
         flask.abort(
             404, 'Could not find Google proxy group for current user in the '
             'given token.')
+
+
+def get_or_create_proxy_group_id():
+
```
Avantol13: add docstrings for these new functions

fence/resources/google/utils.py
```diff
@@ -276,3 +279,128 @@ def create_service_account(client_id, user_id, username, proxy_group_id):
...
+    proxy_group_id = get_proxy_group_id()
+    if not proxy_group_id:
+        user_id=current_token['sub']
```
Avantol13: needs some spaces around `=`

fence/resources/google/utils.py
```diff
@@ -276,3 +279,128 @@ def create_service_account(client_id, user_id, username, proxy_group_id):
...
+        username = (
+            current_token.get('context', {})
+            .get('user', {})
+            .get('name')
```
Avantol13: change you change `.get('name')` to `.get('name', '')` so if for some reason the name is empty, we pass an empty string and not `None`

fence/resources/google/utils.py
```diff
@@ -276,3 +279,128 @@ def create_service_account(client_id, user_id, username, proxy_group_id):
...
+            current_session
+            .query(User)
+            .filter(User.id == current_token['sub'])
+            .first())
+        proxy_group_id = user.google_proxy_group_id
+
+    return proxy_group_id
+
+
+def create_proxy_group(user_id, username):
```
Avantol13: I don't think we need this, we want to encourage the user of the get_or_create so I'd just not include this at all

fence/resources/google/utils.py
```diff
@@ -276,3 +279,128 @@ def create_service_account(client_id, user_id, username, proxy_group_id):
...
+            .get('name')
+        )
+        proxy_group_id = create_proxy_group(user_id, username).id
+
+    return proxy_group_id
+
+
+def get_proxy_group_id():
+
+    proxy_group_id = (
```
Avantol13: you can replace instances where you do this with the function `get_users_proxy_group_from_token`

fence/scripting/fence_create.py
```diff
@@ -320,16 +320,16 @@ def google_init(db):
...
-                    response = g_mgr.create_proxy_group_for_user(
-                        user.id, user.username)
-                except Exception as exc:
-                    raise Exception(
-                        'Unable to create proxy group for user {} with id: {}. '
-                        'Google API Error: {}'
-                        .format(user.username, user.id, exc))
 
-                group = response["group"]
+                group = g_mgr.create_proxy_group_for_user(
```
Avantol13: We shouldn't need this google_init function to create proxy groups at all now... you can leave the function but don't do any proxy group or service account creation

fence/sync/sync_users.py
```diff
@@ -711,8 +711,6 @@ def _sync(self, sess):
     driver = SQLAlchemyDriver(DB)
```
Avantol13: You can take out this whole block

fence/resources/google/utils.py
Avantol13: or, better yet, keep it separated but prefix the function name with a `_` to indicate that it's "internal" and hopefully discourage people from using it

benJrohrer: I went with the latter suggestion.

fence/scripting/fence_create.py
benJrohrer: Does this function do anything other than create a proxy group and a service account? From what I can tell thats it in entirety.

PR #257, opened 2018-06-15
PXD-278: feat(assume-role): cross accounts assuming role

thanh-nguyen-dang:
Add across accounts presigned url
- Change structure of S3_BUCKETS to have type of bucket
- if internal -> refer to key AWS_CREDENTIALS
- if external -> refer to key in ASSUME_ROLES, then get reference to AWS_CREDENTIALS
r @zflamig @philloooo @rudyardrichter 

related prs:
https://github.com/uc-cdis/cdis-python-utils/pull/29

### DIFF
--- a/fence/local_settings.example.py
+++ b/fence/local_settings.example.py
@@ -110,10 +110,21 @@ AWS_CREDENTIALS = {
     }
 }
 
+ASSUMED_ROLES = {
+    'arn:aws:iam::role1': 'CRED1'
+}
+
 S3_BUCKETS = {
-    'bucket1': 'CRED1',
-    'bucket2': 'CRED2',
-    'bucket3': 'CRED1'
+    'bucket1': {
+        'cred': 'CRED1',
+    },
+    'bucket2': {
+        'cred': 'CRED2',
+    },
+    'bucket3': {
+        'cred': 'CRED1',
+        'role-arn': 'arn:aws:iam::role1',
+    }
 }
 
 #: Confiure which identity providers this fence instance can use for login.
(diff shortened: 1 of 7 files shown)


fence/local_settings.example.py
```diff
@@ -110,10 +110,23 @@
...
+    'arn:aws:iam::role1': 'CRED1'
+}
+
 S3_BUCKETS = {
-    'bucket1': 'CRED1',
-    'bucket2': 'CRED2',
-    'bucket3': 'CRED1'
+    'bucket1': {
+        'cred': 'CRED1',
+        'type': 'internal'
```
philloooo: I don't think we should add a parameter to differentiate bucket type by external or internal and add  a `ASSUMED_ROLES` config.
I think we should add an optional `role` key to `AWS_CREDENTIALS`, so if
```
AWS_CREDENIALS = {
  'CREDS1':  { 'aws_access_key': 'XXX', 'aws_secret_key': 'XXX', 'role': 'abc'},
  'CREDS2':  { 'aws_access_key': 'XXX', 'aws_secret_key': 'XXX'}
}
```
then it tells us creds1 need to assume role and creds2 doesn't need that.

thanh-nguyen-dang: I also think of that but in a different way.
Doing your suggestion requires repeating `aws_access_key` and `aws_secret_key` for different roles/bucket.

I think that we can do a similar thing in `S3_BUCKET`
```
S3_BUCKETS = {
    'bucket1': {
        'cred': 'CRED1',
    },
    'bucket2': {
        'cred': 'CRED2',
    },
    'bucket3': {
        'cred': 'CRED1',
        'role-arn': 'arn:aws:iam::role1',
    }
}
```

philloooo: cool, that works too

PR #271, opened 2018-06-22
Removed creation of proxy groups / service accounts from GET 

benJrohrer:
*added tests


fence/blueprints/storage_creds/google.py
```diff
@@ -65,14 +65,12 @@ def get(self):
...
-                get_or_create_service_account(
-                    client_id=client_id, user_id=user_id,
-                    username=username, proxy_group_id=proxy_group_id))
-
-            keys = g_cloud_manager.get_service_account_keys_info(
-                service_account.google_unique_id)
+            service_account = get_service_account(client_id, user_id)
+
+            keys = []
+            if (proxy_group_id and service_account):
```
Avantol13: proxy_group_id is no longer defined. need to get it above

PR #279, opened 2018-06-26
PXD-992 feat(google-link): add linked google account expiration to id token

macintoshpie:
Add expiration time for liked google account into the id_token


fence/resources/google/utils.py
```diff
@@ -198,6 +199,31 @@ def get_linked_google_account_email(user_id):
     return email
 
 
+def get_linked_google_account_info(user_id):
+    account_info = {
+        'linked_google_email': None,
+        'linked_google_expires': 0
```
Avantol13: let's keep the names `linked_google_account` and `linked_google_account_exp`

tests/credentials/google/test_credentials.py
```diff
@@ -386,6 +386,87 @@ def test_google_bucket_access_existing_proxy_group(
...
+        path, data=data,
+        headers={'Authorization': 'Bearer ' + encoded_credentials_jwt})
+
+    assert (
+        google_storage_client_mocker.delete_bucket_acl.called is True
+    )
+    assert response.status_code == 200
+
+
+def test_google_bucket_access_existing_proxy_group(
```
Avantol13: this stuff doesn't look related to your changes, perhaps some extra stuff got added during a merge?

fence/resources/google/utils.py
macintoshpie: I think we should keep it as linked_google_email unless we want to change the way some of the other functions use that variable. eg:
https://github.com/uc-cdis/fence/blob/ec62333fc8e41336819025ddc1d490cf7c3d819f/fence/jwt/token.py#L414

fence/resources/google/utils.py
```diff
@@ -187,15 +188,73 @@ def create_google_access_key(client_id, user_id, username, proxy_group_id):
     return key, service_account
 
 
+def get_linked_google_account(user_id):
```
Avantol13: since you're just using this internally, let's change the name to _get_linked_google_account and remove the checks. e.g. just return
```
 (
            current_session.query(UserGoogleAccount)
            .filter(UserGoogleAccount.user_id == user_id).first()
       )
```

fence/resources/google/utils.py
```diff
@@ -187,15 +188,73 @@ def create_google_access_key(client_id, user_id, username, proxy_group_id):
...
+                .filter(
+                    UserGoogleAccountToProxyGroup
+                    .user_google_account_id == g_account.id).first()
+            )
+            if g_account_to_proxy_group:
+                google_account_exp = g_account_to_proxy_group.expires
+    return google_account_exp
+
+
+def get_linked_google_account_info(user_id):
```
Avantol13: I'm thinking we should just remove this and use the above two functions. seems like duplicated functionality. when passing in the kwargs, just specify the `linked_google_email` and `linked_google_account_exp`

tests/link/test_link_id_token.py
```diff
@@ -0,0 +1,64 @@
...
+
+    original_expiration = 1000
+    google_account = 'some-authed-google-account@gmail.com'
+
+    # add google account and link
+    existing_account = UserGoogleAccount(email=google_account, user_id=user_id)
+    db_session.add(existing_account)
+    db_session.commit()
+    g_account_access = UserGoogleAccountToProxyGroup(
+            user_google_account_id=existing_account.id,
```
Avantol13: style nitpick: indentation should be less here

tests/link/test_link_id_token.py
```diff
@@ -0,0 +1,64 @@
...
+            proxy_group_id=proxy_group_id,
+            expires=original_expiration
+    )
+    db_session.add(g_account_access)
+    db_session.commit()
+
+    # get link from database
+    account_in_proxy_group = (
```
Avantol13: I don't think this check is necessary since you're adding this above

PR #290, opened 2018-06-29
PXD-825 feat(consent): add consent page

qingyashu:
FN260 - Implement the new UI page for OIDC


fence/templates/oauthorize.html
```diff
@@ -17,68 +17,133 @@
...
-          <ul>
-            {% for resource in resource_description %}
-              <li>{{resource}}</li>
-            {% endfor %}
-          </ul>
-         <a id="yes" class='yes'>Yes </a>
-         <a id="no" class='no'>No</a>
-      </form>
-    </section>
+    <img class="auth-kidfirst-logo" src="./kidfirst-logo.svg" />
```
zflamig: We can't make this KF specific, needs to be generic with overrides for adding appropriate logos, etc.

Dockerfile
```diff
@@ -69,5 +69,5 @@ RUN ln -s /fence/wsgi.py /var/www/fence/wsgi.py \
 
 EXPOSE 80
 WORKDIR /var/www/fence/
-
+COPY custom/logo/$APP-logo.svg fence/template/logo.svg
 CMD rm -f /var/run/apache2/apache2.pid && /usr/sbin/apache2ctl -D FOREGROUND
```
qingyashu: Remove the COPY,
Change CMD to something like:
```
CMD /fence/dockerrun.bash
```
, and add a `dockerrun.bash` script that looks something like:
```
#!/bin/bash

if [[ -n "$APP" && -f "/fence/custom/logo/$APP-logo.svg" ]]; then
  cp /fence/custom/logo/$APP-logo.svg /fence/template/logo.svg
fi
rm -f /var/run/apache2/apache2.pid && /usr/sbin/apache2ctl -D FOREGROUND
```

qingyashu: Test build with:
```
docker build -t fence:my_fence .
docker run -it --rm --entrypoint=bash fence:my_fence
/fence/dockerrun.bash
```

qingyashu: BTW - this is Reuben - I really hate that we're going down a road where we'll be submitting PR's to customize fence for every new client we get by adding a logo - probably doing more to come.  This is a bad road to go down.
We could wire up fence to load css and images from the portal:
     `/resources/fence/...`
, so at least our per-commons customization is all in one repo.  We need to move toward setting up gitops where we have a git repo per commons to track customizations like this along with the history of configuration changes.  
:-[

fence/templates/oauthorize.html
```diff
@@ -17,68 +17,152 @@
...
+            <div class="auth-list">
+              <ul>
+                {% for resource in resource_description %}
+                  <li class="auth-list-elem">{{resource}}</li>
+                {% endfor %}
+              </ul>
+            </div>
+            <div class="auth-action-line">
+              <div class="introduction">
+                You will be redirected to NIH login after the authorization.
```
zflamig: Lets make this:
You will be redirected to the identity provider after the authorization.

PR #291, opened 2018-07-02
PXD-1004 Fix/stale access token

Avantol13:
Check for valid `access_token` when opening the session. Verify that the user identifier provided in the `access_token` matches the one in the current session.

### DIFF
--- a/fence/resources/user/user_session.py
+++ b/fence/resources/user/user_session.py
@@ -55,8 +55,7 @@ class UserSession(SessionMixin):
             try:
                 jwt_info = validate_jwt(
                     session_token,
-                    public_key=default_public_key(),
-                    aud={'fence'},
+                    aud={'fence'}
                 )
             except JWTError:
                 # if session token is invalid, create a new
@@ -111,7 +110,7 @@ class UserSession(SessionMixin):
         """
         get a value from session json
         """
-        return self.session_token["context"].get(key, *args)
+        return self.session_token['context'].get(key, *args)
 
     def pop(self, key, default):
         return self.session_token['context'].pop(key, default)
@@ -121,7 +120,7 @@ class UserSession(SessionMixin):
         clear current session
         """
         self._encoded_token = None
-        self.session_token = {"context": {}}
+        self.session_token = {'context': {}}
 
     def clear_if_expired(self, app):
         if self._encoded_token:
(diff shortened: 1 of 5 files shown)


fence/resources/user/user_session.py
```diff
@@ -228,6 +243,47 @@ def save_session(self, app, session, response):
...
+        try:
+            user = get_current_user(flask_session=session)
+        except Unauthorized:
+            return None
+
+        # check that the current user is the one from the session and access_token
+        user_sess_id = _get_user_id_from_session(session)
+        token_user_id = _get_user_id_from_access_token(valid_access_token)
+
+        if user.id != user_sess_id:
```
rudyardrichter: can condense to `if user.id != user_sess_id and user.username != user_sess_id`; same with next conditional

tests/session/test_session.py
```diff
@@ -170,10 +180,126 @@ def test_invalid_session_cookie(app):
...
+        )
+        assert response.status_code == 200
+        response_user_id = response.json.get('user_id') or response.json.get('sub')
+        assert response_user_id == test_user_a['user_id']
+
+        user_id = valid_access_token.get('user_id') or valid_access_token.get('sub')
+        assert test_user_a['user_id'] == int(user_id)
+
+
+def _get_cookies_from_response(response):
```
rudyardrichter: You could use `werkzeug.http.parse_cookie` here, werkzeug is required by flask so should be importable already (without changing fence requirements), and saves you the raw cookie parsing

fence/resources/user/user_session.py
```diff
@@ -267,3 +323,35 @@ def _create_access_token_cookie(app, session, response, user):
...
+    """
+    Get user's identifier from the access token claims. It could be their id or
+    username since both are unique.
+    """
+    token_user_id = access_token.get('sub')
+    if token_user_id:
+        try:
+            token_user_id = int(token_user_id)
+        except ValueError:
+            # if we can't cast to an int, don't. could be username
```
rudyardrichter: I think with the current code, since from userdatamodel the `User.id` is an `int` and that's what gets put in the access token, we always expect an integer here/it's a problem if the ID is not an int. So probably either this should not catch the exception, or return None

tests/session/test_session.py
Avantol13: the headers are already parsed behind the scenes in flask and accessible as a `werkzeug.datastructures.Headers` object through `response.headers`. That's list of tuples of the headers. I grab all the `Set-Cookie` ones for the cookies and passed those into `werkzeug.http.parse_cookie`.

```
cookies = [
        werkzeug.http.parse_cookie(header[1])
        for header in response.headers.iteritems()
        if header[0] == 'Set-Cookie'
    ]
```

PR #297, opened 2018-07-05
Feat/google service account endpoints

Avantol13:
* Provide a skeleton (with some muscle) for the `/google/service_accounts` endpoints
* Update swagger API docs
* Few tests to check some endpoints and add helper functions for future work


fence/resources/google/access_utils.py
```diff
@@ -0,0 +1,334 @@
...
+    return []
+
+
+def do_all_users_have_access_to_project(users, project_auth_id):
+    # users will be list of user_ids
+    # TODO actually check
+    return False
+
+
+class ValidityInfo(object):
```
rudyardrichter: For implementing dictionary-like classes, one of these may be a good idea:
- https://docs.python.org/2/library/userdict.html
- https://docs.python.org/2.7/library/collections.html#collections.Mapping

Avantol13: Updated inheritance to be collections.Mapping

fence/resources/google/access_utils.py
```diff
@@ -0,0 +1,342 @@
...
+
+def service_account_has_external_access(service_account):
+    return True
+
+
+def is_service_account_from_google_project(service_account, google_project):
+    return False
+
+
+def is_user_member_of_all_google_projects(user_id, google_project_ids):
```
rudyardrichter: Just curious, is there a reason some of these are separated out to this other file and some (like `_can_user_manage_service_account`) are left in the blueprint file? I would probably suggest to put them all together if there's no reason not to

fence/resources/google/access_utils.py
```diff
@@ -0,0 +1,342 @@
...
+
+    Args:
+        user_id (int): User identifier
+        google_project_ids (List(str)): List of unique google project ids
+
+    Returns:
+        bool: whether or not the given user is a member of ALL of the provided
+              Google project IDs
+    """
+    # TODO actually check
```
rudyardrichter: `NotImplemented`, etc.

PR #308, opened 2018-07-11
PXD-1021 Refactor session token

abgeorge7:
- Moved shared functionality (setting `flask.session` properties) to `login_user`
- Added `user_id` to `flask.session`
- Change `'sub'` field to be `user_id`
- Tests for `login_user` to make sure `flask.session` is being set properly


tests/login/test_login_user.py
```diff
@@ -0,0 +1,32 @@
+import flask
+from flask_sqlalchemy_session import current_session
+from fence.auth import login_user
+from fence.models import User, IdentityProvider
+
+def test_login_user_in_current_session(db_session):
+    email = "testuser@gmail.com"
```
Avantol13: Let's change the names of these tests to `test_login_user_already_in_db` and `test_login_new_user`.

Let's also add some docstrings describing the test, otherwise this PR looks great!! Something like this:
```
"""
Test that when a user already exists in the database and we log them in, the expected information is available in the session afterwards.
"""
```

PR #309, opened 2018-07-12
Feat/expired service account

giangbui:


fence/scripting/fence_create.py
```diff
@@ -542,6 +548,41 @@ def delete_users(DB, usernames):
...
+def delete_expired_service_accounts(DB):
+    """
+    Delete all expired service accounts.
+    """
+    import fence.settings
+    try:
+        cirrus_config.update(**fence.settings.CIRRUS_CFG)
+    except AttributeError:
+        # no cirrus config, continue anyway.
+        pass
```
Avantol13: We need cirrus to remove the service accounts so we shouldn't pass here. If you want to catch the error you can raise it with more information about cirrus needing to be configured

fence/scripting/fence_create.py
```diff
@@ -542,6 +548,41 @@ def delete_users(DB, usernames):
...
+        if len(records_to_delete):
+            with GoogleCloudManager() as manager:
+                for record in records_to_delete:
+                    try:
+                        manager.remove_member_from_group(
+                            record.service_account.email, record.access_group.email)
+                        session.delete(record)
+                        print('Removed expired service account: {}'.format(
+                            record.service_account.email))
+                    except HttpError as e:
```
Avantol13: we may want to be generic and just catch Exception here, in case there are other bugs in cirrus or some other reason it didn't get deleted we want to print that infor too. And it may be good to preface this error with something like: "ERROR: Could not delete service account. Details: {{error message}}"

tests/scripting/test_fence-create.py
```diff
@@ -230,3 +236,149 @@ def test_create_refresh_token_with_found_user(
         .first()
     )
     assert db_token is not None
+
+
+def setup_service_account_to_google_bucket_access_group(db_session):
```
Avantol13: you should preface this function name with `_` to suggest it's private use to this module

tests/scripting/test_fence-create.py
```diff
@@ -230,3 +236,149 @@ def test_create_refresh_token_with_found_user(
...
+        db_session
+        .query(ServiceAccountToGoogleBucketAccessGroup)
+        .all()
+    )
+    assert len(records) == 3
+
+    # call function to delete expired service account
+    delete_expired_service_accounts(app.config['DB'])
+    # check database. Expect 2 deleted
+    records = db_session.query(ServiceAccountToGoogleBucketAccessGroup).all()
```
Avantol13: you could also check that the remaining account is the one you expect

tests/scripting/test_fence-create.py
```diff
@@ -230,3 +236,149 @@ def test_create_refresh_token_with_found_user(
...
+    db_session.add(GoogleBucketAccessGroup(
+        bucket_id=bucket1.id, email='testgroup1@gmail.com',
+        privileges=['read_storage', 'write_storage']))
+    db_session.add(GoogleBucketAccessGroup(
+        bucket_id=bucket1.id, email='testgroup2@gmail.com',
+        privileges=['read_storage']))
+    db_session.commit()
+
+
+def test_delete_expired_service_accounts_with_one_fail(cloud_manager, app, db_session):
```
Avantol13: we should have another test like this one except both accounts are expired and the failure happens on the second service account. Then we check that we DID delete the first expired one

PR #311, opened 2018-07-12
PXD-250 is_valid_service_account_type

benJrohrer:
*added access_utils.py
*added is_valid_service_account_type


fence/resources/google/access_utils.py
```diff
@@ -0,0 +1,16 @@
...
+    "USER_MANAGED",
+]
+
+
+def is_valid_service_account_type(project_id, account_id):
+
+    try:
+        with GoogleCloudManager(project_id) as g_mgr:
+            return g_mgr.get_service_account_type(account_id) in ALLOWED_SERVICE_ACCOUNT_TYPES
+    except:
```
Avantol13: should catch the generic `Exception` here

fence/resources/google/access_utils.py
```diff
@@ -0,0 +1,16 @@
+from cirrus import GoogleCloudManager
+
+ALLOWED_SERVICE_ACCOUNT_TYPES = [
+    "COMPUTE_ENGINE_DEFAULT",
```
Avantol13: these should be abstracted in cirrus and imported here so we don't have to maintain this string

fence/resources/google/access_utils.py
```diff
@@ -0,0 +1,19 @@
...
+    USER_MANAGED_SERVICE_ACCOUNT,
+)
+ALLOWED_SERVICE_ACCOUNT_TYPES = [
+    COMPUTE_ENGINE_DEFAULT_SERVICE_ACCOUNT,
+    USER_MANAGED_SERVICE_ACCOUNT,
+]
+
+
+def is_valid_service_account_type(project_id, account_id):
+
```
Avantol13: docstring please :D

fence/resources/google/access_utils.py
```diff
@@ -0,0 +1,19 @@
+from cirrus import GoogleCloudManager
+from cirrus.google_cloud import (
+    COMPUTE_ENGINE_DEFAULT_SERVICE_ACCOUNT,
+    USER_MANAGED_SERVICE_ACCOUNT,
+)
+ALLOWED_SERVICE_ACCOUNT_TYPES = [
```
Avantol13: style nitpick: newline between these

tests/google/test_access_utils.py
```diff
@@ -0,0 +1,36 @@
...
+)
+from cirrus.google_cloud import (
+    COMPUTE_ENGINE_DEFAULT_SERVICE_ACCOUNT,
+    COMPUTE_ENGINE_API_SERVICE_ACCOUNT,
+    GOOGLE_API_SERVICE_ACCOUNT,
+    USER_MANAGED_SERVICE_ACCOUNT,
+)
+
+
+def test_is_valid_service_account(cloud_manager):
```
Avantol13: I (personally) would break this into four tests, testing `valid` vs `not_valid` for the different types. The more separation in tests, the better. so if something goes wrong, we can narrow it down. this is sorta minor and I'm nitpicking at this point (though I usually nitpick :P) I'm not too worried about this one if you want to leave it

fence/resources/google/access_utils.py
```diff
@@ -0,0 +1,32 @@
...
+
+    Returns:
+        Bool: True if service acocunt type is allowed as defined
+        in ALLOWED_SERVICE_ACCOUNT_TYPES
+    """
+    try:
+        with GoogleCloudManager(project_id) as g_mgr:
+            return g_mgr.get_service_account_type(account_id) in ALLOWED_SERVICE_ACCOUNT_TYPES
+    except Exception:
+        return False
```
Avantol13: we could also log the error for debugging here with `flask.current_app.logger.debug('validity of Google service account {} (google project: {}) type determined False due to error. Details: {}'.format(account_id, project_id, exc)` or something like that. May help us in the future when double checking/ debugging this stuff :D

tests/google/test_access_utils.py
```diff
@@ -0,0 +1,45 @@
...
+from cirrus.google_cloud import (
+    COMPUTE_ENGINE_DEFAULT_SERVICE_ACCOUNT,
+    COMPUTE_ENGINE_API_SERVICE_ACCOUNT,
+    GOOGLE_API_SERVICE_ACCOUNT,
+    USER_MANAGED_SERVICE_ACCOUNT,
+)
+
+
+def test_is_valid_service_account_type_compute_engine_default(cloud_manager):
+
```
Avantol13: some quick docstrings here please. `"""Test that.. blah blah"""`

PR #316, opened 2018-07-16
feat(bucket-creation): script for creating a bucket for access logs

Avantol13:


fence/scripting/fence_create.py
```diff
@@ -829,6 +822,49 @@ def create_or_update_google_bucket(
...
+def create_google_logging_bucket(
+        name, storage_class=None, google_project_id=None):
+    import fence.settings
+    cirrus_config.update(**fence.settings.CIRRUS_CFG)
+
+    google_project_id = google_project_id or cirrus_config.GOOGLE_PROJECT_ID
+
+    # determine project where buckets are located
+    # default to same project, try to get storage creds project from key file
+    storage_creds_project_id = _get_storage_project_id() or google_project_id
```
rudyardrichter: this looks backwards—maybe it's supposed to go this way? just checking

fence/scripting/fence_create.py
```diff
@@ -829,6 +822,49 @@ def create_or_update_google_bucket(
...
+
+def _get_storage_project_id():
+    """
+    Determine project where buckets are located.
+    Try to get storage creds project from key file
+    """
+    storage_creds_project_id = None
+    storage_creds_file = cirrus_config.configs['GOOGLE_STORAGE_CREDS']
+    if os.path.exists(storage_creds_file):
+        with open(storage_creds_file) as file:
```
rudyardrichter: technically `file` is a builtin, so if you're super picky about avoiding these (like I am...) I'd suggest `with open(storage_creds_file) as f` instead

Avantol13: oh, good point. missed that, will correct. I'm picky too :D overriding builtins scares me

fence/scripting/fence_create.py
Avantol13: yup, good catch. fixed in f5e119b94dcc1b417c52e3b17f2dab44f9592884

PR #322, opened 2018-07-18
Feat/has parent organization

benJrohrer:
Implements google_project_has_parent_organization


fence/resources/google/access_utils.py
```diff
@@ -41,7 +41,7 @@ def can_user_manage_service_account(user_id, account_id):
 
 
 def google_project_has_parent_org(google_project):
-    raise NotImplementedError('Functionality not yet available...')
+    return google_project.has_parent_organization()
```
Avantol13: where is this function? the google_project here is a project id, so I think you need to use GoogleCloudManager here

fence/resources/google/access_utils.py
```diff
@@ -40,8 +40,17 @@ def can_user_manage_service_account(user_id, account_id):
         user_id, [service_account_project])
 
 
-def google_project_has_parent_org(google_project):
-    raise NotImplementedError('Functionality not yet available...')
+def google_project_has_parent_org(project_id):
+
```
Avantol13: docstring pls :)

fence/resources/google/access_utils.py
```diff
@@ -40,8 +40,17 @@ def can_user_manage_service_account(user_id, account_id):
...
+    try:
+        with GoogleCloudManager(project_id) as prj:
+            return prj.has_parent_organization()
+    except Exception as exc:
+        flask.current_app.logger.debug((
+            'Could not determine if Google project (id: {}) has parent org'
+            'due to error (Details: {})'.
+            format(project_id, exc)
+        ))
```
Avantol13: we need to return `False` here to be explicit

PR #328, opened 2018-07-19
Feat/valid service accounts

benJrohrer:


fence/resources/google/access_utils.py
```diff
@@ -162,6 +162,43 @@ def do_all_users_have_access_to_project(user_ids, project_auth_id):
...
+
+            members = prj.get_project_membership()
+
+            for mem in members:
+                if mem.member_type == GooglePolicyMember.SERVICE_ACCOUNT:
+                    if mem not in service_accounts:
+                        return False
+
+            return True
+    except Exception as exc:
```
rudyardrichter: What exceptions would happen from the above code, and could this be more specific than a generic `Exception`?

fence/resources/google/access_utils.py
```diff
@@ -162,6 +162,43 @@ def do_all_users_have_access_to_project(user_ids, project_auth_id):
...
+    Args:
+        project_id(str): unique id of project
+    ReturnsL
+        Bool: True iff all service accounts are valid
+    """
+    try:
+        with GoogleCloudManager(project_id) as prj:
+            service_accounts = prj.get_all_service_accounts()
+
+            for acc in service_accounts:
```
rudyardrichter: not significant, but this is one-linerizeable:
```python
if any([service_account_has_external_access(acc) for acc in service_accounts]):
    return False
```

fence/resources/google/utils.py
```diff
@@ -511,9 +511,22 @@ def get_project_access_from_service_accounts(service_accounts):
...
+def get_service_account_ids_from_google_project(project_id):
+
+    try:
+        account_ids = []
+        with GoogleCloudManager(project_id) as prj:
+            accounts = prj.get_all_service_accounts()
+            for acc in accounts:
+                account_ids.append(acc["email"])
+        return account_ids
+    except Exception as exc:
```
rudyardrichter: Also here, use more specific exception if possible

fence/resources/google/utils.py
```diff
@@ -511,9 +511,22 @@ def get_project_access_from_service_accounts(service_accounts):
...
-    raise NotImplementedError('Functionality not yet available...')
+def get_service_account_ids_from_google_project(project_id):
+
+    try:
+        account_ids = []
+        with GoogleCloudManager(project_id) as prj:
+            accounts = prj.get_all_service_accounts()
+            for acc in accounts:
+                account_ids.append(acc["email"])
+        return account_ids
```
rudyardrichter: can condense this whole block to:
```python
with GoogleCloudManager(project_id) as prj:
    return [acc['email'] for acc in prj.get_all_service_accounts()]
```

fence/resources/google/access_utils.py
```diff
@@ -126,8 +126,8 @@ def service_account_has_external_access(service_account):
     raise NotImplementedError('Functionality not yet available...')
 
 
-def is_service_account_from_google_project(service_account, google_project):
-    raise NotImplementedError('Functionality not yet available...')
+def is_service_account_from_google_project(service_account, project_id):
+    return service_account in GoogleCloudManager(project_id).get_all_service_accounts()
```
rudyardrichter: I think this function is supposed to receive string arguments, from the looks of this usage:
https://github.com/uc-cdis/fence/blob/e4bea9f67e23aaff78729da847bdc02b3bb88602/fence/resources/google/validity.py#L290-L293
so you probably want either the `'name'` or `'uniqueId'` fields out of the `get_all_service_accounts()` return.

rudyardrichter: On that note, I think it would be good to give this function a docstring listing the argument types so this is very clear

fence/resources/google/access_utils.py
rudyardrichter: (or if you're feeling functional:
```python
if any(map(service_account_has_external_access, service_accounts)):
    return False
```
)

benJrohrer: I was/am always feeling functional! I wasn't really sure what people thought about this stylistically/readability-wise. I personally like it, but I'm not sure how other people feel. Also are there any performance considerations? Particularly that map will apply the function to all service accounts, whereas the original will short-circuit. any() short-circuits, but only after map applies to all elements of the list. Maybe that's irrelevant here though, I'm not sure of the scope.

rudyardrichter: Oh, true. `map` is a generator in 3 but not 2, so I guess you would have to use something from `functools` or similar in this case. Which reminds me you should omit the `[]` for a generator, in any case:
```python
if any(service_account_has_external_access(acc) for acc in service_accounts):
```

Yeah, I would also like to hear what others think about this. I like it as well for the conciseness.

benJrohrer: Oh! I didn't know map was a generator in 3. That's cool. I will go with the map-less implementation for now, as that's perhaps more universally compatible.

fence/resources/google/access_utils.py
benJrohrer: The exception will come from initializing GoogleCloudManager with an invalid project id. Alex instructed me to use the generic exception. My guess is that there would be some Google Cloud API Error, so I will use the base GoogleAPIError.

fence/resources/google/access_utils.py
```diff
@@ -151,8 +153,24 @@ def service_account_has_external_access(service_account):
...
 
-def is_service_account_from_google_project(service_account, google_project):
-    raise NotImplementedError('Functionality not yet available...')
+def is_service_account_from_google_project(service_account, project_id):
+    """
+    Checks if service account is among project's service acounts
+
+    Args:
+        service_account(str): uniqueId of service account
+        project_id(str): uniqueId of Google Cloud Project
```
rudyardrichter: missing `Return:` section

fence/resources/google/access_utils.py
```diff
@@ -151,8 +153,24 @@ def service_account_has_external_access(service_account):
...
+    """
+    try:
+        return (service_account in
```
rudyardrichter: for readability & formatting I suggest breaking this up:
```python
service_accounts = (
    acc.get('uniqueId')
    for acc in GoogleCloudManager(project_id).get_all_service_accounts()
)
return service_account in service_accounts
```

fence/resources/google/access_utils.py
```diff
@@ -187,6 +205,47 @@ def do_all_users_have_access_to_project(user_ids, project_auth_id):
...
+                sa.get('email'))
+                for sa in service_accounts]
+
+            for mem in members:
+                if mem.member_type == GooglePolicyMember.SERVICE_ACCOUNT:
+                    if mem not in sa_members:
+                        return False
+
+            return True
+    except GoogleCloudError as exc:
```
rudyardrichter: it looks like the `try` block could be tightened up a bit as well, since `prj` is last used to assign `members`, so the `sa_members = ...` and below don't need to be in the `try`/`except` (I think?)

PR #330, opened 2018-07-20
Feat/account external access

giangbui:
PXD-251


fence/resources/google/access_utils.py
```diff
@@ -76,7 +78,30 @@ def is_valid_service_account_type(project_id, account_id):
...
+    """
+    with GoogleCloudManager() as g_mgr:
+        response = g_mgr.get_service_account_policy(service_account)
+        if response.status_code != 200:
+            raise GoogleAPIError('Unable to get IAM policy for service account {}\n{}.'
+                                .format(service_account, response.json()))
+        json_obj = response.json()
+        # In the case that a service account does not have any role, Google API
+        # returns a json object without bindings key
+        if json_obj.has_key('bindings'):
```
Avantol13: `has_key` is removed in python3, so it'd be more forwards-compatible to use the syntax `if 'bindings' in json_obj`, which does the same thing

giangbui: Good to know. Fixed in 96a4640d799de0cd3e6e20e757406cdd3aedd39e

PR #337, opened 2018-07-20
feat(exec-num): adding namespace selection by executor number

stevencui729:


Jenkinsfile
```diff
@@ -59,13 +59,21 @@ pipeline {
...
     }
-    stage('ModifyManifest') {
+    stage('SelectNamespace') {
       steps {
         script {
           String[] namespaces = ['qa-bloodpac', 'qa-brain', 'qa-kidsfirst', 'qa-niaid']
-          randNum = new Random().nextInt() % namespaces.length
-          env.KUBECTL_NAMESPACE = namespaces[randNum]
+          int modNum = namespaces.length/2
+          int randNum = new Random().nextInt(modNum) + ((env.EXECUTOR_NUMBER as Integer) * 2)
```
frickjack: Would you wrap this whole thing in a modulo operator, so it will still run if we add executors?
     `(... number) % namespaces.length`

Jenkinsfile
```diff
@@ -59,13 +59,21 @@ pipeline {
...
+          env.KUBECTL_NAMESPACE = namespaces[randNum]
+          println "selected namespace $env.KUBECTL_NAMESPACE on executor $env.EXECUTOR_NUMBER"
+        }
+      }
+    }
+    stage('ModifyManifest') {
+      steps {
+        script {
           dirname="$env.KUBECTL_NAMESPACE"+'.planx-pla.net'
           service = "$env.JOB_NAME".split('/')[1]
```
frickjack: You can get the domain associated with the namespace by running something like:
     `kubectl --namespace $env.KUBECTL_NAMESPACE get configmap global -ojsonpath=.data.hostname`

PR #339, opened 2018-07-20
feat(error-page): add app name into error page

Avantol13:


fence/error_handler.py
```diff
@@ -12,6 +12,7 @@
 def get_error_response(error):
     details, status_code = get_error_details_and_status(error)
     support_email = flask.current_app.config.get('SUPPORT_EMAIL_FOR_ERRORS')
+    app_name = flask.current_app.config.get('APP_NAME', '')
```
rudyardrichter: Maybe default to `'Gen3 Auth Service'` or similar?

Avantol13: I think the default image says Gen3 Commons so I'll put that in there

PR #340, opened 2018-07-23
PXD-1193 formatted error page

abgeorge7:
Error page based on spec- left CSS in the same file because I was having issues importing it into the html file.


fence/templates/error.html
```diff
@@ -1,61 +1,147 @@
...
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js"></script>
+    <script src="/api/static/js/bootstrap.min.js"></script>
+  </head>
+
+  <body>
+    <main>
+      <img class="error-page__logo" src="{{url_for('static', filename='img/logo.svg') }}" />
+      <div class="error-page">
+        <div class="error-page__banner">
+          <h1>{{app_name}}</h1>
```
abgeorge7: Just a note that `app_name` is blank when I test it on my local machine - not sure if that's expected or not

PR #349, opened 2018-07-26
Feat/get service accounts

giangbui:
PXD-249

### DIFF
--- a/fence/resources/google/access_utils.py
+++ b/fence/resources/google/access_utils.py
@@ -199,7 +199,7 @@ def do_all_users_have_access_to_project(user_ids, project_auth_id):
     return True
 
 
-def do_get_service_account_from_google_project(project_id, service_account_id):
+def do_get_service_account_from_google_project(project_id):
     """
     Get service account given project id and service account id
 
@@ -210,12 +210,19 @@ def do_get_service_account_from_google_project(project_id, service_account_id):
     Returns:
         str: json string representing service account
     """
+    service_account_jsons = []
     with GoogleCloudManager(project_id) as g_mgr:
-        response = g_mgr.get_service_account(service_account_id)
-        if response.status_code != 200:
-            raise GoogleAPIError('Unable to get service account {}\n{}.'
-                                 .format(service_account_id, response.json()))
-        return response.json()
+        service_accounts = g_mgr.get_all_service_accounts()
+        for account in service_accounts:
+            account_id = account.get('name', '').split('/')[-1]
+            response = g_mgr.get_service_account(account_id)
+            if response.status_code != 200:
+                raise GoogleAPIError('Unable to get service account {}\n{}.'
+                                     .format(account_id, response.json())
+                                     )
+            service_account_jsons.append(response.json())
+
+    return service_account_jsons
 
 
 # TODO this should be in cirrus rather than fence...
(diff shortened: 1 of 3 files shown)


fence/resources/google/access_utils.py
```diff
@@ -176,17 +176,55 @@ def is_user_member_of_all_google_projects(user_id, google_project_ids):
...
+    Returns:
+        bool: whether all users have access to the google project
+    """
     for user_id in user_ids:
-        access_privillege = current_session.query(AccessPrivilege).filter(
-            AccessPrivilege.user_id == user_id and AccessPrivilege.project_id == project_auth_id).first()
+        access_privillege = (
+                current_session
+                .query(AccessPrivilege)
+                .filter(AccessPrivilege.user_id == user_id and AccessPrivilege.project_id == project_auth_id)
```
Avantol13: the `project_auth_id` is `Project.auth_id`, not `Project.id`, so you'll have to adjust this query a bit. `AccessPrivilege.project_id` gives you a `Project.id`

fence/resources/google/access_utils.py
```diff
@@ -176,17 +176,55 @@ def is_user_member_of_all_google_projects(user_id, google_project_ids):
...
+
+    Returns:
+        str: json string representing service account
+    """
+    service_account_jsons = []
+    with GoogleCloudManager(project_id) as g_mgr:
+        service_accounts = g_mgr.get_all_service_accounts()
+        for account in service_accounts:
+            account_id = account.get('name', '').split('/')[-1]
+            response = g_mgr.get_service_account(account_id)
```
Avantol13: I don't think you need to call `get_service_account`. `get_all_service_accounts` gets the information for all the service accounts by doing just that. So I think you can just return the reponse from get_all_service_accounts

giangbui: The json returned by get_all_service_account is a list of elements with following format `projects/test-project/serviceAccounts/test-service-account@mail.com`. We need to parse it to get correct service account.

fence/resources/google/access_utils.py
giangbui: That code is not related to this pr. I just reformat it. Anyway I fixed it in 7ee6681223b9129b94532fd2c85cf2339b770cdb

fence/resources/google/access_utils.py
giangbui: hmm.. That function only returns the list of service account with general information? It was not my expectation. Adjusted in 7ee6681223b9129b94532fd2c85cf2339b770cdb

fence/resources/google/access_utils.py
```diff
@@ -176,17 +176,44 @@ def is_user_member_of_all_google_projects(user_id, google_project_ids):
...
+    Get service account given project id and service account id
+
+    Args:
+        project_id(str): google project id
+        service_account_id(str): service account id
+
+    Returns:
+        str: json string representing service account
+    """
+    with GoogleCloudManager(project_id) as g_mgr:
```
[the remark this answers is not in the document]
Avantol13: can you add a try except around https://github.com/uc-cdis/fence/blob/master/fence/blueprints/google.py#L123 where this eventually gets used? And the except should return/raise a 400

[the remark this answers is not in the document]
giangbui: Actually, Google API always return json object even user does not have access to the project or project name is wrong etc.. That is reason why we need to refactor cirrus in the future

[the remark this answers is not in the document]
Avantol13: the GoogleCloudManager.__init__ will fail for the project if it doesn't have the rights credentials I believe.... I could definitely be wrong though

Avantol13 (dismissed):

PR #353, opened 2018-07-30
PXD-1030 feat(monitoring-sa-endpoint): first revision of the endpoint

Avantol13:
* Return fence's configured Google Creds email (to be used to monitor user-registered google service accounts and projects)


fence/blueprints/google.py
```diff
@@ -435,6 +444,32 @@ def _get_project_access_error_status(validity_info):
 
 
 def _get_monitoring_account_email():
-    # TODO get monitoring service account from CIRRUS_CFG. Will be the service
-    #      accont email used for the fence service
-    raise NotImplementedError('Functionality not yet available...')
+    """
```
rudyardrichter: I feel like this might not need to be a separate function, since it looks to be used only in _get_monitoring_service_account_response, and is sufficiently simple/wouldn't need testing on its own

openapis/swagger.yaml
```diff
@@ -850,6 +850,10 @@ paths:
         removed, all access to data will be removed.
       operationId: adminGoogleServiceAccount
       responses:
+        '404':
+          description: |
```
rudyardrichter: just checking—the `|` is what you're after? (as opposed to `>` or `>-`)

fence/blueprints/google.py
rudyardrichter: (I guess I'm not sure though if this is going to be used elsewhere as well later on)

openapis/swagger.yaml
Avantol13: hmm.. I guess idk what the difference is. I thought | just meant use markdown

rudyardrichter: [it's complicated](http://symfony.com/doc/current/components/yaml/yaml_format.html)

rudyardrichter: [this answer](https://stackoverflow.com/a/21699210) is concise and quite helpful

fence/blueprints/google.py
Avantol13: good point, I'll consolidate

PR #363, opened 2018-08-02
PXD-260 Feat/update service account

giangbui:


fence/resources/google/access_utils.py
```diff
@@ -284,6 +288,212 @@ def get_service_account_email(id_from_url):
...
+    granting_project_ids = set()
+    for project_auth_id in project_access:
+        project = (
+            session.query(Project).
+            filter_by(auth_id=project_auth_id).
+            first()
+        )
+        if not project:
+            raise fence.errors.NotFound('There is no {} in Fence db'
+                                        .format(project_auth_id))
```
Avantol13: style nitpick: this linining up indention with previous part is technically against PEP8 and our style guide (full disclosure: I used to do this and I think it's like that a lot of places in cirrus). We're supposed to push the whole internal part to the next line like: 
```
            raise fence.errors.NotFound(
                'There is no {} in Fence db'.format(project_auth_id))
```
where you put the ending `)` is kinda up to you. But technically we're not supposed to hang our indents without a fresh line like above. if it's longer you just keep lining it up one indent in: 
```
raise Exception(
    'this is a really long line with so much stuff in it I need a new line '
    'so I just go right underneath it like this.')
```

giangbui: https://www.python.org/dev/peps/pep-0008/
According to PEP8, it should be aligned with opening delimiter. That one always make me very confused. I don't know which way is the best but we should follow the standard one. If we want to follow our own style, we should have a document about that

fence/resources/google/access_utils.py
```diff
@@ -284,6 +288,212 @@ def get_service_account_email(id_from_url):
...
+                raise GoogleAPIError('Can not add {} to group {}. Detail {}'
+                                     .format(service_account.email, access_group.email, exc))
+
+
+def _revoke_user_service_account_from_db(session, to_delete_project_ids, service_account):
+    """
+    Remove service account from GoogleBucketAccessGroup
+
+    Args:
+        sess(current_session): db session
```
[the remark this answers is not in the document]
giangbui: Did not notice since I copied from other code. Thx!

fence/resources/google/access_utils.py
```diff
@@ -284,6 +288,212 @@ def get_service_account_email(id_from_url):
...
+
+        for access_group in access_groups:
+            try:
+                with GoogleCloudManager() as g_manager:
+                    g_manager.remove_member_from_group(
+                        member_email=service_account.email,
+                        group_id=access_group.email
+                    )
+            except Exception as exc:
+                raise GoogleAPIError('Can not remove {} from group {}. Detail {}'
```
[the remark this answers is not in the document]
giangbui: The idea of re-raise the exception here is to have more insight or to provide a clearer message about the exception. The library provides several exceptions, we catch each of them or catch all depending on certain circumstances. Here, we just care about if we can remove/add member to a group or not. If we have any exception which means it fails, we just catch by our own exception with the message coming from the library.

fence/resources/google/access_utils.py
```diff
@@ -284,6 +288,212 @@ def get_service_account_email(id_from_url):
...
+        service_account (UserServiceAccount): user service account
+
+    """
+    for project_id in to_add_project_ids:
+        access_groups = _get_google_access_groups(session, project_id)
+
+        for access_group in access_groups:
+            try:
+                with GoogleCloudManager() as g_manager:
+                    g_manager.add_member_to_group(
```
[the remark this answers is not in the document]
giangbui: Oh. That makes sense. Thanks!

fence/resources/google/access_utils.py
```diff
@@ -284,6 +288,212 @@ def get_service_account_email(id_from_url):
...
+        access_groups = _get_google_access_groups(session, project_id)
+
+        for access_group in access_groups:
+            try:
+                with GoogleCloudManager() as g_manager:
+                    g_manager.remove_member_from_group(
+                        member_email=service_account.email,
+                        group_id=access_group.email
+                    )
+            except Exception as exc:
```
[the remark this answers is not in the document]
giangbui: That makes sense. Thanks!

fence/resources/google/access_utils.py
```diff
@@ -284,6 +288,212 @@ def get_service_account_email(id_from_url):
...
+        session.query(ServiceAccountAccessPrivilege).
+        filter_by(service_account_id=service_account.id).all()
+    )
+    accessed_project_ids = {ap.project_id for ap in accessed_projects}
+
+    granting_project_ids = set()
+    for project_auth_id in project_access:
+        project = (
+            session.query(Project).
+            filter_by(auth_id=project_auth_id).
```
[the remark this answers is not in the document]
giangbui: That makes me confused also but I think your suggestion is better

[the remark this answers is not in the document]
Avantol13: this appears elsewhere throughout these changes too, just don't want to comment on all of them

tests/google/conftest.py
```diff
@@ -204,29 +205,56 @@ def setup_data(db_session):
...
+
     db_session.add(ServiceAccountAccessPrivilege(project_id=project1.id, service_account_id=user.id))
     db_session.add(ServiceAccountAccessPrivilege(project_id=project2.id, service_account_id=user.id))
 
     access_grp = GoogleBucketAccessGroup(
         bucket_id=bucket.id, email='test@gmail.com'
     )
 
+    access_grp2 = GoogleBucketAccessGroup(
+        bucket_id=bucket2.id, email='test@gmail.com'
```
Avantol13: these should ideally have different email addresses, you're probably not testing that right now but if someone else comes along they may try to and get confused. each bucket access groups should have a different email

tests/google/test_access_utils.py
```diff
@@ -521,3 +519,94 @@ def test_remove_service_account_raise_GoogleAPI_exc(
...
+        cloud_manager.return_value.__enter__.
+        return_value.add_member_to_group.side_effect
+    ) = {}
+
+    service_account = (
+            db_session.
+            query(UserServiceAccount).
+            filter_by(email='test@gmail.com').
+            first()
+            )
```
Avantol13: style nitpick: this `)` should be de-dented (moved back) 4 spaces

tests/google/test_access_utils.py
```diff
@@ -521,3 +519,94 @@ def test_remove_service_account_raise_GoogleAPI_exc(
     with pytest.raises(GoogleAPIError):
         assert force_remove_service_account_from_access('test@gmail.com', 'test')
 
+def test_update_user_service_account_success(cloud_manager, db_session, setup_data):
```
Avantol13: you should have a few different version of this test. adding all new projects, keeping everything the same, removing all projects, adding and removing projects. Also, you should be asserting that the expiration gets updated in all these situations

fence/resources/google/access_utils.py
```diff
@@ -378,12 +429,205 @@ def force_remove_service_account_from_access(
...
+
+    """
+    for project_id in to_add_project_ids:
+        access_groups = _get_google_access_groups(session, project_id)
+
+        for access_group in access_groups:
+            try:
+                # TODO: Need to remove try/catch after major refactor
+                with GoogleCloudManager() as g_manager:
+                    if not g_manager.add_member_to_group(
```
Avantol13: I believe the response that indicates success is some JSON about the added member: https://developers.google.com/admin-sdk/directory/v1/reference/members#resource

Unless you tested and that's not what the API returns, Google's documentation isn't always correct. But if it does give us a JSON blob, we should fix the logic here

fence/resources/google/access_utils.py
```diff
@@ -378,12 +429,205 @@ def force_remove_service_account_from_access(
...
+            session
+            .query(ServiceAccountAccessPrivilege)
+            .filter_by(project_id=project_id, service_account_id=service_account.id)
+            .first()
+        )
+        session.delete(access_project)
+
+        access_groups = _get_google_access_groups(session, project_id)
+        for access_group in access_groups:
+            account_access_bucket_group = (
```
Avantol13: it's possible that an entry doesn't exist here if the service account access expired and it was removed from the group by the cronjob. So we should check if this exists before deleting it. If it doesn't exist, that's fine

fence/resources/google/access_utils.py
```diff
@@ -378,12 +429,205 @@ def force_remove_service_account_from_access(
...
+                        project_id=project_id,
+                        service_account_id=service_account.id
+                )
+        )
+
+        access_groups = _get_google_access_groups(session, project_id)
+
+        # use configured time or 7 days
+        expiration_time = (
+            time.time()
```
Avantol13: should cast to int since time.time() is a float. Just for consistency. so int(time.time())

fence/resources/google/access_utils.py
giangbui: Fixed at 9b9c05e8b4fca00310d061f93a604eb2056a4915

fence/resources/google/access_utils.py
giangbui: Fixed at 9b9c05e8b4fca00310d061f93a604eb2056a4915

fence/resources/google/access_utils.py
giangbui: Fixed at 9b9c05e8b4fca00310d061f93a604eb2056a4915

PR #369, opened 2018-08-02
PXD-248 Feat/register service accounts

benJrohrer:


fence/blueprints/google.py
```diff
@@ -85,8 +95,74 @@ def _post_new_service_account(
...
+                            service_account_id=db_service_account.id,
+                            access_group_id=gbag.id,
+                            expires=time.time()
+                            + flask.current_app.config.get(
+                                'GOOGLE_USER_SERVICE_ACCOUNT_ACCESS_EXPIRES_IN',
+                                604800)))
+                    current_session.add(service_account_to_gbag)
+
+                    with GoogleCloudManager() as prj:
+                        prj.add_member_to_group(
```
Avantol13: I think you need to validate the response from this to make sure it was successful. If not, raise some error

fence/blueprints/google.py
```diff
@@ -416,57 +492,99 @@ def _get_service_account_error_status(
...
+        return {
+            'status': 200,
+            'error': None,
+            'error_description': ''
+        }
+
     response = {
-        'status': 400,
-        'error': None,
+        'status': 403,
```
Avantol13: why populate this with unauthorized already?

Avantol13: oh, you replace it later. this is sorta confusing. I'd almost just set the status and error in the check below and not default up here

tests/google/test_service_accounts.py
```diff
@@ -170,6 +186,83 @@ def test_invalid_service_account_registration_errors(
     assert response.status_code != 200
 
 
+def test_valid_service_account_registration(
```
Avantol13: let's add a few more tests for testing when things are invalid we get the expected response. also a test with a service account that already exists would be good. maybe a test where cloud_manager throws an exception (like if we can't access the user's project) too, and we check the response is reasonable

fence/blueprints/google.py
```diff
@@ -78,15 +88,109 @@ def _post_new_service_account(
         if error_response.get('success') is not True:
             return error_response, 400
 
+        with GoogleCloudManager(google_project_id) as google_project:
```
Avantol13: you don't have to do it this way, we don't really need to hit google's api. We can just hit our db and check if it's there based on the email. UserServiceAccount.email == service_account_email

Avantol13:
can you also update the openapi documentation with any changes in the endpoints/responses you made?

PR #372, opened 2018-08-03
feat(patch-service-account): make endpoint extend access to all datasets

Avantol13:


fence/blueprints/google.py
```diff
@@ -180,19 +182,27 @@ def patch(self, id_):
             )
             return msg, 403
 
-        payload = flask.request.get_json() or {}
-        project_access = payload.get('project_access')
+        try:
+            payload = flask.request.get_json()
```
rudyardrichter: Specifically with `get_json`, you can avoid the `except` by passing `silent=True`, which silences errors:
```python
payload = flask.request.get_json(silent=True) or {}
```

fence/blueprints/google.py
```diff
@@ -180,19 +182,27 @@ def patch(self, id_):
...
+        except Exception:
+            payload = {}
+
+        service_account_email = get_service_account_email(id_)
 
         # check if the user requested to update more than project_access
         if 'project_access' in payload:
+            project_access = payload.get('project_access')
```
rudyardrichter: bit shorter:
```python
project_access = (
    payload.pop('project_access', None)
    or get_current_service_account_project_access(service_account_email)
)
```

fence/resources/google/access_utils.py
```diff
@@ -380,3 +383,110 @@ def force_remove_service_account_from_access(
...
+        filter_by(email=service_account_email).first()
+    )
+
+    if service_account:
+        bucket_access_groups = _get_google_access_groups_for_service_account(
+            service_account)
+
+        # use configured time or 7 days
+        expiration_time = (
+            time.time()
```
rudyardrichter: watch out, `time.time()` is a float. if you want int then `int(time.time())`

fence/resources/google/access_utils.py
```diff
@@ -380,3 +383,110 @@ def force_remove_service_account_from_access(
...
+    access to.
+
+    Args:
+        service_account_email (str): service account email
+
+    Returns:
+        List[str]: List of Project.auth_ids
+
+    Raises:
+        NotFound: if service account doesn't exist
```
rudyardrichter: the NotFound error itself is slightly misleading because it specifically means 404. I can't tell immediately if this is supposed to be 404 or not—just making sure you're aware

fence/resources/google/access_utils.py
```diff
@@ -380,3 +383,110 @@ def force_remove_service_account_from_access(
...
+def _get_google_access_groups_for_service_account(service_account):
+    """
+    Return list of fence.models.GoogleBucketAccessGroup objects that the
+    given service account should have access to based on it's access
+    privileges.
+
+    Args:
+        service_account (fence.models.UserServiceAccount):
+            service account object
+    """
```
rudyardrichter: Missing `Return` type

fence/resources/google/access_utils.py
```diff
@@ -380,3 +383,110 @@ def force_remove_service_account_from_access(
...
+    bucket_access_groups = []
+
+    accessed_projects = [
+        access_privilege.project
+        for access_privilege in service_account.access_privileges
+    ]
+    for project in accessed_projects:
+        for bucket in project.buckets:
+            groups = bucket.google_bucket_access_groups
+            bucket_access_groups.extend(groups)
```
rudyardrichter: you could lump this into one big list comprehension:
```python
return [
    group
    for access_privilege in service_account.access_privileges
    for bucket in access_privilege.project.buckets
    for group in bucket.google_bucket_access_groups
]
```
(I think I wrote that out correctly—maybe double-check)

fence/blueprints/google.py
Avantol13: ooh, fancy. changing to that

fence/blueprints/google.py
Avantol13: I like it, thanks

fence/resources/google/access_utils.py
Avantol13: ahh! good catch

fence/resources/google/access_utils.py
Avantol13: yeah, in this case it should be a 404 since we can't list the projects if the service account doesn't exist

PR #373, opened 2018-08-08
PXD-1028 Feat/verify bucket access grp

giangbui:


fence/scripting/fence_create.py
```diff
@@ -579,6 +580,110 @@ def delete_expired_service_accounts(DB):
...
+                .query(GoogleProxyGroupToGoogleBucketAccessGroup)
+                .filter_by(access_group_id=access_group.id)
+                .all()
+            )
+    ]
+
+    if not any([email for email in account_emails if email == member.get('email')]):
+        try:
+            with GoogleCloudManager() as manager:
+                manager.delete_group(member.get('email'))
```
Avantol13: this isn't the fnction you want to use. You don't want to attempt to delete the group, you want to remove it from the access group instead of trying to delete

fence/scripting/fence_create.py
```diff
@@ -579,6 +580,110 @@ def delete_expired_service_accounts(DB):
...
+            .query(ServiceAccountToGoogleBucketAccessGroup)
+            .filter_by(access_group_id=access_group.id)
+            .all()
+        )
+    ]
+
+    if not any([email for email in account_emails if email == member.get('email')]):
+        try:
+            with GoogleCloudManager() as manager:
+                manager.delete_service_account(member.get('email'))
```
Avantol13: same as above, we don't want to delete it, we want to remove it from the access group

fence/scripting/fence_create.py
```diff
@@ -579,6 +580,110 @@ def delete_expired_service_accounts(DB):
...
+                    return
+
+                for member in members:
+                    if member.get('type') == 'GROUP':
+                        _verify_google_group_member(session, access_group, member)
+                    elif member.get('type') == 'USER':
+                        _verify_google_service_account_member(session, access_group, member)
+
+
+def _verify_google_group_member(session, access_group, member):
```
Avantol13: you could take a GoogleCloudManager as an argument in these helper functions (something like `cloud_manager`) so you don't have to reinitialize a new one using the `with GoogleCloudManager(): ...`

PR #383, opened 2018-08-13
White list service account ids

benJrohrer:
* remove_white_listed_service_account_ids now removes service account ids listed in settings in addition to monitoring service account

* local_settings.py needs to be updated for this to take effect

### DIFF
--- a/fence/resources/google/access_utils.py
+++ b/fence/resources/google/access_utils.py
@@ -856,11 +856,10 @@ def remove_white_listed_service_account_ids(service_account_ids):
         service_account_ids.remove(monitoring_service_account)
 
     if 'WHITE_LISTED_SERVICE_ACCOUNT_EMAILS' in flask.current_app.config:
-        whitelist = (flask.current_app.config
-                     .get('WHITE_LISTED_SERVICE_ACCOUNT_EMAILS'))
-        for email in whitelist:
-            if email in service_account_ids:
-                service_account_ids.remove(email)
+        for email in (flask.current_app.config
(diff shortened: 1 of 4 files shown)


fence/resources/google/access_utils.py
```diff
@@ -854,6 +854,14 @@ def remove_white_listed_service_account_ids(service_account_ids):
     monitoring_service_account = get_monitoring_service_account_email()
     if monitoring_service_account in service_account_ids:
         service_account_ids.remove(monitoring_service_account)
+
+    if 'WHITE_LISTED_SERVICE_ACCOUNT_EMAILS' in flask.current_app.config:
+        whitelist = (flask.current_app.config
+                     .get('WHITE_LISTED_SERVICE_ACCOUNT_EMAILS'))
+        for email in whitelist:
+            if email in service_account_ids:
+                service_account_ids.remove(email)
```
giangbui: You can save 1 indent and several lines of code by rewriting as following:
```javascript
map(service_account_ids.remove, [email for email in (flask
                                                     .current_app
                                                     .config
                                                     .get('WHITE_LISTED_SERVICE_ACCOUNT_EMAILS', [])
                                                     )])

```

giangbui: Is the logic correct? accounts in while list should be not removed right?

Avantol13: yeah, it's maybe a confusing function name I added, but it removes the whitelisted accounts from the full list so we only validate the non-whitelisted accounts

Avantol13: so this logic is correct I believe

PR #388, opened 2018-08-20
fix(sync): bug fixes

rudyardrichter:
fixes bug(s) introduced by [feat/roles](https://github.com/uc-cdis/fence/pull/237)


fence/sync/sync_users.py
```diff
@@ -742,50 +743,41 @@ def _download_dbgap_file_list(self):
                 dbgap_file_list = glob.glob(os.path.join(tmpdir, '*'))
             except Exception as e:
                 self.logger.info(e)
+        permissions = [{'read-storage'}] * len(dbgap_file_list)
```
giangbui: All elements of permissions refer to the same object. I think it is more safe to replace it with
```
permissions = []
        for _ in xrange(len(dbgap_file_list)):
            permissions.append({'read-storage'})
```

giangbui: Can you update the docstring of _parse_csv function in the Example? should be set not list

giangbui: Also fix the docstring of sync_two_phsids_dict

PR #392, opened 2018-08-21
feat(external): function to setup external buckets

philloooo:
🚧


fence/scripting/fence_create.py
```diff
@@ -1096,9 +1096,33 @@ def _create_or_update_google_bucket_and_db(
...
+
+    access_group = _create_google_bucket_access_group(
+        db_session, google_bucket_name, bucket_db_id, google_project_id,
+        privileges)
+    # use storage creds to update bucket iam
+    storage_manager = GoogleCloudManager(
+        storage_creds_project_id,
+        creds=cirrus_config.configs['GOOGLE_STORAGE_CREDS'])
+    with storage_manager as g_mgr:
+        g_mgr.give_group_access_to_bucket(
```
benJrohrer: Shouldn't this be done by ISB?

philloooo: renamed the function to _setup_google_bucket_access_group and let it call _create_google_bucket_access_group then setup iam

Issue #395, opened 2018-08-24
PXD-1687 ⁃ update usersync to explicitly specify resource paths


philloooo:
create a separate field in useryaml called 'resources' for each user, similar to 'projects', just that the identifier is path instead of auth_id. Update usersync to sync that to arborist instead of projects

the resource path would be explicit '/programs/<name>/projects/<code>' or '/programs/<name>'

PR #396, opened 2018-08-28
feat/sync-paths: specify arborist resource paths in user yaml

rudyardrichter:


fence/sync/sync_users.py
```diff
@@ -343,6 +348,9 @@ def _parse_yaml(self, filepath, encrypted=True):
                             privileges[project['auth_id']] = set(
                                 project['privilege']
                             )
+                            privileges[project['resource']] = set(
```
philloooo: why do we share the same dict for resources and projects?

philloooo: also I don't expect fence create project for the resource?

rudyardrichter: no reason for them to share—I'll separate it

rudyardrichter: fence won't create a project (in the userdatamodel sense) for these resources

fence/sync/sync_users.py
```diff
@@ -328,42 +332,94 @@ def _parse_yaml(self, filepath, encrypted=True):
...
-
-                    try:
-                        for project in details.get('projects', {}):
-                            privileges[project['auth_id']] = set(
-                                project['privilege']
-                            )
-                    except KeyError as e:
-                        self.logger.info(e)
-                        continue
+        except IOError as e:
```
philloooo: Should just raise exception if the file doesn't present?
if we return empty data here it will clear the database because of mistake in automation configuration?

fence/sync/sync_users.py
```diff
@@ -328,42 +332,94 @@ def _parse_yaml(self, filepath, encrypted=True):
...
-                        'admin': details.get('admin', False),
-                    }
+        users = data.get('users', {})
+        for username, details in users.iteritems():
+            # users should occur only once each; skip if already processed
+            if username in user_project:
+                self.logger.error(
+                    'user `{}` occurs multiple times (skipping)'
+                    .format(username)
+                )
```
philloooo: didn't see it being skipped? I think we should just raise the exception here too.

fence/sync/sync_users.py
```diff
@@ -817,88 +882,87 @@ def _update_arborist(self, user_projects, session):
...
 
+        created_roles = set()
+        created_policies = set()
+
         for username, projects in user_projects.iteritems():
+            self.logger.info('processing user `{}`'.format(username))
+            projects = {
+                path: permissions
+                for path, permissions in projects.iteritems()
+                if path.startswith('/')
```
philloooo: path will always start with '/' ?
also lets rename projects to resources in this method?

fence/sync/sync_users.py
```diff
@@ -817,88 +882,87 @@ def _update_arborist(self, user_projects, session):
...
-                        policy_id = _format_policy_id(project, permission)
-                        policy = (
-                            session
-                            .query(Policy)
-                            .filter_by(id=policy_id)
-                            .first()
+                    # If everything was created fine, grant a policy to
+                    # this user which contains exactly just this resource,
+                    # with this permission as a role.
+                    project = path.split('/')[-1]
```
philloooo: this is the last resource that's attached to the user? the resource name won't be necessarily unique, so we can't _format_policy_id based on that?

fence/sync/sync_users.py
rudyardrichter: [This is how it is right now](https://github.com/uc-cdis/fence/blob/master/fence/sync/sync_users.py#L363)...I take it I should fix this

fence/sync/sync_users.py
rudyardrichter: "skipping" as in `continue`s to the next user. I'll change it to an `EnvironmentError`.

fence/sync/sync_users.py
philloooo: yeah I realized it's not a new behavior after I comment and read several lines after...

fence/sync/sync_users.py
rudyardrichter: changed in 0cee0f2

fence/sync/sync_users.py
rudyardrichter: fixed in 5cbb8f4

fence/sync/sync_users.py
rudyardrichter: fixed in 5cbb8f4

philloooo:
resolve #395

PR #399, opened 2018-09-04
PXD-1668 fix(google-access): add db entry during syncing to specify that a use…

Avantol13:
* Make sure we add entries to the db during user sync that link a user's proxy group to a bucket access groups (so we know what access they have)


fence/resources/storage/__init__.py
```diff
@@ -375,22 +390,62 @@ def _update_access_to_bucket(self, bucket, provider, storage_username, access):
...
+        """
+        storage_user_access_db_entry = GoogleProxyGroupToGoogleBucketAccessGroup(
+            proxy_group_id=storage_user.google_proxy_group_id,
+            access_group_id=bucket_access_group.id,
+        )
+        session.add(storage_user_access_db_entry)
+        session.commit()
+
+    @staticmethod
+    def _remove_google_db_entry_for_bucket_access(
```
rudyardrichter: this would be a good candidate for being a method on `GoogleProxyGroupToGoogleBucketAccessGroup` to remove user's access

fence/resources/storage/__init__.py
```diff
@@ -339,10 +344,12 @@ def _get_or_create_storage_user(self, username, provider, session):
...
         return user
 
-    def _update_access_to_bucket(self, bucket, provider, storage_username, access):
+    def _update_access_to_bucket(
+        self, bucket, provider, storage_user, storage_username, access, session
+    ):
         # Need different logic for google (since buckets can have multiple
         # access groups)
-        if provider == STORAGE_ACCESS_PROVIDER_NAME:
+        if provider == GOOGLE_PROVIDER:
```
rudyardrichter: I know I'm commenting on this retroactively but the `else` part here would work well as an early return, and then have the chunk for the True-branch for this if dedented

fence/resources/storage/__init__.py
Avantol13: you mean like some of the `delete` functions other models have? I ran into issues (a while back) using those functions because of mismatched flask sessions in tests or something, so I've been avoiding doing that

fence/resources/storage/__init__.py
Avantol13: ah, yup. I'll do that

fence/resources/storage/__init__.py
rudyardrichter: oh, that's unfortunate. especially since this is a `staticmethod` here it seems like it belongs more on `GoogleProxyGroupToGoogleBucketAccessGroup`, but if that wouldn't work, maybe leave a `FIXME` here for if we eventually fix that in the tests?

Avantol13: yeah I wish I remember exactly what the issue was :grimacing: probably shouldn't created a story or documented it. I haven't been using any of the .delete() functions on models since then. in any case, added a FIXME for now

PR #402, opened 2018-09-04
chore(authlib): upgrade authlib to v0.9

rudyardrichter:


PlanXCyborg:

<details>

```diff
--- fence/jwt/token.py
+++ blackened
@@ -67,11 +67,10 @@
         self.kid = kid
         self.claims = claims
 
 
 class UnsignedIDToken(AuthlibCodeIDToken):
-
     def __init__(self, token, header=None, **kwargs):
         header = header or {}
         super(UnsignedIDToken, self).__init__(token, header, **kwargs)
 
     def get_signed_and_encoded_token(self, kid, private_key):
```
</details></br>

fence/models.py
```diff
@@ -80,18 +81,32 @@ class Client(Base, OAuth2ClientMixin):
...
                 kwargs["_allowed_scopes"] = allowed_scopes
-        super(Client, self).__init__(**kwargs)
+        if "redirect_uris" in kwargs:
+            redirect_uris = kwargs.pop("redirect_uris")
+            if isinstance(redirect_uris, list):
+                kwargs["redirect_uri"] = "\n".join(redirect_uris)
+            else:
+                kwargs["redirect_uri"] = redirect_uris
+        # default grant types to just 'authorization_code'
+        if "grant_types" in kwargs:
```
Avantol13: pop has a default so I *think* you can do `kwargs.pop("grant_types", ["authorization_code"])`

fence/models.py
```diff
@@ -130,15 +139,39 @@ def get_by_client_id(client_id):
         with flask.current_app.db.session as session:
             return session.query(Client).filter_by(client_id=client_id).first()
 
+    def check_client_type(self, client_type):
```
Avantol13: I'm confused what this is but maybe I need to read into authlib a bit more to understand

fence/models.py
```diff
@@ -130,15 +139,39 @@ def get_by_client_id(client_id):
...
     def check_requested_scopes(self, scopes):
         return set(self.allowed_scopes).issuperset(scopes)
 
+    def check_token_endpoint_auth_method(self, method):
+        """
+        Only basic auth is supported. If anything else gets added, change this
+        """
+        auth_methods = ["client_secret_basic", "client_secret_post"]
+        return (self.is_confidential and method in auth_methods) or (
+            not self.is_confidential and method == "none"
```
Avantol13: is method really a string "none" and not Python's `None`?

fence/oidc/client.py
```diff
@@ -0,0 +1,12 @@
...
+from fence.models import Client
+
+
+def query_client(client_id):
+    with flask.current_app.db.session as session:
+        return session.query(Client).filter_by(client_id=client_id).first()
+
+
+def authenticate_public_client(query_client, request):
+    print()
```
Avantol13: What's this `print` for?

Avantol13: can just `pass`?

fence/jwt/token.py
```diff
@@ -176,20 +129,18 @@ def from_signed_and_encoded_token(
...
             encoded_token,
             public_key,
             algorithms="RS256",
             verify=verify,
             audience=client_id,
         )
-
-        token = cls(token)
+        headers = {}
+        token = cls(payload, headers)
```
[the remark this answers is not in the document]
rudyardrichter: The authlib token class now takes the headers for the JWT in addition to the claims

fence/jwt/token.py
```diff
@@ -176,20 +129,18 @@ def from_signed_and_encoded_token(
...
 
         if verify:
-            token.validate(
-                client_id=client_id, issuer=issuer, max_age=max_age, nonce=nonce
-            )
+            token.validate(now=max_age)
```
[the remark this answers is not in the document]
rudyardrichter: `max_age` goes in the arguments.

the `validate` interface is different now: authlib implements some basic stuff by default, and anything that needs to be something specific gets passed in as options with this weird syntax like this:

```python
options = {
    'iss': {
        'essential': True,
        'value': '...',
    }
}
token.validate(options=options)
```

The thing is that this isn't really how the JWTs get validated? like our validation logic hasn't changed. The way that this is getting used with our setup, _this_ validation is pretty much just a safety check that nothing weird went into the token (claims having wrong type or something). I can point you to the relevant code in authlib if that would help make more sense of it

fence/jwt/token.py
```diff
@@ -84,58 +84,11 @@ def get_signed_and_encoded_token(self, kid, private_key):
             str: UTF-8 encoded JWT ID token signed with ``private_key``
         """
         headers = {"kid": kid}
-        token = jwt.encode(self.token, private_key, headers=headers, algorithm="RS256")
+        headers.update(self.header)
+        token = jwt.encode(self, private_key, headers=headers, algorithm="RS256")
         token = to_unicode(token)
         return token
 
-    def validate_auth_time(self, max_age):
```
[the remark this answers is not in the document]
rudyardrichter: Partly yes—see my comment below about how this validation is not actually the same as validating an incoming user JWT. authlib has a function now in the token class to check that this field has the correct type and that it exists if it should. it doesn't check the actual _value_, since this validation is (as far as I can tell) only used as a check that no invalid claims are going out from fence, and separate from our validation for incoming JWTs. I'm going to have to double-check that that's accurate

fence/models.py
rudyardrichter: ahh, yes. much better, thank you

fence/oidc/client.py
rudyardrichter: Ooh, yes, leftover from debugging. Thanks

fence/models.py
rudyardrichter: Yep. There's these three methods allowed to handle client auth, which are `["client_secret_basic", "client_secret_post", "none"]` (which are kind of hard-coded into authlib, unfortunately).

Avantol13: Gross.

fence/oidc/grants/oidc_code_grant.py
```diff
@@ -0,0 +1,169 @@
...
+                    .filter_by(nonce=self.request.nonce)
+                    .first()
+                )
+                if not code:
+                    raise InvalidRequestError("Replay attack")
+        return True
+
+    def validate_prompt(self, end_user):
+        """
+        Override method in authlib to fix behavior with login prompt.
```
Avantol13: what's the incorrect behavior?

fence/oidc/grants/refresh_token_grant.py
```diff
@@ -117,27 +114,34 @@ def validate_access_token_request(self):
...
-        self.create_access_token(
-            token, self._authenticated_client, self._authenticated_token
+            scope = credential["aud"]
+
+        client = self.request.client
+        expires_in = credential["exp"]
+        token = self.generate_token(
+            client, self.GRANT_TYPE, user=user, expires_in=expires_in, scope=scope
         )
+        # TODO: logger
```
Avantol13: Is this to log the creation?

fence/oidc/jwt_generator.py
```diff
@@ -13,66 +12,56 @@
...
+        refresh_token:
+            for a refresh token grant, pass in the previous refresh token
+            to return that same token again instead of generating a new one
+            (otherwise this will let the refresh token refresh itself)
+        refresh_token_claims (dict):
+            also for a refresh token grant, pass the previous refresh token
+            claims (to avoid having to encode or decode the refresh token
+            here)
+    """
+    # prevent those bothersome "not bound to session" errors
```
Avantol13: For my clarity: when do these errors occur and how is this preventing them?

fence/oidc/oidc_server.py
```diff
@@ -1,9 +1,32 @@
...
+    in order to authenticate OAuth clients.
+    """
+
+    def authenticate(self, request, methods):
+        """
+        Override method from authlib
+        """
+        client = super(ClientAuthentication, self).authenticate(request, methods)
+        # don't allow confidential clients to not use auth
+        if client.is_confidential:
```
Avantol13: can I get some more clarity on this? why do we have to do this and what are these `methods`?

fence/oidc/grants/oidc_code_grant.py
rudyardrichter: this was the thing I asked you about because it was returning the authorization page first (with 200) instead of redirecting to our IDP like it was supposed to

fence/oidc/jwt_generator.py
rudyardrichter: if you carry around a reference to a sqlalchemy object after the session gets updated somehow then your reference is no longer bound to the session so you get this error (which seems to be really typical in our setup where we have models like `User` that have to get passed around so much). this re-queries for the model to prevent this—not ideal because that means you're not really using the reference correctly. I couldn't find a better way to stop this here though, hence these lines

fence/oidc/oidc_server.py
rudyardrichter: - `request` is an authlib object encapsulating information about the oauth request
- `methods` is a list of strings where the strings are the "client authentication types enum": `"none"`, `"client_secret_basic"`, or `"client_secret_post"`

rudyardrichter: so, what this override is doing is that for confidential clients (`.is_confidential`), it always removes `"none"`as an allowed method

fence/oidc/grants/refresh_token_grant.py
rudyardrichter: yep

PR #408, opened 2018-09-07
fix(sa-registration): bugs with not getting users correctly and not g…

Avantol13:
…etting access groups correctly

* use sets for google bucket access groups because a group can provide access to multiple buckets and we don't want to count it twice (causes errors/issues later on)
* fix function that was incorrectly returning UserGoogleAccount objects instead of generic fence User objects

Note: UserGoogleAccount objects are a User's linked Google account (after they go through oauth)

### DIFF
--- a/fence/resources/google/access_utils.py
+++ b/fence/resources/google/access_utils.py
@@ -682,14 +682,14 @@ def _get_google_access_groups(session, project_id):
     Returns:
         List(GoogleBucketAccessGroup)
     """
-    access_groups = []
+    access_groups = set()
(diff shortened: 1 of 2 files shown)


fence/resources/google/access_utils.py
```diff
@@ -682,12 +682,12 @@ def _get_google_access_groups(session, project_id):
     Returns:
         List(GoogleBucketAccessGroup)
     """
-    access_groups = []
+    access_groups = set()
```
rudyardrichter: docstring for this function no longer accurate

PR #410, opened 2018-09-10
Fix/sa registration

Avantol13:
* allow whitelisting Google parent organizations


fence/resources/google/access_utils.py
```diff
@@ -99,7 +99,7 @@ def google_project_has_parent_org(project_id):
     """
     try:
         with GoogleCloudManager(project_id, use_default=False) as prj:
-            return prj.has_parent_organization()
+            return prj.get_project_organization()
```
rudyardrichter: it looks like this is `str`, not `bool`?

fence/resources/google/access_utils.py
```diff
@@ -834,6 +834,27 @@ def remove_white_listed_service_account_ids(service_account_ids):
     return service_account_ids
 
 
+def remove_white_listed_parent_orgs(parent_orgs):
```
rudyardrichter: I'm confused why this is needed—it looks like this is only used in `check_validity` but it doesn't seem "designed" for that case; would it make more sense if this was `check_valid_parent_orgs` or something?

Avantol13: The function itself removes any configured org names from a list. It's used in check_validity because we only want to invalidate a project if it has a parent org that isn't whitelisted

fence/resources/google/access_utils.py
Avantol13: whoops, need to update docstring :man_facepalming:

fence/resources/google/access_utils.py
rudyardrichter: but here it looks like the value isn't actually getting used:
https://github.com/uc-cdis/fence/pull/410/files#diff-61d2a1b1d71f839686569b101f705d6aR210

PlanXCyborg:

fence/resources/google/access_utils.py
Avantol13: oh, Python truthiness hackiness. It should be an empty list. if it's not, then it's not valid

Avantol13: so at the moment, we don't care about the actual value, just that it's not there. In the future we may want the actual values to display to the user or something, hence the generic removal function

rudyardrichter: so currently this would be sufficient?
```python
def is_org_whitelisted(org):
    return org in flask.current_app.config.get("WHITE_LISTED_GOOGLE_PARENT_ORGS")

...

valid_parent_org = is_org_whitelisted(parent_org)
```
and later if we want to log things not whitelisted:
```python
log(org for org in orgs if not is_org_whitelisted(org))
```

rudyardrichter: (the removal thing just seems unnecessarily convoluted to me, and the `email in parent_orgs` is kind of backwards)

Avantol13: point taken. I'll change it

PR #413, opened 2018-09-11
Fix/sa registration account type

Avantol13:
* We had incorrectly been handling service accounts on the Google project by always running all of the validity checks
    * We only need to run all of them on "User-Managed" service account. Google-managed service accounts on the project only need a subset of the validity checks
* We cannot always obtain the Google project ID from the service account (service accounts have different email signatures depending on their function) <- We did not know this before
    * Change the logic/method for obtaining the project id 

relies on https://github.com/uc-cdis/cirrus/pull/37
and https://github.com/uc-cdis/cloud-automation/pull/462


fence/blueprints/google.py
```diff
@@ -335,31 +333,44 @@ def patch(self, id_):
             id_ (str): Google service account identifier to update
         """
         user_id = current_token["sub"]
-        # check if user has permission to update the service account
-        authorized = can_user_manage_service_account(user_id, id_)
```
Avantol13: this function relied on getting the google project id form the service account email (which we cannot reliably do). So the new method actually hits our db and gets the project from there for the registered service account

fence/blueprints/google.py
```diff
@@ -335,31 +333,44 @@ def patch(self, id_):
...
+            raise NotFound(
+                "Could not find a registered service account from given email {}".format(
+                    service_account_email
+                )
+            )
 
         # check if the user requested to update more than project_access
-        project_access = payload.pop(
-            "project_access", None
-        ) or get_current_service_account_project_access(service_account_email)
```
Avantol13: this function hit the db again to search for the registered account. since we already have that now, no need to have this function

fence/blueprints/google.py
```diff
@@ -388,8 +399,22 @@ def delete(self, id_):
             id_ (str): Google service account email to delete
         """
         user_id = current_token["sub"]
+
+        service_account_email = get_service_account_email(id_)
+        registered_service_account = get_registered_service_account_from_email(
+            service_account_email
+        )
+        if not registered_service_account:
+            raise NotFound(
```
Avantol13: fix delete bug: we weren't checking if a service account existed before trying to delete

fence/blueprints/google.py
```diff
@@ -594,7 +621,7 @@ def _get_service_account_error_status(
 
 
 def _get_service_account_email_error_status(validity_info):
-    service_accounts_validity = validity_info.get("service_accounts")
```
Avantol13: new format for validity output includes a separate item for the new service account that the user is attempting to register

fence/local_settings.example.py
```diff
@@ -157,6 +157,29 @@
 dir_path = "/secrets"
 fence_creds = os.path.join(dir_path, "fence_credentials.json")
 
+GOOGLE_MANAGED_SERVICE_ACCOUNT_DOMAINS = {
```
Avantol13: these were provided by ISB-CGC as a starting point for determining if a service account is Google-managed. unfortunately, Google does not have a standard naming convention and User-managed account also end in gserviceaccount.com. So this list contains all known Google-managed account domains.

fence/resources/google/access_utils.py
```diff
@@ -235,10 +218,21 @@ def is_service_account_from_google_project(service_account_email, project_id):
...
     """
     try:
-        sa_google_project = get_google_project_from_service_account_email(
-            service_account_email
-        )
+        service_account_name = service_account_email.split("@")[0]
+
+        if is_google_managed_service_account(service_account_email):
+            return (
+                service_account_name == "service-{}".format(project_number)
```
Avantol13: these parsing methods were also provided by ISB-CGC who has been using them in production. While Google doesn't have a standard way, there are known-patterns to check

Avantol13: and the point of this is to determine if the service account was created in and belongs to the provided project. No outside accounts are allowed

PR #418, opened 2018-09-17
PXD-1371 Feat/email

giangbui:
FN364 ⁃ email users/client when a user-registered service account gets kicked off for being invalid

### DIFF
--- a/fence/local_settings.example.py
+++ b/fence/local_settings.example.py
@@ -180,6 +180,21 @@ GOOGLE_MANAGED_SERVICE_ACCOUNT_DOMAINS = {
     "storage-transfer-service.iam.gserviceaccount.com",
 }
 
+REMOVE_SERVICE_ACCOUNT_EMAIL_NOTIFICATION = {
+    "domain": "smtp domain",
+    "subject": "User service account removal notification",
+    "from": "test@planx-pla.net",
+    "contact number": "123456789",
+    "content": """
+    
+    Service accouts {} were removed from access control lists because some \
+    users or service accounts of GCP project {} are not authorized to access\
+    the data sets associated to the service accounts, or do not \
+    adhere to ISB-CGC's security policies.
+    
+    """
+}
+
 SUPPORT_EMAIL_FOR_ERRORS = None
 dbGaP = {}
 if os.path.exists(fence_creds):
--- a/fence/local_settings.example.py
+++ b/fence/local_settings.example.py
@@ -186,12 +186,12 @@ REMOVE_SERVICE_ACCOUNT_EMAIL_NOTIFICATION = {
     "from": "test@planx-pla.net",
     "contact number": "123456789",
     "content": """
-    
+
     Service accouts {} were removed from access control lists because some \
     users or service accounts of GCP project {} are not authorized to access\
     the data sets associated to the service accounts, or do not \
-    adhere to ISB-CGC's security policies.
-    
+    adhere to the security policies.
+
     """
 }
 
--- a/fence/local_settings.example.py
+++ b/fence/local_settings.example.py
@@ -183,14 +183,15 @@ GOOGLE_MANAGED_SERVICE_ACCOUNT_DOMAINS = {
 REMOVE_SERVICE_ACCOUNT_EMAIL_NOTIFICATION = {
     "domain": "smtp domain",
     "subject": "User service account removal notification",
-    "from": "test@planx-pla.net",
+    "from": "do-not-reply@planx-pla.net",
+    "admin": [],
     "contact number": "123456789",
     "content": """
 
     Service accouts {} were removed from access control lists because some \
-    users or service accounts of GCP project {} are not authorized to access\
-    the data sets associated to the service accounts, or do not \
-    adhere to the security policies.
+users or service accounts of GCP project {} are not authorized to access \
+the data sets associated to the service accounts, or do not \
+adhere to the security policies.
 
     """
 }
--- a/fence/local_settings.example.py
+++ b/fence/local_settings.example.py
@@ -186,13 +186,11 @@ REMOVE_SERVICE_ACCOUNT_EMAIL_NOTIFICATION = {
     "from": "do-not-reply@planx-pla.net",
     "admin": [],
     "contact number": "123456789",
-    "content_singular": """
-
+    "content": """
     The service accounts were removed from access control data because some \
 users or service accounts of GCP project {} are not authorized to access \
 the data sets associated to the service accounts, or do not \
 adhere to the security policies.
-
     """
 }
 
--- a/fence/local_settings.example.py
+++ b/fence/local_settings.example.py
@@ -186,9 +186,9 @@ REMOVE_SERVICE_ACCOUNT_EMAIL_NOTIFICATION = {
     "from": "do-not-reply@planx-pla.net",
     "admin": [],
     "contact number": "123456789",
-    "content": """
+    "content_singular": """
 
-    Service accounts {} were removed from access control lists because some \
+    The service accounts were removed from access control data because some \
 users or service accounts of GCP project {} are not authorized to access \
 the data sets associated to the service accounts, or do not \
 adhere to the security policies.
--- a/fence/local_settings.example.py
+++ b/fence/local_settings.example.py
@@ -202,3 +202,4 @@ if os.path.exists(fence_creds):
         GOOGLE_MANAGED_SERVICE_ACCOUNT_DOMAINS.update(
             data.get("GOOGLE_MANAGED_SERVICE_ACCOUNT_DOMAINS", [])
         )
+        GUN_MAIL = data.get("GUN_MAIL")
--- a/fence/local_settings.example.py
+++ b/fence/local_settings.example.py
@@ -188,7 +188,7 @@ REMOVE_SERVICE_ACCOUNT_EMAIL_NOTIFICATION = {
     "contact number": "123456789",
     "content": """
 
-    Service accouts {} were removed from access control lists because some \
+    Service accounts {} were removed from access control lists because some \
 users or service accounts of GCP project {} are not authorized to access \
 the data sets associated to the service accounts, or do not \
 adhere to the security policies.
--- a/fence/resources/google/access_utils.py
+++ b/fence/resources/google/access_utils.py
@@ -66,7 +66,7 @@ def get_google_project_number(google_project_id):
         return None
 
 
-def get_google_project_parent_org(project_id):
+def get_google_project_parent_org(project_id, console_log=None):
     """
     Checks if google project has parent org. Wraps
     GoogleCloudManager.get_project_organization()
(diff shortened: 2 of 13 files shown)


fence/local_settings.example.py
```diff
@@ -180,6 +180,21 @@
...
+    "domain": "smtp domain",
+    "subject": "User service account removal notification",
+    "from": "test@planx-pla.net",
+    "contact number": "123456789",
+    "content": """
+    
+    Service accouts {} were removed from access control lists because some \
+    users or service accounts of GCP project {} are not authorized to access\
+    the data sets associated to the service accounts, or do not \
+    adhere to ISB-CGC's security policies.
```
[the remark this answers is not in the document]
giangbui: Fixed at 64454c99f8fca397472dcd0f73841e9246e4d53f

fence/local_settings.example.py
```diff
@@ -202,3 +216,4 @@
         GOOGLE_MANAGED_SERVICE_ACCOUNT_DOMAINS.update(
             data.get("GOOGLE_MANAGED_SERVICE_ACCOUNT_DOMAINS", [])
         )
+        GUN_MAIL = data.get("GUN_MAIL")
```
Avantol13: do you have a cloud-automation PR to add this to the fence config there?

fence/scripting/google_monitor.py
```diff
@@ -104,6 +116,34 @@ def _is_valid_google_project(google_project_id, db=None):
...
+
+    valid_type = service_account_validity['valid_type']
+    no_external_access = service_account_validity['no_external_access']
+
+    removal_reason = None
+
+    if not valid_type:
+        removal_reason = "it must be a compute engine service account or an user-managed service account.\n"
+    elif not no_external_access:
+        removal_reason = "it does not have external access.\n"
```
Avantol13: this should be "it has either roles attached to it or service account keys generated. We do not allow this because we need to restrict external access."

fence/scripting/google_monitor.py
```diff
@@ -104,6 +116,34 @@ def _is_valid_google_project(google_project_id, db=None):
...
+    if not isinstance(service_account_validity, GoogleServiceAccountValidity):
+        return None
+
+    valid_type = service_account_validity['valid_type']
+    no_external_access = service_account_validity['no_external_access']
+
+    removal_reason = None
+
+    if not valid_type:
+        removal_reason = "it must be a compute engine service account or an user-managed service account.\n"
```
Avantol13: I'd capitalize Google product name: "it must be a Compute Engine service account or a user-managed service account"

fence/utils.py
```diff
@@ -189,3 +191,44 @@ def split_url_and_query_params(url):
...
+    """
+    from fence.settings import GUN_MAIL
+
+    api_key = GUN_MAIL[smtp_domain]['api_key']
+    email_url = GUN_MAIL[smtp_domain]['api_url'] + '/messages'
+
+    return requests.post(email_url, auth=('api', api_key), data={
+        'from': from_email,
+        'to': to_emails,
+        ''
```
Avantol13: what's this empty arg? maybe add a comment why this is like this

fence/scripting/google_monitor.py
```diff
@@ -104,6 +116,34 @@ def _is_valid_google_project(google_project_id, db=None):
...
+    elif not no_external_access:
```
Avantol13: It could have both valid_type false and no_external_access, so this should be able to handle the case where both fail.

Avantol13: right now, because of the elif, it won't get both.

fence/scripting/google_monitor.py
```diff
@@ -64,9 +73,12 @@ def validation_check(db):
                 force_remove_service_account_from_access(
                     sa_email, google_project_id, db=db
                 )
-            continue
+            invalid_service_account_reasons[sa_email] = "the project is invalid"
```
Avantol13: You can actually parse the validity for the project to get a more specific reason here. maybe write a "_get_project_removal_reason"

fence/scripting/google_monitor.py
```diff
@@ -138,3 +178,89 @@ def _get_project_service_account_mapping(registered_service_accounts):
...
+    """
+    try:
+        with GoogleCloudManager(project_id, use_default=False) as prj:
+            members = prj.get_project_membership(project_id)
+            for member in members:
+                if not (
+                    member.member_type == GooglePolicyMember.SERVICE_ACCOUNT
+                    or member.member_type == GooglePolicyMember.USER
+                ):
+                    raise NotSupported(
```
Avantol13: you shouldn't raise an error here because one of the reasons the project may be invalid is because of a wrong member type. and we should still be able to send an email. Just ignore all members who are not USER types

fence/scripting/google_monitor.py
```diff
@@ -138,3 +178,89 @@ def _get_project_service_account_mapping(registered_service_accounts):
...
+            return list({
+                u.email_id
+                for u in users
+                for role in u.roles
+                if role.name.upper() == "OWNER"})
+
+    except Exception as exc:
+        print(
+            (
+                "validity of Google Project (id: {}) members "
```
Avantol13: We shouldn't be checking the validity here, we just need to get the owner

fence/local_settings.example.py
```diff
@@ -180,6 +180,20 @@
     "storage-transfer-service.iam.gserviceaccount.com",
 }
 
+REMOVE_SERVICE_ACCOUNT_EMAIL_NOTIFICATION = {
+    "domain": "smtp domain",
+    "subject": "User service account removal notification",
```
Avantol13: Is that what ISB uses? Seems like something like "Data Access Revoked: Invalid Google Service Account Removed" might be better

fence/local_settings.example.py
giangbui: it is in json config file. There is `kube-setup-secret` to handle it already

fence/local_settings.example.py
giangbui: That subject is followed by the sample email they provided.

fence/local_settings.example.py
Avantol13: I don't see the fence_settings file pulling that variable in here: https://github.com/uc-cdis/cloud-automation/blob/master/apis_configs/fence_settings.py#L125-L158 you have to add logic to this settings file to grab the new config

fence/local_settings.example.py
Avantol13: Okay :man_shrugging:

fence/resources/google/access_utils.py
```diff
@@ -240,16 +263,20 @@ def is_service_account_from_google_project(
         )
 
     except Exception as exc:
-        flask.current_app.logger.debug(
-            (
+        stm = (
```
Avantol13: what does stm mean? maybe chance to `message` or `msg`

fence/resources/google/access_utils.py
```diff
@@ -268,16 +297,18 @@ def is_user_member_of_all_google_projects(user_id, google_project_ids, db=None):
...
+        ).format(user_id, google_project_ids)
+        if console_log:
+            print(stm)
+        else:
+            flask.current_app.logger.debug(stm)
         return False
 
     linked_google_account = (
-        current_session.query(UserGoogleAccount)
+        session.query(UserGoogleAccount)
```
Avantol13: where is `session` coming from? should you pass it in?

fence/resources/google/access_utils.py
```diff
@@ -822,6 +863,9 @@ def is_org_whitelisted(parent_org):
     Returns:
         bool: whether or not the provide Google parent organization is whitelisted
     """
+    if settings_from_local:
+        return local_settings.WHITE_LISTED_GOOGLE_PARENT_ORGS
```
Avantol13: this is not the same logic as the other check. it need to return whether or not the given parent org is in the local_settings. I don't actually think I like the `settings_from_local` arg in all these functions (it muddies the logic a bit).

Do you need this in all these functions? I was assuming you'd just pass in the information you need and get it from `local_settings` in the fence-create script.

e.g. here you'd have a `white_listed_parent_orgs=None` and then use the `flask.current_app.config` if it's `None`. Then in fence-create you just get the `local_settings.WHITE_LISTED_PARENT_ORGS` and pass it into this function.

That's a lot more clear to me and makes sure that the weird `local_settings` logic is only in the fence-create stuff where it's needed. 

Also if we every move configs out of local_settings (which was the plan with fence's config refactor) this portion of code would be really difficult to fix if we're using local_settings everything

Sorry, I know it's a big change, but when we change the config in the future it'll be painful to fix this so I'd steer away from it and only use local_settings in fence-create.

Also, probably the real fix for the logging is to abstract it somehow and then print to console if there is no flask.current_app, but that's a way bigger change too

fence/resources/google/validity.py
```diff
@@ -187,7 +187,7 @@ def __init__(
         self._info["service_accounts"] = {}
         self._info["access"] = {}
 
-    def check_validity(self, early_return=True, db=None):
+    def check_validity(self, early_return=True, db=None, settings_from_local=False, console_log=False):
```
Avantol13: Similar to my comment above, I don't really like passing this console_log into every function, it muddies the function interface. I think the real solution going forward is to abstract our logging and print to console if we don't find a flask.current_app. tldr; I'd remove these and only include the ones you need for your stuff to work

fence/scripting/google_monitor.py
```diff
@@ -138,3 +211,70 @@ def _get_project_service_account_mapping(registered_service_accounts):
...
+        users = [
+            member
+            for member in members
+            if member.member_type == GooglePolicyMember.USER
+        ]
+
+        return list({
+            u.email_id
+            for u in users
+            for role in u.roles
```
Avantol13: Ooh, fancy! A double `for` loop :smile: I like it.

fence/resources/google/access_utils.py
giangbui: https://github.com/uc-cdis/fence/pull/418/files/1db434bd49919429c63b2cb15cddf2039277a95b#diff-304e6573254ad3a1ae9058a5e159983bR297

giangbui: This function already created a session!

fence/resources/google/access_utils.py
giangbui: stm -> statment. I think msg is better choice

fence/scripting/google_monitor.py
```diff
@@ -104,6 +119,64 @@ def _is_valid_google_project(google_project_id, db=None):
...
+    """
+    Get service account removal reason
+
+    Args:
+        service_account_validity(GoogleServiceAccountValidity): service account validity
+
+    Returns:
+        str: the reason service account was removed
+    """
+    if not isinstance(service_account_validity, GoogleServiceAccountValidity):
```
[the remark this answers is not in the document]
giangbui: For the case that the _is_valid_google_project return False, we don't want to go further. The logic of _is_valid_google_project is not clear

fence/scripting/google_monitor.py
```diff
@@ -104,6 +119,64 @@ def _is_valid_google_project(google_project_id, db=None):
...
+    """
+    Get service account removal reason
+
+    Args:
+        google_project_validity(GoogleProjectValidity): google project validity
+
+    Returns:
+        str: the reason project was removed
+    """
+    if not isinstance(google_project_validity, GoogleProjectValidity):
```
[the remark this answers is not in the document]
giangbui: The output of _is_valid_google_project is either class instance or bool. I think _is_valid_google_project need to be rewritten. I don't want to change it since it has been used by other functions

fence/scripting/google_monitor.py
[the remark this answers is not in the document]
Avantol13: can you instead change the code in _is_valid_service_account to return None instead of False? and then check for None here and you can return a reason "got an unexpected error during validation and could not fully validate." That would be a little better than the isinstance check

fence/scripting/google_monitor.py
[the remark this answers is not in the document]
Avantol13: You can change it like I recommended above for the service account one. Change it to return None instead of False.

fence/resources/google/access_utils.py
Avantol13: Oh. :man_facepalming: I missed that

PR #421, opened 2018-09-21
fix(cookies): set cookies to expire based on their name *not* value

Avantol13:
http://flask.pocoo.org/docs/0.12/api/#flask.Request.cookies

cookies
A dict with the contents of all cookies transmitted with the request.

We were getting the *values* and setting new cookies with those as the *name*. Need to set cookies for the *names* stored in the dict's keys.


PlanXCyborg:

PR #438, opened 2018-10-08
fix(authlib): interface changed, .session is no longer available

Avantol13:
authutils has new authlib for this fence_client. the interface changed, can no longer directly access `.session`, there's now a function to return the session.

Fence 500's on master for multi-tenant flow without this fix.

Seems like `authlib` decided to make retrieving the `session` an internal function, which seems strange. Perhaps there's a better way to do the access token fetching with the correct redirect. But for now, this is the quick fix

### DIFF
--- a/fence/blueprints/login/fence_login.py
+++ b/fence/blueprints/login/fence_login.py
@@ -56,7 +56,7 @@ class FenceLogin(Resource):
                 " login page for the original application to continue."
             )
         # Get the token response and log in the user.
-        redirect_uri = flask.current_app.fence_client.session.redirect_uri
+        redirect_uri = flask.current_app.fence_client._get_session().redirect_uri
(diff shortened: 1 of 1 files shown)


PlanXCyborg:

fence/blueprints/login/fence_login.py
```diff
@@ -56,7 +56,7 @@ def get(self):
                 " login page for the original application to continue."
             )
         # Get the token response and log in the user.
-        redirect_uri = flask.current_app.fence_client.session.redirect_uri
+        redirect_uri = flask.current_app.fence_client._get_session().redirect_uri
```
lepture: `redirect_uri` is automatically handled by Authlib. Check https://docs.authlib.org/en/latest/client/flask.html#implement-the-server

Using `authorize_redirect` and `authorize_access_token`.

PR #454, opened 2018-10-18
Fix/validation

Avantol13:
add check back in so cronjobs don't try to validate a `None` user. We should just not do a user validation check if a user isn't provided


PlanXCyborg:

PR #455, opened 2018-10-19
Service Account Exists Fix PXD-2061

benJrohrer:
Check if SA account exists by trying to get IAM policy, if it does then pass through to has_external_access check, if not then add it to error response


PlanXCyborg:

<details>

```diff
--- fence/blueprints/google.py
+++ blackened
@@ -697,22 +697,22 @@
     }
 
     for sa_account_id, sa_validity in service_accounts_validity:
         if sa_account_id == validity_info.new_service_account:
             if not sa_validity:
-                if sa_validity['exists']:
+                if sa_validity["exists"]:
                     response["status"] = 403
                     response["error"] = "unauthorized"
                     response[
                         "error_description"
                     ] = "Service account requested for registration is invalid."
                 else:
                     response["status"] = 404
                     response["error"] = "Not Found"
-                    response["error_description"] = (
-                        "Service account could not be found."
-                    )
+                    response[
+                        "error_description"
+                    ] = "Service account could not be found."
 
             response["service_account_validity"] = {
                 str(sa_account_id): sa_validity.get_info()
             }
 

--- fence/resources/google/access_utils.py
+++ blackened
@@ -59,13 +59,14 @@
     """
     try:
         response = google_cloud_manager.get_project_info()
         return response.get("projectNumber")
     except Exception as exc:
-        logger.error("Could not determine google project number for Project"
-                     "ID: {} due to error. (Details : {})"
-                     .format(google_project_id, exc))
+        logger.error(
+            "Could not determine google project number for Project"
+            "ID: {} due to error. (Details : {})".format(google_project_id, exc)
+        )
         return None
 
 
 def get_google_project_membership(project_id, google_cloud_manager):
     """
@@ -96,20 +97,21 @@
     """
     try:
         return google_cloud_manager.get_project_organization()
     except Exception as exc:
         logger.error(
-                "Could not determine if Google project (id: {}) has parent org"
-                "due to error (Details: {})"
-                .format(getattr(google_cloud_manager, "project_id", "unknown")
-                        , exc)
+            "Could not determine if Google project (id: {}) has parent org"
+            "due to error (Details: {})".format(
+                getattr(google_cloud_manager, "project_id", "unknown"), exc
+            )
         )
         return None
 
 
 def get_google_project_valid_users_and_service_accounts(
-        project_id, google_cloud_manager, membership=None):
+    project_id, google_cloud_manager, membership=None
+):
     """
     Gets google project members of type
     USER or SERVICE_ACCOUNT and raises an error if it finds a member
     that isn't one of those types.
 
@@ -171,22 +173,23 @@
         Bool: True if service acocunt type is allowed as defined
         in ALLOWED_SERVICE_ACCOUNT_TYPES
     """
     try:
         sa_type = google_cloud_manager.get_service_account_type(account_id)
-        return (
-            sa_type
-            in ALLOWED_SERVICE_ACCOUNT_TYPES
-        )
+        return sa_type in ALLOWED_SERVICE_ACCOUNT_TYPES
     except Exception as exc:
         logger.error(
             "validity of Google service account {} (google project: {}) type "
-            "determined False due to error. Details: {}"
-            .format(account_id, google_cloud_manager.project_id, exc))
-
-
-def service_account_has_external_access(service_account, google_cloud_manager, policy=None):
+            "determined False due to error. Details: {}".format(
+                account_id, google_cloud_manager.project_id, exc
+            )
+        )
+
+
+def service_account_has_external_access(
+    service_account, google_cloud_manager, policy=None
+):
     """
     Checks if service account has external access or not.
 
     Args:
         service_account(str): service account
@@ -194,15 +197,18 @@
         policy(dict): response from previous call to get_service_account_policy
 
     Returns:
         bool: whether or not the service account has external access
     """
-    response = policy or google_cloud_manager.get_service_account_policy(service_account)
+    response = policy or google_cloud_manager.get_service_account_policy(
+        service_account
+    )
     if response.status_code != 200:
         logger.error(
-            "Unable to get IAM policy for service account {}\n{}."
-            .format(service_account, response.json())
+            "Unable to get IAM policy for service account {}\n{}.".format(
+                service_account, response.json()
+            )
         )
         # if there is an exception, assume it has external access
         return True
 
     json_obj = response.json()
@@ -265,11 +271,11 @@
         )
         return False
 
 
 def is_user_member_of_google_project(
-        user_id, google_cloud_manager, db=None, membership=None
+    user_id, google_cloud_manager, db=None, membership=None
 ):
     """
         Return whether or not the given user is a member of the provided
         Google project ID.
 
@@ -290,44 +296,46 @@
     session = get_db_session(db)
     user = session.query(User).filter_by(id=user_id).first()
     if not user:
         logger.error(
             "Could not determine if user (id: {} is from project:"
-            " {} due to error. User does not exist..."
-                .format(user_id, google_cloud_manager.project_id))
+            " {} due to error. User does not exist...".format(
+                user_id, google_cloud_manager.project_id
+            )
+        )
         return False
 
     linked_google_account = (
         session.query(UserGoogleAccount)
-            .filter(UserGoogleAccount.user_id == user_id)
-            .first()
+        .filter(UserGoogleAccount.user_id == user_id)
+        .first()
     )
 
     try:
         members = membership or google_cloud_manager.get_project_membership()
         member_emails = [member.email_id.lower() for member in members]
         # first check if user.email is in project, then linked account
         if not (user.email and user.email in member_emails):
             if not (
-                linked_google_account
-                and linked_google_account.email in member_emails
+                linked_google_account and linked_google_account.email in member_emails
             ):
                 # no user email is in project
                 return False
     except Exception as exc:
         logger.error(
             "Could not determine if user (id: {}) is from project:"
-            " {} due to error. Details: {}"
-            .format(user.id,
-                    getattr(google_cloud_manager, "project_id", "unknown"),
-                    exc))
+            " {} due to error. Details: {}".format(
+                user.id, getattr(google_cloud_manager, "project_id", "unknown"), exc
+            )
+        )
         return False
 
     return True
 
+
 def is_user_member_of_all_google_projects(
-        user_id, google_project_ids, db=None, membership=None,
+    user_id, google_project_ids, db=None, membership=None
 ):
     """
     Return whether or not the given user is a member of ALL of the provided
     Google project IDs.
 
@@ -346,11 +354,13 @@
               Google project IDs
     """
     is_member = False
     for google_project_id in google_project_ids:
         with GoogleCloudManager(google_project_id) as google_cloud_manager:
-            is_member = is_user_member_of_google_project(user_id, google_cloud_manager, db, membership)
+            is_member = is_user_member_of_google_project(
+                user_id, google_cloud_manager, db, membership
+            )
 
             if not is_member:
                 return False
 
     return is_member

--- fence/resources/google/validity.py
+++ blackened
@@ -178,11 +178,13 @@
         """
         self.google_project_id = google_project_id
         self.new_service_account = new_service_account
         self.new_service_account_access = new_service_account_access or []
         self.user_id = user_id
-        self.google_cloud_manager = google_cloud_manager or GoogleCloudManager(google_project_id)
+        self.google_cloud_manager = google_cloud_manager or GoogleCloudManager(
+            google_project_id
+        )
 
         super(GoogleProjectValidity, self).__init__(*args, **kwargs)
 
         # setup default values for error information, will get updated in
         # check_validity
@@ -205,20 +207,23 @@
             early_return (bool, optional): Description
         """
 
         self.google_cloud_manager.open()
 
-        google_project_number = get_google_project_number(self.google_project_id, self.google_cloud_manager)
+        google_project_number = get_google_project_number(
+            self.google_project_id, self.google_cloud_manager
+        )
         has_access = bool(google_project_number)
 
         self.set("monitor_has_access", has_access)
         # always early return if we can't access the project
         if not has_access:
             return
 
         membership = get_google_project_membership(
-            self.google_project_id, self.google_cloud_manager)
+            self.google_project_id, self.google_cloud_manager
+        )
 
         user_has_access = is_user_member_of_google_project(
             self.user_id, self.google_cloud_manager, membership=membership, db=db
         )
         self.set("user_has_access", user_has_access)
@@ -247,15 +252,12 @@
             return
 
         user_members = None
         service_account_members = []
         try:
-            user_members, service_account_members = (
-                get_google_project_valid_users_and_service_accounts(
-                    self.google_project_id,
-                    self.google_cloud_manager,
-                    membership=membership)
+            user_members, service_account_members = get_google_project_valid_users_and_service_accounts(
+                self.google_project_id, self.google_cloud_manager, membership=membership
             )
             self.set("valid_member_types", True)
         except Exception:
             self.set("valid_member_types", False)
             if early_return:
@@ -276,13 +278,14 @@
         # validity. then check all the service accounts. Top level will be
         # invalid if any service accounts are invalid
         new_service_account_validity = ValidityInfo()
         if self.new_service_account:
             service_account_validity_info = GoogleServiceAccountValidity(
-                self.new_service_account, self.google_project_id,
+                self.new_service_account,
+                self.google_project_id,
                 google_project_number=google_project_number,
-                google_cloud_manager=self.google_cloud_manager
+                google_cloud_manager=self.google_cloud_manager,
             )
 
             service_account_id = str(self.new_service_account)
 
             google_sa_domains = (
@@ -340,13 +343,14 @@
         service_accounts_validity = ValidityInfo()
         for service_account in service_accounts:
             service_account_id = str(service_account)
 
             service_account_validity_info = GoogleServiceAccountValidity(
-                service_account, self.google_project_id,
+                service_account,
+                self.google_project_id,
                 google_project_number=google_project_number,
-                google_cloud_manager=self.google_cloud_manager
+                google_cloud_manager=self.google_cloud_manager,
             )
 
             google_sa_domains = (
                 config.get("GOOGLE_MANAGED_SERVICE_ACCOUNT_DOMAINS") if config else None
             )
@@ -455,31 +459,37 @@
             'owned_by_project': True
         }
     """
 
     def __init__(
-        self, account_id, google_project_id, google_cloud_manager=None,
-            google_project_number=None, *args, **kwargs
+        self,
+        account_id,
+        google_project_id,
+        google_cloud_manager=None,
+        google_project_number=None,
+        *args,
+        **kwargs
     ):
         self.account_id = account_id
         self.google_project_id = google_project_id
 
         # default to the given project id if not provided
         self.google_project_number = google_project_number or google_project_id
-        self.google_cloud_manager = (
-                google_cloud_manager or
-                GoogleCloudManager(google_project_id))
+        self.google_cloud_manager = google_cloud_manager or GoogleCloudManager(
+            google_project_id
+        )
         super(GoogleServiceAccountValidity, self).__init__(*args, **kwargs)
 
         # setup default values for error information, will get updated in
         # check_validity
         self._info["owned_by_project"] = None
         self._info["valid_type"] = None
         self._info["no_external_access"] = None
 
-    def check_validity(self, early_return=True,
-                       check_type_and_access=True, config=None):
+    def check_validity(
+        self, early_return=True, check_type_and_access=True, config=None
+    ):
 
         self.google_cloud_manager.open()
 
         google_managed_sa_domains = (
             config["GOOGLE_MANAGED_SERVICE_ACCOUNT_DOMAINS"] if config else None
@@ -520,13 +530,11 @@
             if not sa_exists:
                 return
 
             no_external_access = not (
                 service_account_has_external_access(
-                    self.account_id,
-                    self.google_cloud_manager,
-                    sa_policy_response
+                    self.account_id, self.google_cloud_manager, sa_policy_response
                 )
             )
             self.set("no_external_access", no_external_access)
             if not no_external_access and early_return:
                 return
```
</details></br>

fence/resources/google/access_utils.py
```diff
@@ -939,3 +940,17 @@ def force_add_service_accounts_to_access(
...
+    sa_policy_response = google_cloud_manager.get_service_account_policy(account)
+    if sa_policy_response.status_code != 200:
+        if sa_policy_response.status_code == 404:
+            sa_exists = False
+        else:
+            sa_exists = None
+    else:
+        sa_exists = True
+
+    return sa_exists, sa_policy_response
```
Avantol13: this has kinda a weird flow for non-404 errors. If we get a 403, we don't set sa_exists to True, which means we continue validating. and then we get to service_account_has_external_access and still return True, setting the misleading error message in the response. 

I think a cleaner flow may be something like: 
```
try:
    policy = (
        get_service_account_policy(
            self.account_id, self.google_cloud_manager)
    )
    sa_exists = True
except NotFound:
    sa_exists = False

self.set("exists", sa_exists)
if not sa_exists:
    return
```

and that way you don't have to deal with returning a tuple and clarifying the function name

fence/resources/google/access_utils.py
```diff
@@ -939,3 +940,17 @@ def force_add_service_accounts_to_access(
             add_user_service_account_to_google(
                 session, project_ids, google_project_id, sa
             )
+
+
+def get_service_account_policy_if_exists(account, google_cloud_manager):
```
Avantol13: sort of a misleading name since it's getting the policy if the *service account* exists, but it sorta reads like if the policy itself exists. Also, I would expect the return value to be just the policy by the function name (but it's a tuple() instead)

fence/resources/google/access_utils.py
Avantol13: you would do the 404 check still but instead of the logic for setting sa_exists within that function, you'd just raise a NotFound

tests/google/test_access_utils.py
```diff
@@ -183,6 +184,14 @@ def test_service_account_has_external_access_no_authorization(cloud_manager):
     )
 
 
+def test_service_account_does_not_exist(cloud_manager):
```
Avantol13: you should also consider writing a regression test for the core issue: when we get a 404 from google for service accounts `no_external_access` is set to True in the final response. check out the tests in /fence/tests/google/test_service_accounts.py. it might make sense to add some more there

Avantol13: to make sure no_external_access is NOT set to True and that we get the new sa_exists as True

fence/blueprints/google.py
```diff
@@ -699,11 +699,18 @@ def _get_service_account_email_error_status(validity_info):
...
-                ] = "Service account requested for registration is invalid."
+                if sa_validity['exists'] or sa_validity['exists'] is None:
+                    response["status"] = 403
+                    response["error"] = "unauthorized"
+                    response[
+                        "error_description"
+                    ] = "Service account requested for registration is invalid."
+                else:
+                    response["status"] = 404
+                    response["error"] = "Not Found"
```
Avantol13: these errors are client facing and should be somewhat consistent. the other errors use lower_case_underscore naming conventions . It probably makes sense to have a top-level enum for these errors so they're not hard-coded here too... we already have errors `monitor_not_found` and `project_not_found` so for consistency this could be `service_account_not_found`

fence/resources/google/access_utils.py
```diff
@@ -939,3 +940,12 @@ def force_add_service_accounts_to_access(
             add_user_service_account_to_google(
                 session, project_ids, google_project_id, sa
             )
+
+
+def get_service_account_policy(account, google_cloud_manager):
+
```
Avantol13: missing docstring

fence/resources/google/access_utils.py
```diff
@@ -939,3 +940,12 @@ def force_add_service_accounts_to_access(
...
+    sa_policy = google_cloud_manager.get_service_account_policy(account)
+    if sa_policy.status_code != 200:
+        raise NotFound
```
Avantol13: should provide a message in the NotFound error and you can use that when you catch it later

fence/resources/google/validity.py
```diff
@@ -517,9 +521,22 @@ def check_validity(
             if not valid_type and early_return:
                 return
 
+            try:
+                sa_policy = get_service_account_policy(
+                    self.account_id, self.google_cloud_manager)
+                sa_exists = True
```
Avantol13: you should move this check to the top and do this first. If you can't get the policy then you can set sa_exists to False and return early

fence/blueprints/google.py
```diff
@@ -699,11 +699,18 @@ def _get_service_account_email_error_status(validity_info):
     for sa_account_id, sa_validity in service_accounts_validity:
         if sa_account_id == validity_info.new_service_account:
             if not sa_validity:
-                response["status"] = 403
-                response["error"] = "unauthorized"
-                response[
-                    "error_description"
-                ] = "Service account requested for registration is invalid."
+                if sa_validity['exists'] or sa_validity['exists'] is None:
```
Avantol13: I think this should be `if not sa_validity['exists']`? Oh, wait. I'd flip this logic. Do 
```
                if not sa_validity['exists']:
                    # do 404 error
                else:
                    # do 403 error
```

tests/google/conftest.py
```diff
@@ -111,9 +111,24 @@ def valid_service_account_patcher():
...
+        patch(
+            "fence.resources.google.validity.get_service_account_policy",
+            get_policy_mock
+        )
+    )
+
     valid_type_mock.return_value = True
     external_access_mock.return_value = False
     from_google_project_mock.return_value = True
+    get_policy_mock.return_value = (True, None)
```
Avantol13: this can just be `None` right? doens't need to be a tuple?

PR #458, opened 2018-10-22
chore(template): add pull request template

philloooo:
add a pull request template that will automatically populate when a dev make a new PR. The developer will need to fill with appropriate content in each PR. 
We will automate the release generation to compile PR descriptions to release notes.

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -12,8 +12,8 @@ Description about what this pull request does.
 ### Improvements
 
 
-### Other Changes
-
-
 ### Dependency updates
+
+
+### Deployment changes
 
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,0 +1,19 @@
+Description about what this pull request does.
+
+### New Features
+N/A
+
+### Breaking Changes
+N/A
+
+### Bug Fixes
+N/A
+
+### Improvements
+N/A
+
+### Other Changes
+N/A
+
+### Dependency updates
+N/A
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,19 +1,19 @@
 Description about what this pull request does.
 
 ### New Features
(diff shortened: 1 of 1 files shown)


pull_request_template.md
```diff
@@ -0,0 +1,19 @@
...
+### Breaking Changes
+N/A
+
+### Bug Fixes
+N/A
+
+### Improvements
+N/A
+
+### Other Changes
```
Avantol13: what would go here? seems like everything should fit somewhere above

pull_request_template.md
```diff
@@ -0,0 +1,19 @@
+Description about what this pull request does.
+
+### New Features
```
Avantol13: what's the format here? just `-`s or a paragraph of text? Will this be defined in our wiki somewhere? seems like it should maybe be self-descriptive. e.g. instead of N/A just have an empty entry ?
```
### New Features
* 
```

Avantol13: Why third level heading?

pull_request_template.md
philloooo: I don't know, wasn't sure if all categories above caught everything

pull_request_template.md
philloooo: sure, updated that

philloooo: Cuz I feel <third level is too big for this...

philloooo:
the issue tracker is in pr title

pull_request_template.md
Avantol13: gotcha, wasn't sure if the parsing would expect that or not. doesn't really matter either way I guess

pull_request_template.md
Avantol13: hmm, yeah, guess it's safe to have a catch-all category just in case

PR #462, opened 2018-10-24
PXD-2080 fix(usersync): generate default project name

giangbui:
### Bug Fixes
Fence usersync gets confused if all the fields in the Project table don't match an update - SqlAlchemy should consider the project-name a primary key - something like that:
```
sqlalchemy.exc.IntegrityError: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely) (IntegrityError) duplicate key value violates unique constraint "project_name_key"
```

Just provide more intuitive message so that user can know how to fix it


PlanXCyborg:

PR #468, opened 2018-10-26
PXD-2228 Cronjob removes valid project fix

benJrohrer:


PlanXCyborg:

<details>

```diff
--- fence/resources/google/access_utils.py
+++ blackened
@@ -782,14 +782,17 @@
         # use configured time or 7 days
         expiration_time = int(time.time()) + flask.current_app.config.get(
             "GOOGLE_USER_SERVICE_ACCOUNT_ACCESS_EXPIRES_IN", 604800
         )
         logger.debug("time.time() is: {}".format(time.time()))
-        logger.debug("expiration window is: {}".format(
-            flask.current_app.config.get(
-                "GOOGLE_USER_SERVICE_ACCOUNT_ACCESS_EXPIRES_IN", 604800
-            )))
+        logger.debug(
+            "expiration window is: {}".format(
+                flask.current_app.config.get(
+                    "GOOGLE_USER_SERVICE_ACCOUNT_ACCESS_EXPIRES_IN", 604800
+                )
+            )
+        )
         logger.debug("expiration time is: {}".format(expiration_time))
 
         for access_group in bucket_access_groups:
             bucket_access = (
                 session.query(ServiceAccountToGoogleBucketAccessGroup)
```
</details></br>

fence/resources/google/validity.py
```diff
@@ -522,28 +542,64 @@ def check_validity(
...
+                policy_accessible = False
+                gcm.close()
+                return
+            finally:
+                self.set("policy_accessible", policy_accessible)
+        else:
+            policy_accessible = None
+            sa_policy = None
+
+        if check_access:
```
Avantol13: I would rename this everywhere to be `check_external_access`. it's a small but kinda important clarification in what it's actually checking. It's not checking the SA's access, it's checking if there's external access to the SA

fence/scripting/google_monitor.py
```diff
@@ -132,7 +136,27 @@ def _is_valid_service_account(sa_email, google_project_id, config=None):
...
+                check_type=True,
+                check_policy_accessible=True,
+                check_access=False,
+                config=config,
+            )
+        else:
+            sa_validity.check_validity(
+                early_return=True,
+                check_type=True,
+                check_policy_accessible=True,
```
Avantol13: can you just default all these checks to `True` in `check_validity` and then set to `False` where needed? maybe being explicit here is better though. idk, again, up to you, might be better to keep it explicit

fence/resources/google/validity.py
```diff
@@ -522,28 +542,64 @@ def check_validity(
...
+
+            if not policy_accessible:
+                logger.warning(
```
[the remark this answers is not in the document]
benJrohrer: It's possible that someone erroneously tells us check_external_access while simultaneously telling us to not check_policy_accessible. So I opted to warn instead. Do you think it still makes sense to return early?

fence/scripting/google_monitor.py
benJrohrer: They already default to True, but I included them explicitly in the function call. I can only include the Falses, wasn't sure if brevity or explicitness was better here.

fence/resources/google/validity.py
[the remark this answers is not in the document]
benJrohrer: I made the function take in a policy, but if its None, it makes a request to the API to get it.

fence/resources/google/validity.py
```diff
@@ -522,28 +542,65 @@ def check_validity(
...
+            gcm = self.google_cloud_manager
+        else:
+            self.google_cloud_manager.close()
+            try:
+                # check to see if we can access the project the SA belongs to
+                project_id = self.account_id.split("@")[-1].split(".")[0]
+                gcm = GoogleCloudManager(project_id)
+                gcm.open()
+            except Exception:
+                logger.debug("The Service Act.'s project couldn't be accessed")
```
Avantol13:
```suggestion
                logger.debug("Could not access the Google Project for Service Account {}. Unable to continue validity checking without access to the project, early exit.".format(self.account_id))
```

fence/resources/google/validity.py
```diff
@@ -522,28 +542,65 @@ def check_validity(
...
+                gcm.close()
+                return
+            finally:
+                self.set("policy_accessible", policy_accessible)
+
+        if check_external_access:
+
+            if not policy_accessible:
+                logger.warning(
+                    "Access check requires Service Account policy"
```
Avantol13: Maybe: 
```
"Invalid function use. External Access check requires Service Account Policy "
"& may fail if policy is not accessible. If you want to check external access, make sure you are also checking policy_accessible. "
```

PR #471, opened 2018-10-30
Google Project/Service Account Validity Checking

benJrohrer:
Add info on project/service account validity checking logic.


docs/google_architecture.md
```diff
@@ -156,6 +156,59 @@ In order to register a Service Account, *all* users in the Google Project must h
 
 ---
 
+#### Google Project and Service Account Validity Checking
+
+In order to register a Service Account, the Service Account and its associated Project must pass a series of validity checks.
+
+Projects are always validated against the following checks:
+* Fence monitor has access to project
```
Avantol13: when referring to a Google project, I'd recommend saying "Google Project" instead of just project, since project gets used so many other places to mean different things

docs/google_architecture.md
```diff
@@ -156,6 +156,59 @@ In order to register a Service Account, *all* users in the Google Project must h
...
+   * Key: `monitor_has_access`
+   * Checks if the Fence monitoring service account can access the project.
+* User has access to project
+   * Key: `user_has_access`
+   * Checks if the current user is an authorized member on the project.
+* Project has valid parent organization
+   * Key: `valid_parent_org`
+   * Checks if the project either has no parent organization, or if it does, it is included on the whitelist of parent. organizations (defined in Fence config).
```
Avantol13:
```suggestion
   * Checks if the project either has no parent organization, or if it does, it is included on the whitelist of parent organizations (defined in Fence config).
```

docs/google_architecture.md
```diff
@@ -156,6 +156,59 @@ In order to register a Service Account, *all* users in the Google Project must h
...
+* Project only has valid member types
+   * Key: `valid_member_types`
+   * Checks if the project ony has members that are User Accounts or Service Accounts.
```
Avantol13:
```suggestion
   * Checks if the project only has members that are User Accounts or Service Accounts.
```

docs/google_architecture.md
```diff
@@ -156,6 +156,59 @@ In order to register a Service Account, *all* users in the Google Project must h
...
+* Project members exist in fence
+   * Key: `members_exist_in_fence`
+   * Checks if the User members on the project exist in the fence DB
+* Project has valid service accounts
```
Avantol13: should _try_ to be consistent with whether or not you're capitalizing Service Account, but not really a big deal

docs/google_architecture.md
```diff
@@ -156,6 +156,59 @@ In order to register a Service Account, *all* users in the Google Project must h
...
+   * Key: `service_accounts`
+   * Checks if the Service Account members on the project pass the Service Account validity checks detailed below.
+  
+Service Accounts on the project, as well as the Service Account being registered, are validated against some combination of the following checks (which ones ultimately depend on the type of Service Account and whether or not the Service Account is currently being registered or not).
```
Avantol13:
```suggestion
Service Accounts on the project, as well as the Service Account being registered, are validated against some combination of the following checks (which checks occur ultimately depend on the type of Service Account and whether or not the Service Account is currently being registered or not).
```

docs/google_architecture.md
```diff
@@ -156,6 +156,59 @@ In order to register a Service Account, *all* users in the Google Project must h
...
+
+* Service Account is owned by Project identified in the request
+   * Key: `owned_by_project`
+   * Checks if the Service Account is owned by the project
+* Service Account is of valid type
```
Avantol13:
```suggestion
* Service Account is of valid type for registration
```

docs/google_architecture.md
```diff
@@ -156,6 +156,59 @@ In order to register a Service Account, *all* users in the Google Project must h
...
+   * Key: `valid_type`
+   * Checks if the Service Account is a User Managed Service Account or a Compute Engine API Service Account
+* Service Account policy is accessible
+   * Key: `policy_accessible`
+   * Checks if the Service Account's policy is accessible
```
Avantol13:
```suggestion
   * Checks if  fence's monitoring Service Account can read the Service Account's policy
```

docs/google_architecture.md
```diff
@@ -156,6 +156,59 @@ In order to register a Service Account, *all* users in the Google Project must h
...
+   * Key: `no_external_access`
+   * Checks if the Service Account has no IAM roles and no access keys
+
+These checks are applied according to the following logic:
+
+* If the service account is User Managed:
+   * check: `owned_by_project`, `valid_type`, `policy_accessible`, `no_external_access`
+* ElIf the service account is Google Managed:
+   * If the service account is registered or currently being registered:
+      * check: `owned_by_project`, `valid_type`, `policy_accessible`
```
Avantol13: maybe add a sub-bullet about why we don't check external access? like `we don't check external access (roles/keys) since the account is Google-managed and will have roles for use inside the Google Cloud Platform`

docs/google_architecture.md
```diff
@@ -156,6 +156,59 @@ In order to register a Service Account, *all* users in the Google Project must h
...
+   * ElIf the service account is not registered/being registered and is a member on the project:
+      * check: `owned_by_project`
```
Avantol13: another sub-bullet explanation maybe? `no need to check for valid type for registration since it isn't being registered. Also, some google-managed accounts do not allow reading their Policy, meaning that we cannot safely do policy_accessible/no_external_access checks. This is okay because we should, in general, assume that the google-managed accounts that are *not* registered for data access are okay so long as they're in the project security boundary.`

docs/google_architecture.md
```diff
@@ -156,6 +156,59 @@ In order to register a Service Account, *all* users in the Google Project must h
...
+
+Service Accounts are validated according to this logic regardless of whether or not it is during Service Account registration or the `google-manage-user-registrations` cronjob. However, during the cronjob validation, service accounts are validated before the project is validated.
```
Avantol13:
```suggestion
Service Accounts are validated according to this logic regardless of whether or not it is during Service Account registration or the `google-manage-user-registrations` cronjob. However, during the cronjob validation, registered service accounts are validated before their respective project is validated.

The Service Accounts are validated first in the cronjob so that if multiple SA's are registered and only one is non-conforming, we can remove that single account without invalidating the entire project. If the project as a whole is invalid, we must remove all SAs though.
```

docs/google_architecture.md
```diff
@@ -156,6 +156,63 @@ In order to register a Service Account, *all* users in the Google Project must h
...
+
+* If the service account is User Managed:
+   * check: `owned_by_project`, `valid_type`, `policy_accessible`, `no_external_access`
+* ElIf the service account is Google Managed:
+   * If the service account is registered or currently being registered:
+      * check: `owned_by_project`, `valid_type`, `policy_accessible`
+   * ElIf the service account is not registered/being registered and is a member on the project:
+      * check: `owned_by_project`
+      * Note: `valid_type` is not checked because the Service Account is not being registered. `valid_type` is checked when a Service Account is registered, and the type of a Service Account cannot change. `policy_accessible` is not checked because (1) some Google Managed service accounts do not allow their policies to be read and (2) we assume that the Google Managed accounts that are *not* registered for data access are okay so long as they're in the project security boundary.
+   * Note: no_external_access` is not checked, as a the Service Account is Google Managed and will therefore have roles for use within the Google Cloud Platform
```
Avantol13: missing a `

PR #477, opened 2018-11-02
pointing fence to cirrus branch fix/remove-member-400

benJrohrer:


PlanXCyborg:

fence/resources/google/access_utils.py
```diff
@@ -543,6 +543,22 @@ def force_remove_service_account_from_access(
...
+
+    Args:
+        service_account_email(str): service account to be removed from db
+        db(None, str): Optional db connection string
+    """
+    session = get_db_session(db)
+    service_account = (
+        session.query(UserServiceAccount).filter_by(email=service_account_email).first()
+    )
+    session.delete(service_account)
```
Avantol13: you should check if it exists and make this super safe. e.g. if it doesn't exist then we can consider it removed

fence/resources/google/validity.py
```diff
@@ -612,4 +603,21 @@ def check_validity(
...
 
+        # check if the SA is an allowed type
+        if check_type:
+
+            if not policy_accessible:
+                logger.warning(
+                    "Policy access was not checked. If the service account's "
+                    "policy is not accessible or the service account does not "
+                    "exist, this check may fail."
+                )
```
Avantol13: don't we need to return early here?

fence/scripting/google_monitor.py
```diff
@@ -202,15 +214,25 @@ def _get_service_account_removal_reasons(service_account_validity):
         return removal_reasons
 
     if service_account_validity["valid_type"] is False:
+        print("\tvalid_type: False")
```
Avantol13: you should remove these print statements, or at least change it to use the logger and be a bit more descriptive

tests/google/conftest.py
```diff
@@ -571,7 +573,12 @@ def invalid_service_account_patcher(db_session):
 
     def mock_is_valid(sa_email, *args, **kwargs):
         if sa_email == invalid_service_account:
-            return False
+            validity = GoogleServiceAccountValidity("account_id", "project_id")
+            # set overall validity to False
+            # set policy_accessible to True so the SA is not removed from the DB
```
Avantol13: in which test does it expect to be able to read that field? just the cronjob?

fence/scripting/google_monitor.py
```diff
@@ -64,15 +66,25 @@ def validation_check(db, config=None):
                 force_remove_service_account_from_access(
                     sa_email, google_project_id, db=db
                 )
+                if validity_info["policy_accessible"] is False:
+                    print(
```
Avantol13: should use logger here

fence/resources/google/validity.py
benJrohrer: I don't believe so. The only way we reach this is if policy_accessible is None (if it's False we would've returned early above). It its None, that simply means that the policy_access check was turned off. Doing the get_type check doesn't actually require the policy, its just that if the policy isn't accessible (policy_accessible = False, see above), we assume that the SA most likely doesn't exist. So, we can still do the get_type check, it may just fail if the SA doesn't exist. Maybe I could word it better ...

tests/google/conftest.py
benJrohrer: test_validation_check_one_invalid

If there is an invalid service account, we now make the distinction between "it was invalid because it doesn't exist (policy inaccessible)" and "it was invalid for other reasons". If it is invalid because it doesn't exist, we remove it from the fence database. However, in these test cases we really don't want to remove it from the fence db, we just want to check that it was removed from access. A unit test could probably be added for checking that when policy_accessible is false, the SA is removed from the database.

fence/resources/google/validity.py
Avantol13: ohhh, I see. You should maybe add a comment after the logger warning explaining why we don't want to return early. do we ever *not* check policy_accessible?

tests/google/conftest.py
Avantol13: yeah, I'd add that unit test for the new case

PR #500, opened 2018-11-15
docs(readme): Expand and rewrite to align with OAuth/OIDC terminology

roofridge:
1. The OAuth 2.0 RFC and the OIDC spec both establish some well-defined terms that have very particular meanings within those specs. This PR attempts to rewrite the Fence README so that its use of these terms aligns correctly with their definitions in these specs (where before there were a few overloaded or not-well-defined terms). 

2. It also adds further explanation for the diagrams and 

3. makes some phrasing edits for further clarity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,10 +8,23 @@ A `fence` separates protected resources from the outside world and allows
 only trusted entities to enter.
 
 Fence is a core service of the Gen3 stack that has multiple capabilities:
-1. Act as an auth broker to integrate with an [Identity Provider](#identity-provider) and provide downstream authentication and authorization for Gen3 services.
+1. Act as an [auth broker](#auth-broker) to integrate with one or more [IdPs](#IdP) and provide downstream authentication and authorization for Gen3 services.
 2. [Manage tokens](#token-management).
 3. Act as an [OIDC provider](#oidc--oauth2) to support external applications to use Gen3 services.
 4. [Issue short lived, cloud native credentials to access data in various cloud storage services](#accessing-data)
+
+
+## Contents
+ 
+1. [API Documentation](#API-documentation)
+1. [Terminologies](#Terminologies)
+1. [Identity Providers](#identity-provider) 
+1. [OIDC & OAuth2](#oidc--oauth2) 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -100,6 +135,8 @@ If a third-party application wants to use Gen3 resources like
 `fence`/`sheepdog`/`peregrine`, they call those services with an `Access Token`
 passed in an `Authorization` header.
 
+In the following flow, `3rd Party App` is the `RP`; `Protected Endpoint` is an endpoint of a Gen3 Resource (the `microservice`), and both of these are part of a `resource server`; and `Fence` is the `OP`. Here, importantly, `Fence` may be interfacing with another IdP _or_ with another `Fence` instance in order to implement the OIDC layer. Either way, note that the `Fence` blob in this diagram actually abstracts Fence in concert with some IdP, which may or may not also be Fence.
```
rudyardrichter: at the end here—note that the IdP which may or may not be fence would be a _different_ fence

README.md
```diff
@@ -8,12 +8,25 @@ A `fence` separates protected resources from the outside world and allows
 only trusted entities to enter.
 
 Fence is a core service of the Gen3 stack that has multiple capabilities:
-1. Act as an auth broker to integrate with an [Identity Provider](#identity-provider) and provide downstream authentication and authorization for Gen3 services.
+1. Act as an [auth broker](#auth-broker) to integrate with an [Identity Provider](#IdP) and provide downstream authentication and authorization for Gen3 services.
```
rudyardrichter: this is slightly outdated, it could actually say "one or more IdPs"

README.md
```diff
@@ -8,12 +8,25 @@ A `fence` separates protected resources from the outside world and allows
...
 2. [Manage tokens](#token-management).
 3. Act as an [OIDC provider](#oidc--oauth2) to support external applications to use Gen3 services.
 4. [Issue short lived, cloud native credentials to access data in various cloud storage services](#accessing-data)
 
 
+## Contents
```
rudyardrichter: :ok_hand:

roofridge: 😂

PR #515, opened 2018-12-07
Feat/data delete

rudyardrichter:
### New Features

Adds endpoint `DELETE /data/<id>` for deleting data file locations in s3/indexd for a given file.


fence/blueprints/data/indexd.py
```diff
@@ -239,6 +248,41 @@ def check_authorization(self, action):
...
+            locations_to_delete = self.indexed_file_locations
+        else:
+            locations_to_delete = [
+                location
+                for location in locations_to_delete
+                if location.url in urls
+            ]
+        for location in locations_to_delete:
+            bucket = location.bucket_name()
+            flask.current_app.boto.delete_data_file(bucket, self.file_id)
```
philloooo: can we handle the case that fence doesn't have permission and report error message about this fence instance doesn't have delete permission?

rudyardrichter: it does, see here: https://github.com/uc-cdis/fence/pull/515/files#diff-942c41520d0fb54ff40ca0148c1c6c74R47

PR #531, opened 2019-01-08
fix(upload-role): move arborist setup to yaml

rudyardrichter:
### Bug Fixes

The deployment with the previous fence will run into problems if arborist is rolled after fence, because that will discard the role for data upload which fence creates in it. The solution is to use the user.yaml file and have the fence sidecar to arborist handle the upload role setup.

### Deployment changes

Differences in user.yaml required for using with arborist: documentation in progress

### DIFF
--- a/fence/sync/sync_users.py
+++ b/fence/sync/sync_users.py
@@ -57,12 +57,162 @@ def download_dir(sftp, remote_dir, local_dir):
 
 
 def arborist_role_for_permission(permission):
+    """
+    For the programs/projects in the existing fence access control model, in order to
+    use arborist for checking permissions we generate a policy for each combination of
+    program/project and privilege. The roles involved all contain only one permission,
+    for one privilege from the project access model.
+    """
     return {
         "id": permission,
         "permissions": [
             {"id": permission, "action": {"service": "", "method": permission}}
         ],
     }
+
+
+@contextmanager
+def _read_file(filepath, encrypted=True, key=None, logger=None):
+    """
+    Context manager for reading and optionally decrypting file it only
+    decrypts files encrypted by unix 'crypt' tool which is used by dbGaP.
+
+    Args:
+        filepath (str): path to the file
+        encrypted (bool): whether the file is encrypted
+
+    Returns:
+        Generator[file-like class]: file like object for the file
+    """
+
+    if encrypted:
+        has_crypt = sp.call(["which", "crypt"])
+        if has_crypt != 0:
+            if logger:
+                logger.error("Need to install crypt to decrypt files from dbgap")
+            exit(1)
+        p = sp.Popen(
+            ["crypt", key],
+            stdin=open(filepath, "r"),
+            stdout=sp.PIPE,
+            stderr=open(os.devnull, "w"),
+        )
+        yield StringIO(p.communicate()[0])
+    else:
+        f = open(filepath, "r")
+        yield f
+        f.close()
+
+
+class UserYAML(object):
+    """
+    Representation of the information in a YAML file describing user, project, and RBAC
+    information for access control.
+    """
+
+    def __init__(
+        self, projects=None, user_info=None, policies=None, rbac=None, logger=None,
+        user_rbac=None,
+    ):
+        self.projects = projects or {}
+        self.user_info = user_info or {}
+        self.user_rbac = user_rbac or {}
+        self.policies = policies or {}
+        self.rbac = rbac or {}
+        self.logger = logger
+
+
+    @classmethod
+    def from_file(cls, filepath, encrypted=True, key=None, logger=None):
+        with _read_file(filepath, encrypted=encrypted, key=key, logger=logger) as f:
+            data = yaml.safe_load(f)
+        projects = dict()
+        user_info = dict()
+        policies = dict()
+
+        users = data.get("users", {})
+        for username, details in users.iteritems():
+            # users should occur only once each; skip if already processed
+            if username in projects:
+                if logger:
+                    logger.error("user `{}` occurs multiple times".format(username))
+                raise EnvironmentError("invalid yaml file")
+
+            privileges = {}
+            try:
+                for project in details.get("projects", {}):
+                    privileges[project["auth_id"]] = set(project["privilege"])
+            except KeyError as e:
+                if logger:
+                    logger.error("project missing field: {}".format(e))
+                continue
+
+            user_info[username] = {
+                "email": details.get("email", username),
+                "display_name": details.get("display_name", ""),
+                "phone_number": details.get("phone_number", ""),
+                "tags": details.get("tags", {}),
+                "admin": details.get("admin", False),
+            }
+            projects[username] = privileges
+
+            # list of policies we want to grant to this user, which get sent to arborist
+            # to check if they're allowed to do certain things
+            policies[username] = details.get("policies", [])
+
+
+        # resources should be the resource tree to construct in arborist
+        user_rbac = dict()
+        for username, user_detail in users.iteritems():
+            # users should occur only once each; skip if already processed
+            if username in user_rbac:
+                msg = "invalid yaml file: user `{}` occurs multiple times".format(
+                    username
+                )
+                logger.error(msg)
+                raise EnvironmentError(msg)
+            resource_permissions = dict()
+            for project in user_detail.get("projects", {}):
+                # project may not have `resource` field
+                try:
+                    resource = project["resource"]
+                except KeyError:
+                    continue
+                resource_permissions[resource] = set(project["privilege"])
+            user_rbac[username] = resource_permissions
+
+        rbac = data.get("rbac")
+
+        return cls(
+            projects=projects,
+            user_info=user_info,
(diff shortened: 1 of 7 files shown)


PlanXCyborg:

<details>

```diff
--- fence/sync/sync_users.py
+++ blackened
@@ -109,21 +109,25 @@
     Representation of the information in a YAML file describing user, project, and RBAC
     information for access control.
     """
 
     def __init__(
-        self, projects=None, user_info=None, policies=None, rbac=None, logger=None,
+        self,
+        projects=None,
+        user_info=None,
+        policies=None,
+        rbac=None,
+        logger=None,
         user_rbac=None,
     ):
         self.projects = projects or {}
         self.user_info = user_info or {}
         self.user_rbac = user_rbac or {}
         self.policies = policies or {}
         self.rbac = rbac or {}
         self.logger = logger
 
-
     @classmethod
     def from_file(cls, filepath, encrypted=True, key=None, logger=None):
         with _read_file(filepath, encrypted=encrypted, key=key, logger=logger) as f:
             data = yaml.safe_load(f)
         projects = dict()
@@ -157,11 +161,10 @@
             projects[username] = privileges
 
             # list of policies we want to grant to this user, which get sent to arborist
             # to check if they're allowed to do certain things
             policies[username] = details.get("policies", [])
-
 
         # resources should be the resource tree to construct in arborist
         user_rbac = dict()
         for username, user_detail in users.iteritems():
             # users should occur only once each; skip if already processed
@@ -384,11 +387,13 @@
                 continue
             if not self._match_pattern(filepath, encrypted=encrypted):
                 continue
 
             dbgap_key = getattr(self, "dbgap_key", None)
-            with _read_file(filepath, encrypted=encrypted, key=dbgap_key, logger=self.logger) as f:
+            with _read_file(
+                filepath, encrypted=encrypted, key=dbgap_key, logger=self.logger
+            ) as f:
                 csv = DictReader(f, quotechar='"', skipinitialspace=True)
                 for row in csv:
                     username = row.get("login", "")
                     if username == "":
                         continue
@@ -463,11 +468,13 @@
         user_project = dict()
         user_info = dict()
         user_policies = dict()
 
         dbgap_key = getattr(self, "dbgap_key")
-        with _read_file(filepath, encrypted=encrypted, key=dbgap_key, logger=self.logger) as f:
+        with _read_file(
+            filepath, encrypted=encrypted, key=dbgap_key, logger=self.logger
+        ) as f:
             data = yaml.safe_load(f)
 
         users = data.get("users", {})
         for username, details in users.iteritems():
             # users should occur only once each; skip if already processed
@@ -949,13 +956,11 @@
             dict(zip(local_csv_file_list, permissions)), encrypted=False, sess=sess
         )
 
         try:
             user_yaml = UserYAML.from_file(
-                self.sync_from_local_yaml_file,
-                encrypted=False,
-                logger=self.logger,
+                self.sync_from_local_yaml_file, encrypted=False, logger=self.logger
             )
             resources = user_yaml.rbac["resources"]
         except EnvironmentError as e:
             self.logger.error(str(e))
             self.logger.error("aborting early")
```
</details></br>

fence/sync/sync_users.py
```diff
@@ -65,6 +71,150 @@ def arborist_role_for_permission(permission):
...
+            projects[username] = privileges
+
+            # list of policies we want to grant to this user, which get sent to arborist
+            # to check if they're allowed to do certain things
+            policies[username] = details.get("policies", [])
+
+
+        # resources should be the resource tree to construct in arborist
+        user_rbac = dict()
+        for username, user_detail in users.iteritems():
```
Avantol13:
```suggestion
        for username, details in users.iteritems():
```

minor consistency change to match above loop

fence/sync/sync_users.py
```diff
@@ -65,6 +71,150 @@ def arborist_role_for_permission(permission):
...
+            # users should occur only once each; skip if already processed
+            if username in user_rbac:
+                msg = "invalid yaml file: user `{}` occurs multiple times".format(
+                    username
+                )
+                logger.error(msg)
```
Avantol13: if you're passing this in/defaulting to `None`, you'll need an `if logger:` here

fence/sync/sync_users.py
```diff
@@ -65,6 +71,150 @@ def arborist_role_for_permission(permission):
...
+            for project in user_detail.get("projects", {}):
+                # project may not have `resource` field
+                try:
+                    resource = project["resource"]
+                except KeyError:
+                    continue
+                resource_permissions[resource] = set(project["privilege"])
+            user_rbac[username] = resource_permissions
+
+        rbac = data.get("rbac")
```
Avantol13: not backwards compatible with previous versions of user yaml since `resources` moved to `rbac` <- more of a note to self, not a real comment

fence/sync/sync_users.py
```diff
@@ -65,6 +71,150 @@ def arborist_role_for_permission(permission):
...
+            projects=projects,
+            user_info=user_info,
+            user_rbac=user_rbac,
+            policies=policies,
+            rbac=rbac,
+            logger=logger,
+        )
+
+    @property
+    def user_arborist_info(self):
```
Avantol13: isn't this duplicated logic from above? do you need this?

fence/sync/sync_users.py
Avantol13: later you expect a `resources` field in `rbac`. perhaps you should default that here to prevent key error later? Or catch error later, I'll add a comment there too

fence/sync/sync_users.py
```diff
@@ -874,12 +950,12 @@ def _sync(self, sess):
...
-                self.sync_from_local_yaml_file, encrypted=False
-            )
-            user_arborist_info, resources = self._parse_resources_from_yaml(
-                self.sync_from_local_yaml_file, encrypted=False
+            user_yaml = UserYAML.from_file(
+                self.sync_from_local_yaml_file,
+                encrypted=False,
+                logger=self.logger,
             )
+            resources = user_yaml.rbac["resources"]
```
Avantol13: Should safely handle if that key doesn't exist. Either default it when creating the UserYaml instance, catch KeyError, or default here with a `.get`. Not sure what makes the most sense

fence/sync/sync_users.py
```diff
@@ -65,6 +71,150 @@ def arborist_role_for_permission(permission):
...
+        self.user_info = user_info or {}
+        self.user_rbac = user_rbac or {}
+        self.policies = policies or {}
+        self.rbac = rbac or {}
+        self.logger = logger
+
+
+    @classmethod
+    def from_file(cls, filepath, encrypted=True, key=None, logger=None):
+        with _read_file(filepath, encrypted=encrypted, key=key, logger=logger) as f:
```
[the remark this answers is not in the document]
rudyardrichter: if `_read_file` fails right now either it will be with `EnvironmentError` (which is caught at 958), or it does an `exit(1)` (not pretty...j

fence/sync/sync_users.py
```diff
@@ -65,6 +71,150 @@ def arborist_role_for_permission(permission):
...
+                    logger.error("user `{}` occurs multiple times".format(username))
+                raise EnvironmentError("invalid yaml file")
+
+            privileges = {}
+            try:
+                for project in details.get("projects", {}):
+                    privileges[project["auth_id"]] = set(project["privilege"])
+            except KeyError as e:
+                if logger:
+                    logger.error("project missing field: {}".format(e))
```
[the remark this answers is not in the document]
rudyardrichter: as it's a classmethod the logger doesn't exist on the thing yet—`from_file` is basically an `__init__` alternative

fence/sync/sync_users.py
```diff
@@ -344,8 +464,9 @@ def _parse_yaml(self, filepath, encrypted=True):
         user_info = dict()
         user_policies = dict()
 
-        with self._read_file(filepath, encrypted=encrypted) as stream:
-            data = yaml.safe_load(stream)
+        dbgap_key = getattr(self, "dbgap_key")
```
Avantol13:
```suggestion
        dbgap_key = getattr(self, "dbgap_key", None)
```

another minor consistency thing to match above

fence/sync/sync_users.py
rudyardrichter: true, no reason for this not to try the old one. will update it, thanks

fence/sync/sync_users.py
rudyardrichter: yep, this died. thanks for pointing out

fence/sync/sync_users.py
rudyardrichter: will update so the `UserYAML` handles any case, and the rest of the code all gets it from `user_yaml.rbac["resources"]`

fence/sync/sync_users.py
rudyardrichter: updated to be completely backwards compatible, see a few lines below here

fence/sync/sync_users.py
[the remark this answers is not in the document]
Avantol13: ah, duh. Missed that

PR #533, opened 2019-01-11
Feat/no google plus

Avantol13:
### Improvements
- Google Plus API No longer required, uses OIDC UserInfo endpoint instead

### DIFF
--- a/fence/resources/openid/google_oauth2.py
+++ b/fence/resources/openid/google_oauth2.py
@@ -1,11 +1,21 @@
 from oauth2client.client import OAuth2WebServerFlow
+
 import httplib2
 import json
+import requests
 
 
 class Oauth2Client(object):
-    """client for interacting with google oauth 2,
-    as google openid connect is supported under oauth2"""
+    """
+    client for interacting with google oauth 2,
+    as google openid connect is supported under oauth2
+
+    https://developers.google.com/api-client-library/python/guide/aaa_oauth
+    """
+
+    GOOGLE_DISCOVERY_URL = (
+        "https://accounts.google.com/.well-known/openid-configuration"
+    )
 
     def __init__(self, settings, logger, HTTP_PROXY=None):
         self.logger = logger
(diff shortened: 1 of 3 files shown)


PlanXCyborg:

fence/resources/openid/google_oauth2.py
```diff
@@ -56,3 +67,14 @@ def get_user_id(self, code):
...
             return {"error": "Can't get your email: {}".format(e)}
+
+    def get_userinfo_endpoint(self):
+        """
+        Return the url for Google's Userinfo endpoint by the recommended method of
+        using their discovery url. Default to current userinfo url as identified
+        09 JAN 2019.
+        """
+        document = requests.get(self.GOOGLE_DISCOVERY_URL)
+        return document.json().get(
```
zflamig: Can we check for a few cases and have warning logs here?
1. if we don't get a 200 back, we should log and then print what we are defaulting to
2. if we get 200 back, but userinfo_endpoint isn't present we should log that + what we default to
3. we get 200 back, but the userinfo_endpoint differs from what we default to

When I check the discovery endpoint right now I see:
"userinfo_endpoint": "https://openidconnect.googleapis.com/v1/userinfo"
too?

zflamig:
Is this a breaking change in that sense that we need to generate new API keys from Google for all our commons?

fence/resources/openid/google_oauth2.py
Avantol13: Yeah, I can add logging for those cases. I just double checked the endpoint, I see the same as you. I think I was using the oauth2 userinfo by default before

Avantol13:
> Is this a breaking change in that sense that we need to generate new API keys from Google for all our commons?

I don't believe so. We have client_id/secrets setup for oauth2 flow for all the envs, which just need to be given valid redirects (and this PR does not change those).

For oauth2 endpoints, we [don't need to pass an API Key](https://support.google.com/googleapi/answer/6158862?hl=en) along with requests. The API keys we do have are for hitting other Google endpoints for Project management, Group creation, etc. to support the different Data Access methods we have for Google.

rudyardrichter:
it works fine for me too with zero deployment changes, and code LGTM.

PlanXCyborg:
Jenkins Build 7 : time taken 1 hr 0 min

PlanXCyborg:
Jenkins Build 8 : time taken 1 hr 0 min

PlanXCyborg:
Jenkins Build 9 : time taken 1 min 7 sec

PlanXCyborg:
Jenkins Build 10 : time taken 57 min

PlanXCyborg:
Jenkins Build 11 : time taken 52 min

PR #537, opened 2019-01-17
Update swagger.yaml

vpsx:
add 'gs' to available protocols for /data/download


PlanXCyborg:
Jenkins Build 1 : time taken 33 min

PR #540, opened 2019-01-24
fix(redirect): validate login/out redirect URLs

rudyardrichter:
Validate redirect URLs provided to the login and logout endpoints. By default only redirects back to the same domain as fence or for saved OAuth clients are allowed.

### Improvements

- The login endpoints now verify that the redirect provided is valid according to fence: it redirects back to the Gen3 application, to an OAuth client, or to some other approved URL from the configuration.
Deployment changes
- Added optional new LOGIN_REDIRECT_WHITELIST config variable to allow for redirecting to specific domains for login/logout.


PlanXCyborg:
Jenkins Build 1 : time taken 46 sec

PlanXCyborg:

PlanXCyborg:
Jenkins Build 3 : time taken 46 sec

PlanXCyborg:
Jenkins Build 4 : time taken 48 sec

PlanXCyborg:
Jenkins Build 5 : time taken 37 sec

PlanXCyborg:
Jenkins Build 6 : time taken 40 sec

PlanXCyborg:
Jenkins Build 7 : time taken 39 sec

PlanXCyborg:
Jenkins Build 8 : time taken 40 sec

PlanXCyborg:
Jenkins Build 9 : time taken 1 min 18 sec

PlanXCyborg:
Jenkins Build 10 : time taken 1 min 0 sec

PlanXCyborg:
Jenkins Build 12 : time taken 39 min

fence/blueprints/login/redirect.py
```diff
@@ -0,0 +1,48 @@
...
+folder).
+"""
+
+import flask
+
+from fence.blueprints.login.utils import allowed_login_redirects, domain
+from fence.errors import UserError
+
+
+def validate_redirect(url, oauth_client=None):
```
Avantol13: is the oauth_client necessary? the allowed redirects already have client domains in there. also, I don't see the arg used anywhere

fence/blueprints/login/redirect.py
```diff
@@ -0,0 +1,48 @@
...
+    if oauth_client:
+        invalid_redirect = domain(url) not in map(
+            domain, oauth_client.redirect_uris
+        )
+    else:
+        invalid_redirect = domain(url) not in allowed_redirects
+
+    if invalid_redirect:
+        flask.current_app.logger.error(
+            "invalid redirect {}. expected one of: {}".format(
```
Avantol13: may want to change this wording (and perhaps elsewhere too) to convey that the domains are what we're checking against, so the redirect doesn't have to exactly match one in the `allowed_redirects`

fence/config-default.yaml
```diff
@@ -199,6 +199,11 @@ SESSION_ALLOWED_SCOPES:
 #   Sibboleth? Use: '{{BASE_URL}}/login/shib'
 DEFAULT_LOGIN_URL: '{{BASE_URL}}/login/google'
 
+# `LOGIN_REDIRECT_WHITELIST` is a list of extra whitelisted URLs which can be redirected
```
Avantol13: this is perhaps misleading, it's not full URLs that are being whitelisted, just the domains. I think that should be made clear here. overly verbose me might rename to `LOGIN_REDIRECT_DOMAIN_WHITELIST` but I think modifying the comment would be sufficient

rudyardrichter: `allowed_login_redirects` returns domains too. You can give it a full URL and it'll trim it to the same thing it's checking in the redirects: https://github.com/uc-cdis/fence/pull/540/files#diff-634a45d7b35d181a54ef9e3ee4344f9eR25

fence/blueprints/login/redirect.py
rudyardrichter: Same idea as the other `allowed_redirects` thing, those should actually be domains

fence/config-default.yaml
Avantol13: yeah, I guess what I'm trying to say is that a person configuring this may not be aware of that logic and think that this is a list of exact whitelisted redirect urls and add a ton of paths in the same domain they don't really need

rudyardrichter: oh, yep. I can add some more info

PR #543, opened 2019-01-25
fix(logging): only disable debug by default

rudyardrichter:
Set default log level to INFO; verbose includes DEBUG


PlanXCyborg:
Jenkins Build 1 : time taken 18 min

PR #547, opened 2019-01-29
fix(refresh): handle scope in refresh token request

Avantol13:
### Bug Fixes
- handle scope param when hitting oauth2/token for token refresh

### DIFF
--- a/tests/rfc6749/test_oauth2.py
+++ b/tests/rfc6749/test_oauth2.py
@@ -40,36 +40,22 @@ def test_oauth2_token_post_public_client(oauth_test_client_public):
     oauth_test_client_public.token()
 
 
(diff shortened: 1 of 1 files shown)


PlanXCyborg:

tests/rfc6749/test_oauth2.py
```diff
@@ -48,6 +48,14 @@ def test_oauth2_token_refresh(oauth_test_client):
     oauth_test_client.refresh()
 
 
+def test_oauth2_token_refresh_with_scope(oauth_test_client):
```
philloooo: let's use the pytest parameterize for with or without scope?

Avantol13: ah, good idea. will update

PR #548, opened 2019-01-29
fix(refresh): don't provide new refresh token when hitting token endp…

Avantol13:
### Bug Fixes
- don't provide new refresh token when hitting token endpoint with `refresh_token` grant type

### DIFF
--- a/fence/oidc/jwt_generator.py
+++ b/fence/oidc/jwt_generator.py
@@ -60,7 +60,7 @@ def generate_implicit_response(
     user=None,
     scope=None,
     nonce=None,
-    **kwargs,
+    **kwargs
 ):
     # prevent those bothersome "not bound to session" errors
     if user not in current_session:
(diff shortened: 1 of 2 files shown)


PlanXCyborg:

fence/oidc/jwt_generator.py
```diff
@@ -46,8 +46,12 @@ def generate_token(client, grant_type, **kwargs):
             claims (to avoid having to encode or decode the refresh token
             here)
     """
-    if grant_type == "authorization_code" or grant_type == "refresh_token":
+    if grant_type == "authorization_code":
         return generate_token_response(client, grant_type, **kwargs)
+    elif grant_type == "refresh_token":
+        tokens = generate_token_response(client, grant_type, **kwargs)
+        tokens["refresh_token"] = ""
```
rudyardrichter: if we want clients to keep using this the same way then this should return the same refresh token

PlanXCyborg:
Jenkins Build 1 : time taken 38 min

PR #555, opened 2019-02-01
Feat/lazy session

paulineribeyre:
Do not create a fence cookie for anonymous sessions. Because this breaks the windmill integration tests, also add logic to facilitate integration test login.

### Improvements
- no cookie creation when it's not needed
- add a dev cookie to fence for easy login into windmill during integration tests


PlanXCyborg:
Jenkins Build 2 : time taken 54 min

PlanXCyborg:
Jenkins Build 3 : time taken 23 min

PlanXCyborg:
Jenkins Build 4 : time taken 39 min

fence/resources/user/user_session.py
```diff
@@ -59,8 +59,10 @@ def __init__(self, session_token):
                 # if session token is invalid, create a new
                 # empty one silently
                 jwt_info = self._get_initial_session_token()
-        else:
+        elif access_token:
             jwt_info = self._get_initial_session_token()
+        else:
+            jwt_info = None
```
Avantol13:
```suggestion
            jwt_info = {"context": {}}
```
this is what it sets it to when `clear` is called. things like https://github.com/uc-cdis/fence/blob/master/fence/resources/user/user_session.py#L334 will raise exception if this is `None`

paulineribeyre: oooooh thank you

PR #558, opened 2019-02-05
Feat/google expire

paulineribeyre:
Add `expires_in` parameter to the POST /credentials/google endpoint

### New Features
- POST /credentials/google endpoint: add optional expires_in param

### Improvements
- GET /credentials/google endpoint: show the actual expiration set in our DB


PlanXCyborg:

<details>

```diff
--- fence/blueprints/storage_creds/google.py
+++ blackened
@@ -72,21 +72,23 @@
             keys = g_cloud_manager.get_service_account_keys_info(
                 service_account.google_unique_id
             )
 
             # replace Google's expiration date by the one in our DB
-            reg = re.compile(".+\/keys\/(.+)") # get key_id from xx/keys/key_id
+            reg = re.compile(".+\/keys\/(.+)")  # get key_id from xx/keys/key_id
             for k in keys:
                 key_id = reg.findall(k["name"])[0]
                 db_entry = (
                     current_session.query(GoogleServiceAccountKey)
                     .filter_by(key_id=key_id)
                     .first()
                 )
 
                 # convert timestamp to date - use the same format as Google API
-                exp_date = datetime.utcfromtimestamp(db_entry.expires).strftime("%Y-%m-%dT%H:%M:%SZ")
+                exp_date = datetime.utcfromtimestamp(db_entry.expires).strftime(
+                    "%Y-%m-%dT%H:%M:%SZ"
+                )
 
                 k["validBeforeTime"] = exp_date
 
             result = {"access_keys": keys}
 
@@ -151,12 +153,11 @@
         different mechanism than the Client SAs was required.
         """
         # x days * 24 hr/day * 60 min/hr * 60 s/min = y seconds
         max_expire = cirrus_config.SERVICE_KEY_EXPIRATION_IN_DAYS * 24 * 60 * 60
         expires_in = min(
-            int(flask.request.args.get("expires_in", max_expire)),
-            max_expire
+            int(flask.request.args.get("expires_in", max_expire)), max_expire
         )
 
         expiration_time = int(time.time()) + int(expires_in)
         key_id = key.get("private_key_id")
         add_custom_service_account_key_expiration(
```
</details></br>

fence/blueprints/storage_creds/google.py
```diff
@@ -71,6 +72,24 @@ def get(self):
...
+            for k in keys:
+                key_id = reg.findall(k["name"])[0]
+                db_entry = (
+                    current_session.query(GoogleServiceAccountKey)
+                    .filter_by(key_id=key_id)
+                    .first()
+                )
+
+                # convert timestamp to date - use the same format as Google API
+                exp_date = datetime.utcfromtimestamp(db_entry.expires).strftime(
```
Avantol13: if db_entry is `None` we'll want to do some error handling and then raise an exception here. We should catch it, log an error with some details, and then re-raise (since this is an unhandled exception, there should always be entries in our db for these keys). in the log we can add details like the service account name/email in google, key ids we got from google, and that we couldn't find an entry in our db. Actually, as part of the error handling, we should probably remove the keys from Google too (since they could have had access for longer than intended if we're not tracking in our db). you can look at code in the delete endpoint for the cirrus calls for that.

PR #559, opened 2019-02-06
fix(sa-expiration): actually expire user-issued keys, even if there a…

Avantol13:
…re no client-issued keys

### Bug Fixes
- handle the expiration of user-issued google service account keys when there are no client-issued keys


PlanXCyborg:

PR #566, opened 2019-02-12
Feat/google link expire

paulineribeyre:
Add optional `expires_in` parameter to the GET and PATCH /link/google endpoints

### New Features
- add `expires_in` param to `_link_google_account`, which is used by GET /link/google. the param is stored in cookie field `google_link_expires_in`
- `_force_update_user_google_account`, which is used by PATCH /link/google and GET /link/google/callback, gets the param from the cookie

### DIFF
--- a/fence/blueprints/link.py
+++ b/fence/blueprints/link.py
@@ -112,10 +112,16 @@ class GoogleLinkRedirect(Resource):
             # save off provided redirect in session and initiate Google AuthN
             flask.session["redirect"] = provided_redirect
 
+            # time (in seconds) during which the link will be valid
+            max_expire = config["GOOGLE_ACCOUNT_ACCESS_EXPIRES_IN"]
+            expires_in = min(
+                int(flask.request.args.get("expires_in", max_expire)), max_expire
+            )
+
             # if we're mocking Google login, skip to callback
             if config.get("MOCK_GOOGLE_AUTH", False):
                 flask.redirect_url = (
-                    config["BASE_URL"].strip("/") + "/link/google/callback?code=abc"
+                    config["BASE_URL"].strip("/") + "/link/google/callback?code=abc&expires_in=" + str(expires_in)
                 )
                 response = flask.redirect(flask.redirect_url)
                 # pass-through the authorization header. The user's username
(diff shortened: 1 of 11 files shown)


PlanXCyborg:

fence/blueprints/link.py
```diff
@@ -442,7 +456,13 @@ def _force_update_user_google_account(
                 "was attempted and failed."
             )
 
+    # timestamp at which the link will expire
     expiration = get_default_google_account_expiration()
+    if "google_link_expires_in" in flask.session:
+        requested_expiration = int(time.time()) + flask.session.pop(
+            "google_link_expires_in", None
```
Avantol13:
```suggestion
            "google_link_expires_in", expiration-int(time.time())
```

I don't think Python likes adding `None` to an int

Avantol13: oh, but you already check if it's in the session.. so I think you can just pop it safely without a default

Avantol13:
```suggestion
            "google_link_expires_in"
```

paulineribeyre: pop requires a default -> will fix that separately

paulineribeyre: `_clear_google_link_info_from_session` is called before the param is used here: https://github.com/uc-cdis/fence/blob/master/fence/blueprints/link.py#L273-L284
I can either leave it as it is (since it's popped as soon as it's used it should be okay?) or add it as parameter to `_force_update_user_google_account`

fence/blueprints/storage_creds/google.py
```diff
@@ -171,11 +172,16 @@ def handle_user_service_account_creds(self, key, service_account):
...
-        expires_in = min(
-            int(flask.request.args.get("expires_in", max_expire)), max_expire
-        )
+        expires_in = cirrus_config.SERVICE_KEY_EXPIRATION_IN_DAYS * 24 * 60 * 60
+        if "expires_in" in flask.request.args:
+            try:
+                requested_expires_in = int(flask.request.args["expires_in"])
+                assert requested_expires_in > 0
+                expires_in = min(expires_in, requested_expires_in)
+            except:
```
Avantol13:
```suggestion
            except Exception:
```

PR #567, opened 2019-02-13
feat(google-creds): allow a user/client to delete all keys owned by t…

Avantol13:
…hem with one api call

### New Features
- /credentials/google/* endpoint allows users/clients to remove all temporary google keys owned by them, without them needing to know key_ids

### DIFF
--- a/fence/blueprints/storage_creds/google.py
+++ b/fence/blueprints/storage_creds/google.py
@@ -284,9 +284,19 @@ def _delete_service_account_key(g_cloud, service_account_id, access_key):
     Internal function for deleting a given key for a service account, also
     removes entry from our db if it exists
     """
-    response = g_cloud.delete_service_account_key(service_account_id, access_key)
+    try:
+        response = g_cloud.delete_service_account_key(service_account_id, access_key)
+    except Exception:
+        logger.debug(
+            "Deleting service account {} key {} from Google FAILED. Response: {}. "
+            "We did NOT delete it from our DB either.".format(
+                service_account_id, access_key, str(response)
+            )
+        )
+        raise
+
     logger.debug(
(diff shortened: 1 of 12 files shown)


PlanXCyborg:

fence/blueprints/storage_creds/google.py
```diff
@@ -235,3 +232,20 @@ def delete(self, access_key):
...
 
         return "", 204
+
+
+def _delete_service_account_key(g_cloud, service_account_id, access_key):
+    """
+    Internal function for deleting a given key for a service account, also
+    removes entry from our db if it exists
+    """
+    g_cloud.delete_service_account_key(service_account_id, access_key)
```
rudyardrichter: probably good to check & log the response from the delete

openapis/swagger.yaml
```diff
@@ -481,7 +481,7 @@ paths:
         - name: access_key
           required: true
           in: path
-          description: Google Key ID
+          description: Google Key ID. Alternatively, you can provide * to delete ALL keys owned by you.
```
rudyardrichter: the `*` thing feels a little funny, I'm not positive if there's any convention for this but I would think just `DELETE /credentials/google/` should have the same meaning as this.

Avantol13: I did think about that and tried to search for convention, I'm workied about DELETE /credentials/google because if clients have code to delete keys and somehow accidentally send an empty string, it would clear everything. So I figured it would make it more accident-proof if DELETE all required an explicit string but I'm open for debate

fence/blueprints/storage_creds/google.py
Avantol13: ah, good point, will add some logging here

PR #569, opened 2019-02-14
fix(google-sa): use email internally as it is globally unique in goog…

Avantol13:
…le and "google_unique_id" is only unique within a Google Project

Issues arise when using a service accounts "uniqueID" because it is ONLY unique within a single Google project. Also, if you recreate the same service account (e.g. same email address) you'll end up with a different uniqueID.

Modify code to use "email" internally to avoid issues, since the email *is* globally unique. Also use "email" in calls to Google API so that if it gets deleted from Google and recreated, the api calls will still work (previously was using uniqueID, which will have changed with a delete/recreate).

### Improvements
- Better handling of service accounts by using their globally unique email internally
- Don't create duplicate entries in edge table ProjectToBucket

### DIFF
--- a/fence/resources/google/utils.py
+++ b/fence/resources/google/utils.py
@@ -265,15 +265,6 @@ def add_custom_service_account_key_expiration(
     current_session.commit()
 
 
-def get_or_create_service_account(client_id, user_id, username, proxy_group_id):
-    # underlying cloud api library effectively handles conflicts now without error
-    service_account = create_service_account(
-        client_id, user_id, username, proxy_group_id
-    )
-
-    return service_account
-
-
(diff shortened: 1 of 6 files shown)


PlanXCyborg:

fence/resources/google/utils.py
```diff
@@ -266,12 +266,10 @@ def add_custom_service_account_key_expiration(
 
 
 def get_or_create_service_account(client_id, user_id, username, proxy_group_id):
```
philloooo: remove this function and rename the `create_service_account` to this?

Avantol13: fixed in bb1d792

PR #571, opened 2019-02-18
fix(white-list)

paulineribeyre:
Fixes this error:
`[  DEBUG] Removing whitelisted SAs None from the SAs on the project xxx.`
`ERROR in error_handler: Catch exception`
`   File "/fence/fence/resources/google/access_utils.py", line 1001, in remove_white_listed_service_account_ids`
`     for email in white_listed_sa_emails:`
 `TypeError: 'NoneType' object is not iterable`

### Bug Fixes
- if the config var `WHITE_LISTED_SERVICE_ACCOUNT_EMAILS` is not set, default to an empty list

PR #577, opened 2019-02-20
fix(config): parse from config file if not passed in

Avantol13:
https://github.com/uc-cdis/fence/pull/571 defaulted config to empty list but https://github.com/uc-cdis/fence/blob/9f35d0af0c50d840a93c704c9fc05d34d3535e5c/fence/resources/google/access_utils.py#L976 expects it to be None to default to the configuration file. 

overall confusing logic because previously a config was passed through (this is an ugly solution to the issue where fence-create can't rely on flask app config) but new config doesn't rely on that. Should cleanup code to not do this pass-through logic.

### Bug Fixes
- load from config file for google sa validation in every instance of fence-create call
- ensure we're comparings strings to strings for Google parent orgs from config (could be interpretted as ints in the YAML if not correctly quoted)

### DIFF
--- a/fence/resources/google/access_utils.py
+++ b/fence/resources/google/access_utils.py
@@ -976,6 +976,12 @@ def remove_white_listed_service_account_ids(
     if white_listed_sa_emails is None:
         white_listed_sa_emails = config.get("WHITE_LISTED_SERVICE_ACCOUNT_EMAILS", [])
 
+    logger.debug(
+        "Removing whitelisted SAs {} from the SAs on the project.".format(
+            white_listed_sa_emails
+        )
+    )
+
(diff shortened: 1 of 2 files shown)


PlanXCyborg:

<details>

```diff
--- fence/resources/google/access_utils.py
+++ blackened
@@ -390,19 +390,14 @@
         session.query(UserGoogleAccount)
         .filter(UserGoogleAccount.email == linked_email)
         .first()
     )
     if linked_account:
-        user = (
-            session.query(User)
-            .filter(User.id == linked_account.user_id)
-            .first()
-        )
+        user = session.query(User).filter(User.id == linked_account.user_id).first()
         return user
     else:
         return None
-
 
 
 def get_user_by_email(user_email, db=None):
     """
     Return user from fence DB
```
</details></br>

PR #582, opened 2019-03-04
fix(google-sa-validation): handle edge-case where monitoring SA loses…

Avantol13:
### Bug Fixes
- fix google sa validation edge-case where monitor loses access, correctly handle and log details


PlanXCyborg:

PR #587, opened 2019-03-18
PXD-2252 Fix/google deprecated

giangbui:

Replace  google deprecated oauth2client with authlib and modify orcid and microsoft flows so that all the identity providers oauth2 flows are the same templates

### New Features
- Implemented base oauth2 class so that from which all IdP oauth2 classes can inherit


PlanXCyborg:

fence/resources/openid/google_oauth2.py
```diff
@@ -23,86 +23,92 @@ class Oauth2Client(object):
...
+                    "http": "http://"
+                    + self.HTTP_PROXY["host"]
+                    + ":"
+                    + str(self.HTTP_PROXY["port"])
+                }
+            token = self.session.fetch_token(
+                url=token_endpoint, code=code, proxies=proxies
+            )
+
+            parts = token["id_token"].split(".")
```
rudyardrichter: tokens are JWTs and google exposes their JWKS endpoint too from the discovery endpoint (`jwks_uri`). authlib has some [utilities for handling jwts](https://docs.authlib.org/en/latest/jose/jwt.html#jwt-payload-claims-validation), let's use those

fence/resources/openid/google_oauth2.py
```diff
@@ -23,86 +23,92 @@ class Oauth2Client(object):
...
-                    return {"email": user_profile.get("email")}
-                else:
-                    return {
-                        "error": (
-                            "Your email is not verified: {}".format(
-                                user_profile.get("error", "")
-                            )
-                        )
-                    }
+                proxies = {
```
rudyardrichter: more readable imo
```suggestion
                url = "http://{}:{}".format(self.HTTP_PROXY["host"], str(self.HTTP_PROXY["port])
                proxies = {"http": url}
```

fence/resources/openid/google_oauth2.py
giangbui: Updated in 7745da9f8b59680c1e417a484008c2a56a8d0f90

fence/resources/openid/google_oauth2.py
giangbui: fixed in 7745da9f8b59680c1e417a484008c2a56a8d0f90

fence/resources/openid/google_oauth2.py
```diff
@@ -23,86 +23,97 @@ class Oauth2Client(object):
...
+            )
+
+            keys = self.session.request("GET", jwks_uri).json()["keys"]
+            claims = jwt.decode(
+                token["id_token"],
+                keys,
+                options={"verify_aud": False, "verify_at_hash": False},
+            )
+
+            if claims["email"]:
```
zflamig: We should check email_verified too?

fence/resources/openid/google_oauth2.py
```diff
@@ -23,86 +23,97 @@ class Oauth2Client(object):
...
+
+            token = self.session.fetch_token(
+                url=token_endpoint, code=code, proxies=proxies
+            )
+
+            jwks_uri = self.get_discovered_endpoint(
+                "jwks_uri", "https://www.googleapis.com/oauth2/v3/certs"
+            )
+
+            keys = self.session.request("GET", jwks_uri).json()["keys"]
```
zflamig: Can we update Microsoft and ORCID to use this flow too?

fence/resources/openid/google_oauth2.py
```diff
@@ -23,86 +23,97 @@ class Oauth2Client(object):
...
+                        endpoint_key, document.json(), default_endpoint
+                    )
                 )
-                userinfo_endpoint = default_userinfo
-            elif userinfo_endpoint != default_userinfo:
+                return_value = default_endpoint
+            elif return_value != default_endpoint:
                 logger.info(
-                    "Google's userinfo endpoint {} differs from our "
+                    "ORCID's {}, {} differs from our "
```
zflamig: Google not ORCID here

fence/resources/openid/google_oauth2.py
zflamig: Also I don't think we need to send the token for this call, so we can use requests lib and include the proxies?

Avantol13:
General comment, might be cool to use a base class now that we have a ton of different IDPs... might be cool to do something similar to this for the clients : https://github.com/uc-cdis/fence/pull/593 <- perhaps you could just integrate this to your PR too? open to debate of course

fence/resources/openid/google_oauth2.py
giangbui: Add email verification for google only in cc3446048954a83a8ddcc561c3d936fea4187717. Microsoft and Orcid don't have that field in their claims

fence/resources/openid/google_oauth2.py
giangbui: Update in cc3446048954a83a8ddcc561c3d936fea4187717

fence/resources/openid/google_oauth2.py
```diff
@@ -23,86 +23,99 @@ class Oauth2Client(object):
...
         self.logger = logger
         self.settings = settings
-        self.flow = OAuth2WebServerFlow(
+        self.session = OAuth2Session(
             client_id=settings["client_id"],
             client_secret=settings["client_secret"],
             scope="openid email",
             redirect_uri=settings["redirect_url"],
         )
+        self.discovered_data = requests.get(self.GOOGLE_DISCOVERY_URL)
```
Avantol13: you should probably call this the `discovery_url`

Avantol13: oh, wait you actually GET here? I wouldn't do that unless you check that the response is OK

fence/resources/openid/google_oauth2.py
```diff
@@ -23,86 +23,99 @@ class Oauth2Client(object):
...
             else:
-                return {"error": "Can't get user's email"}
+                return {"error": "Can't get user's Google email!"}
         except Exception as e:
             self.logger.exception("Can't get user info")
-            return {"error": "Can't get your email: {}".format(e)}
+        return {"error": "Can't get your Google email: {}".format(e)}
 
-    def get_userinfo_endpoint(self):
+    def get_discovered_endpoint(self, endpoint_key, default_endpoint):
```
Avantol13: I think a better name would be `get_from_discovery_doc`

fence/resources/openid/google_oauth2.py
giangbui: Fixed in cc3446048954a83a8ddcc561c3d936fea4187717

fence/resources/openid/google_oauth2.py
Avantol13: you could just do this in the function you call to get info from the doc. seems scary to have a request come out of an __init__, especially since these objects are created and if fence is running for a while, the doc may update and we'll never GET the new one

fence/resources/openid/google_oauth2.py
```diff
@@ -23,86 +23,99 @@ class Oauth2Client(object):
...
         """
-        Return the url for Google's Userinfo endpoint by the recommended method of
-        using their discovery url. Default to current userinfo url as identified
-        09 JAN 2019.
+        Return the url for Google's endpoint by the recommended method of
```
Avantol13: should update this, it no longer returns the url for the discovery url, it tries to get a value from the discovery document given a key

fence/resources/openid/google_oauth2.py
Avantol13: `default_endpoint` is sorta confusing, isn't this a `default_value`? it's not an endpoint

fence/resources/openid/google_oauth2.py
giangbui: It is very rare situation that IdP changes their APIs without make any announcement. BTW, I updated the code in b8c8cbb710e60174a5996f6797bea6f3168af547 so that anytime the oauth 2 flow runs, it will hit the API.

fence/resources/openid/google_oauth2.py
giangbui: I rename it to `get_value_from_discovery_doc`

giangbui: Fixed in b8c8cbb710e60174a5996f6797bea6f3168af547

fence/resources/openid/google_oauth2.py
giangbui: Fixed in b8c8cbb710e60174a5996f6797bea6f3168af547

fence/resources/openid/idp_oauth2.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+
+    def get_value_from_discovery_doc(self, key, default_value):
+        """
+        Given a key return a value by the recommended method of
+        using their discovery url. Default to current url as identified
+        14 MAR 2019.
+        """
+
+        document = requests.get(self.discovery_url)
```
zflamig: lets make sure we cache this so we don't send multiple requests for the same document?

fence/resources/openid/google_oauth2.py
giangbui: Renamed `discovered_data` to `discovery_url` and check response status in b8c8cbb710e60174a5996f6797bea6f3168af547

fence/resources/openid/idp_oauth2.py
Avantol13: @giangbui something like this might be useful: https://pypi.org/project/cached-property/

maybe in the `Oauth2ClientBase`:
```
    @cached_property
    def discovery_doc(self):
        return requests.get(self.discovery_url)
```

and then the function to get value could look something like? 
```
    def get_value_from_discovery_doc(self, key, default_value):
        """
        Given a key return a value by the recommended method of
        using their discovery url.
        """
        return_value = default_value

        if self.discovery_doc.status_code == requests.codes.ok:
            return_value = self.discovery_doc.json().get(key)
            if not return_value:
                self.logger.warning(
                    "could not retrieve `{}` from {} response {}. "
                    "Defaulting to {}".format(
                        key, self.idp, self.discovery_doc.json(), default_value
                    )
                )
                return_value = default_value
            elif return_value != default_value:
                self.logger.info(
                    "{}'s {}, {} differs from our "
                    "default {}. Using {}'s...".format(
                        self.idp, key, return_value, default_value, self.idp
                    )
                )
        else:
            # invalidate the cache
            del self.__dict__["discovery_doc"]

            self.logger.error(
                "{} ERROR from {} API, could not retrieve `{}` from response {}. Defaulting to {}".format(
                    self.discovery_doc.status_code,
                    self.idp,
                    key,
                    self.discovery_doc.json(),
                    default_value,
                )
            )
            return_value = default_value

        return return_value
```

giangbui: Great! That is what I am searching now. I am wondering what is its `time to live`? Maybe we should use `cached_property_with_ttl`?

fence/resources/openid/idp_oauth2.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+
+    def get_jwt_keys(self, jwks_uri):
+        """
+        Get jwt keys from provider's api
+        Return None if there is an error while retrieving keys from the api
+        """
+        resp = requests.get(url=jwks_uri, proxies=self.get_proxies())
+        if resp.status_code != requests.codes.ok:
+            logger.error(
```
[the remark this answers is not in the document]
giangbui: Good catch! Thanks!

fence/resources/openid/idp_oauth2.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        Get jwt identity claims
+        """
+        token = self.get_token(token_endpoint, code)
+        keys = self.get_jwt_keys(jwks_endpoint)
+
+        return jwt.decode(
+            token["id_token"],
+            keys,
+            options={"verify_aud": False, "verify_at_hash": False},
```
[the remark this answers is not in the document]
giangbui: The `verify_aud ` validation is not important for us ( Thanks Rudy for pointing out)
this section from google's OIDC explains it
`Normally, it is critical that you validate an ID token before you use it, but since you are communicating directly with Google over an intermediary-free HTTPS channel and using your client secret to authenticate yourself to Google, you can be confident that the token you receive really comes from Google and is valid.`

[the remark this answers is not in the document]
giangbui: From the document
https://python-jose.readthedocs.io/en/latest/jwt/api.html

`If the “at_hash” claim is included in the claim set, then the access_token must be included, and it must match the “at_hash” claim`. Since the token is valid as explained above so we can skip providing the `access_token`

fence/resources/openid/idp_oauth2.py
```diff
@@ -0,0 +1,100 @@
...
+        self.session = OAuth2Session(
+            client_id=settings["client_id"],
+            client_secret=settings["client_secret"],
+            scope=scope,
+            redirect_uri=settings["redirect_url"],
+        )
+        self.discovery_url = discovery_url
+        self.idp = idp
+        self.HTTP_PROXY = HTTP_PROXY
+
```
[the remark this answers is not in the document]
giangbui: Added in 688048858bda53a9c885643b94db2daba8b99a9c

fence/resources/openid/google_oauth2.py
```diff
@@ -21,88 +14,47 @@ class Oauth2Client(object):
...
-            logger.error(
-                "{} ERROR from Google API, could not retrieve `userinfo_endpoint` from "
-                "Google response {}. Defaulting to {}".format(
-                    document.status_code, document, default_userinfo
-                )
-            )
-            userinfo_endpoint = default_userinfo
-
-        return userinfo_endpoint
+        return {"error": "Can't get your Google email: {}".format(e)}
```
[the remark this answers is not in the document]
giangbui: Fixed in 688048858bda53a9c885643b94db2daba8b99a9c

fence/resources/openid/microsoft_oauth2.py
```diff
@@ -19,50 +13,39 @@ class Oauth2Client(object):
...
+            settings,
+            logger,
             scope="openid email",
-            redirect_uri=settings["redirect_url"],
+            discovery_url=self.MICROSOFT_DISCOVERY_URL,
+            idp="Microsoft",
+            HTTP_PROXY=HTTP_PROXY,
         )
-        self.discovered_data = requests.get(self.MICROSOFT_DISCOVERY_URL)
         self.auth_url = self.get_auth_url()
```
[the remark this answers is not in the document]
giangbui: Fixed in 688048858bda53a9c885643b94db2daba8b99a9c

fence/resources/openid/idp_oauth2.py
```diff
@@ -0,0 +1,100 @@
...
+            token["id_token"],
+            keys,
+            options={"verify_aud": False, "verify_at_hash": False},
+        )
+
+    def get_value_from_discovery_doc(self, key, default_value):
+        """
+        Given a key return a value by the recommended method of
+        using their discovery url. Default to current url as identified
+        14 MAR 2019.
```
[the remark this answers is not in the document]
giangbui: Removed in 688048858bda53a9c885643b94db2daba8b99a9c

fence/resources/openid/orcid_oauth2.py
```diff
@@ -39,68 +34,19 @@ def get_auth_url(self):
...
+            )
+            jwks_endpoint = self.get_value_from_discovery_doc(
+                "jwks_uri", "https://orcid.org/oauth/jwks"
             )
+            claims = self.get_jwt_claims_identity(token_endpoint, jwks_endpoint, code)
 
-            if token["orcid"]:
-                return token["orcid"]
+            if claims["sub"]:
+                return claims["sub"]
```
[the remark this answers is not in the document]
giangbui: I changed the logic of `get` method in https://github.com/uc-cdis/fence/blob/688048858bda53a9c885643b94db2daba8b99a9c/fence/blueprints/login/orcid.py#L25

fence/resources/openid/idp_oauth2.py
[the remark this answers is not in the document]
Avantol13: @giangbui can you add those details in a comment for future reference?

PR #588, opened 2019-03-18
[PXP-2318] chore(app-debug): Set log level based on config['DEBUG']

vpsx:
todo: update cdislogging tag and authutils tag

tldr: Updated cdislogging. In Fence, **stop using app.logger; only use `get_logger()` from cdislogging**, both to instantiate and to update logger objects. Supply the optional `log_level` argument to `get_logger()` to set logging levels, bearing in mind the tree structure of the loggers. You probably just want to set log level once in `app_config()`, and in all other modules, simply inherit the level and handlers from this parent logger by calling `get_logger()` with no `log_level` arg (which will set level to default NOTSET).

- [PXP-2318] is to set `app.debug` in `app_config()` based on `config["DEBUG"]`.
- However, toggling `app.debug` turns out to toggle not just DEBUG level logging but also INFO and WARNING levels.
- This is because flask 0.12 does this: https://github.com/pallets/flask/blob/1c2a2cd0014693f1e1f894e498c2dc50dbb44f66/flask/logging.py#L65-L89

- In both dev and prod, we would like to log at either DEBUG or INFO level depending on `config["DEBUG"]`.
- ~So if we want this behavior we need to always have `app.debug==True` and always have `app.config['LOGGER_HANDLER_POLICY']=='always'`, and then in `app_config()`, set the `logger.level`--not `app.debug`--based on `config["DEBUG"]`. So probably the story needs to be modified.~

- So, we are not going to use the app logger at all, and instead use only cdislogging `get_logger()`. 
- `get_logger()` should be used both to instantiate and to update logger objects. Supply `log_level` argument to `get_logger()` to set logging levels, bearing in mind the tree structure of the loggers.

- `app.debug` will still indeed be set based on `config["DEBUG"]`. Not a problem if we avoid using the flask app logger.

### Improvements
Set log level based on config['DEBUG']

### DIFF
--- a/fence/__init__.py
+++ b/fence/__init__.py
@@ -37,6 +37,8 @@ import fence.blueprints.google
 
 from cdislogging import get_logger
 
+# import pdb
+# pdb.set_trace()
 logger = get_logger(__name__)
 
 app = flask.Flask(__name__)
@@ -153,6 +155,9 @@ def app_config(
     logger.error("This is an ERROR")
     logger.info("Done logging info, warn, debug, error before loading config")
     print("Before loading config....... app.debug is... " + str(app.debug))
+    print("Before loading config....... logger.level is..." + str(logger.level))
+    print("Effective level: " + str(logger.getEffectiveLevel()))
+    print("Manager disable level: " + str(logger.manager.disable))
 
     logger.info("Loading settings...")
     # not using app.config.from_object because we don't want all the extra flask cfg
@@ -162,6 +167,9 @@ def app_config(
 
     # dump the settings into the config singleton before loading a configuration file
     config.update(dict(settings_cfg))
+
+    # import pdb
+    # pdb.set_trace()
 
     # load the configuration file, this overwrites anything from settings/local_settings
     config.load(config_path, file_name)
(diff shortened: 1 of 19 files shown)


PlanXCyborg:

fence/__init__.py
```diff
@@ -173,6 +173,19 @@ def app_config(
         config["STORAGE_CREDENTIALS"], logger=app.logger
     )
 
+    # app.debug should always be True bc we want at least INFO lvl logging in dev+prod
+    # and flask 0.12 will set lvl to ERROR if app.debug is False
+    app.debug = True
+    logger.level = 10 if config["DEBUG"] == True else 20
+
+    if (
```
vpsx: Ehh, probably unnecessary?

fence/__init__.py
```diff
@@ -37,7 +37,7 @@
 import fence.blueprints.link
 import fence.blueprints.google
 
-from cdislogging import get_logger
+from cdislogging import get_logger, get_stream_handler
```
Avantol13: I don't think we use `get_stream_handler`, can remove?

PR #593, opened 2019-03-21
feat(login): base class, more DRY. Try to use more sane naming conven…

Avantol13:
### Improvements
- Base class for different IDPs for Login
- Update default configuration with details about setting up ORCID & Microsoft OAuth 2 Clients


PlanXCyborg:

fence/blueprints/login/base.py
```diff
@@ -0,0 +1,74 @@
+import flask
+from flask_restful import Resource
+
+from fence.auth import login_user
+from fence.config import config
+from fence.errors import UserError
+
+
+class DefaultOAuth2Login(Resource):
+    def __init__(self, idp_name, client, mock_username="test@example.com"):
```
rudyardrichter: It feels a little weird having the mock username as an option here, is there a case where just the config is not sufficient?

fence/blueprints/login/base.py
```diff
@@ -0,0 +1,74 @@
...
+        self.idp_name = idp_name
+        self.client = client
+        self.mock_username = mock_username
+
+    def get(self):
+        flask.redirect_url = flask.request.args.get("redirect")
+        if flask.redirect_url:
+            flask.session["redirect"] = flask.redirect_url
+
+        config_name = "MOCK_{}_AUTH".format(self.idp_name.upper())
```
rudyardrichter: using this format feels a little magical to me, what would you think about organizing it like this?
```yaml
OPENID_CONNECT:
  google:
    ...
    mock: true
    mock_username: 'testuser@gmail.com'
```
```python
if config.get(self.idp_name, {}).get("mock"):
    ...
```

fence/blueprints/login/base.py
```diff
@@ -0,0 +1,74 @@
...
+        if config.get(config_name, False):
+            username = flask.request.cookies.get(
+                config.get("DEV_LOGIN_COOKIE_NAME"), self.mock_username
+            )
+            return _login(username, self.idp_name)
+
+        return flask.redirect(self.client.get_auth_url())
+
+
+class DefaultOAuth2Callback(Resource):
```
rudyardrichter: :ok_hand:

fence/blueprints/login/base.py
Avantol13: ooh, that's definitely more elegant. for backwards compatibility I think I need to at least still support the MOCK_GOOGLE_AUTH config :thinking: but I'll try to update this and let you know

Avantol13: actually... I think I have a neat solution for the backwards compatibility thing...

PR #596, opened 2019-03-26
Feat/public urls google

Avantol13:
Public files in Google require user to be authenticated but don't restrict who can access. Current default "signing" method for public files is to *not* sign. However, that means that applications cannot effectively access the files since the method Google uses to determine authentication is browser cookies. To ensure consistent signing and programmatic access (even if data is public) changing default to forcibly sign the URL (so long as access_token is provided in request).

### New Features
- optional `no_force_sign` argument to /data/download to suggest not to sign the url if the data is public

### Improvements
- will sign public Google storage object urls by default now using provided or anonymous creds (better support programmatic access) 
- will sign public AWS S3 objects by default (with provided user creds or anonymous user creds) UNLESS the entire bucket is configured as public


PlanXCyborg:

fence/blueprints/data/indexd.py
```diff
@@ -508,15 +525,26 @@ class GoogleStorageIndexedFileLocation(IndexedFileLocation):
...
+        # if requested not to sign and it's public, don'return Google's
+        # public url to the file
+
+        try:
+            set_current_token(validate_request(aud={"user"}))
+        except JWTError:
+            # this is fine b/c it might be public data, current_token just won't be set
+            pass
+
+        if (public_data and not current_token) or (
```
philloooo: signing the url doesn't require authenticated user.
public data download needs to allow anonymous call to this endpoint and it should still return a signed url

philloooo: the behavior should be (ture for both aws and google):
- if user token is presented, the url will be signed with username logged in bucket log
- if user token is presigned, the url will be signed without username logged in bucket log.

We need to also update aws logic to set username in the signed url if token is present, right now for public url, user token is ignored. We should update the logic to be:
```
 username = get_username_from_token(default='anonymous')
```

fence/blueprints/data/indexd.py
```diff
@@ -421,8 +433,10 @@ def get_credential_to_access_bucket(self, aws_creds, expires_in):
         cred_key = get_value(
             bucket_cred, "cred", InternalError("credential of that bucket is missing")
         )
+
+        # if public, just get the first credentials to sign with
         if cred_key == "*":
```
philloooo: we don't want to change the behavior for this

PR #609, opened 2019-04-05
feat(scopes): move allowed scopes to config

Avantol13:
### Improvements
- Allowed scopes for user/client/session now in configuration instead of hard-coded

### DIFF
--- a/fence/blueprints/link.py
+++ b/fence/blueprints/link.py
@@ -40,8 +40,11 @@ def make_link_blueprint():
     blueprint = flask.Blueprint("link", __name__)
     blueprint_api = RestfulApi(blueprint)
 
-    blueprint_api.add_resource(GoogleLinkRedirect, "/google", strict_slashes=False)
-    blueprint_api.add_resource(GoogleCallback, "/google/callback", strict_slashes=False)
+    if config["ALLOW_GOOGLE_LINKING"]:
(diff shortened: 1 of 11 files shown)


PlanXCyborg:

<details>

```diff
--- fence/jwt/token.py
+++ blackened
@@ -504,6 +504,5 @@
     token = UnsignedIDToken(claims, options=token_options)
     token.validate()
 
     return token
 
-
```
</details></br>

fence/blueprints/link.py
```diff
@@ -40,8 +40,11 @@ def make_link_blueprint():
     blueprint = flask.Blueprint("link", __name__)
     blueprint_api = RestfulApi(blueprint)
 
-    blueprint_api.add_resource(GoogleLinkRedirect, "/google", strict_slashes=False)
-    blueprint_api.add_resource(GoogleCallback, "/google/callback", strict_slashes=False)
+    if config["ALLOW_GOOGLE_LINKING"]:
```
Avantol13: this conditional resource addition is b/c we don't have a scope for Google Linking (there's a story to do this.. but these endpoints are open by default). So added config switch to just completely kill them off

Avantol13: They would allow someone to get their personal Google account to access data in Google, which if we want to house mult projects per bucket, we cannot safely do this. So with this disabled and some other configurations to restrict scopes on the tokens, we can effectively only allow signed URLs in google

PR #611, opened 2019-04-09
feat:S3 endpoint

bwalsh:
Enables ceph support via ceph's s3 gateway.  Adds support for s3 `endpoint`, which allows fence to connect to s3 instances other than AWS.

### New Features
- Implemented `endpoint`:
  * Reads endpoint property from fence-config.yaml
  * If present, uses fence/resources/aws/boto_manager:: presigned_url  

### Breaking Changes
* None

### Bug Fixes
* None

### Improvements
Enables ceph support via ceph's s3 gateway

### Dependency updates
* None

### Deployment changes
* None


fence/jwt/token.py
```diff
@@ -429,6 +429,22 @@ def generate_signed_access_token(
             "issuing JWT access token\n" + json.dumps(claims, indent=4)
         )
 
+    # provide comma separated list of roles and user_name
+    # as simplified, shortened (non-nested) properties
+    simplify = lambda p : p.replace('programs.','') \
```
Avantol13: We are still working on a better way to represent the policies in tokens (this feature is a WIP), but we wanted to keep the fully policy name here so unfortunately shortening it like this will not work. in other words, we should be able to map 1:1 policies in this field to our AuthZ policy names, which, at the moment (as you noticed) are a bit verbose. This hopefully will not be the case down the road, but for now, I don't think we can support this shortening

fence/blueprints/data/indexd.py
```diff
@@ -485,6 +485,10 @@ def get_signed_url(
         )
 
         credential = self.get_credential_to_access_bucket(aws_creds, expires_in)
+        if 'endpoint_url' in credential:
```
Avantol13: if you're adding a new field (even if it's optional) to the configuration, it would be awesome to add a quick comment in the `config-default.yaml` describing the optional configuration

Dockerfile
```diff
@@ -1,6 +1,3 @@
-# To run: docker run -d -v /path/to/fence-config.yaml:/var/www/fence/fence-config.yaml --name=fence -p 80:80 fence
```
Avantol13: can you add these comments back in? actually, I think this branch is out of date with master. we've updated the dockerfile to use nginx and different base image (not full ubuntu). If you could merge master into your branch / rebase, that would be awesome!

fence/blueprints/data/indexd.py
```diff
@@ -507,15 +511,25 @@ def get_signed_url(
...
+                http_url,
+                ACTION_DICT["s3"][action],
+                credential,
+                "s3",
+                region,
+                expires_in,
+                user_info,
+            )
+        else:
+            url = flask.current_app.boto.presigned_url(
```
Avantol13: I have a couple other comments that should be addressed first, and once they are, we'd like to do some testing with this feature on our end

fence/jwt/token.py
bwalsh: I've removed the abbreviations, but kept the simplification.  
Integration of [downstream apps](https://opendistro.github.io/for-elasticsearch-docs/docs/security-configuration/configuration/#configure-json-web-tokens) require a simple, non-nested key to roles and subject at the root level.

fence/blueprints/data/indexd.py
bwalsh: comment added [here](https://github.com/ohsu-comp-bio/fence/blob/s3-endpoint/tests/test-fence-config.yaml#L299)

Dockerfile
bwalsh: done

fence/blueprints/data/indexd.py
bwalsh: I've repurposed one of your tests [here](https://github.com/ohsu-comp-bio/fence/blob/s3-endpoint/tests/data/test_non_aws_s3.py) but is fails to exercise the [new code](https://github.com/ohsu-comp-bio/fence/blob/s3-endpoint/fence/blueprints/data/indexd.py#L475) I suspect I don't fully understand the test [configuration](https://github.com/ohsu-comp-bio/fence/blob/s3-endpoint/tests/conftest.py#L470).

bwalsh:
deprecating, will re-open a separate PR

PR #613, opened 2019-04-09
feat/arborist-sync

rudyardrichter:
### Improvements

- Update user sync for compatibility with changes to arborist.
  - Policies now owned by arborist
  - Arborist needs (read-only) copy of users.

### Breaking Changes

- Intended to work only with arborist >= 2.0 (using new database). Won't work with older arborist API.


PlanXCyborg:

<details>

```diff
--- fence/__init__.py
+++ blackened
@@ -14,11 +14,13 @@
 from fence.oidc.client import query_client
 from fence.oidc.server import server
 from fence.rbac.client import ArboristClient
 from fence.resources.aws.boto_manager import BotoManager
 from fence.resources.openid.google_oauth2 import GoogleOauth2Client as GoogleClient
-from fence.resources.openid.microsoft_oauth2 import MicrosoftOauth2Client as MicrosoftClient
+from fence.resources.openid.microsoft_oauth2 import (
+    MicrosoftOauth2Client as MicrosoftClient,
+)
 from fence.resources.openid.orcid_oauth2 import OrcidOauth2Client as ORCIDClient
 from fence.resources.storage import StorageManager
 from fence.resources.user.user_session import UserSessionInterface
 from fence.error_handler import get_error_response
 from fence.utils import random_str

--- fence/blueprints/data/indexd.py
+++ blackened
@@ -274,19 +274,14 @@
 
     def check_rbac(self, action):
         if "rbac" not in self.index_document:
             raise ValueError("index record missing `rbac`")
         request = {
-            "user": {
-                "jwt": get_jwt(),
-            },
+            "user": {"jwt": get_jwt()},
             "request": {
                 "resource": self.index_document["rbac"],
-                "action": {
-                    "service": "fence",
-                    "method": action,
-                },
+                "action": {"service": "fence", "method": action},
             },
         }
         return flask.current_app.arborist.auth_request(request)
 
     @cached_property

--- fence/rbac/client.py
+++ blackened
@@ -246,13 +246,11 @@
             msg = data["error"]
             if isinstance(data["error"], dict):
                 msg = data["error"].get("message", msg)
             resource = resource_json.get("path", "/" + resource_json.get("name"))
             self.logger.error(
-                "could not create resource `{}` in arborist: {}".format(
-                    resource, msg
-                )
+                "could not create resource `{}` in arborist: {}".format(resource, msg)
             )
             raise ArboristError(data["error"])
         self.logger.info("created resource {}".format(resource_json["name"]))
         return data
 
@@ -415,13 +413,11 @@
         response = requests.delete(url)
         data = _request_get_json(response)
         if response.status_code != 204:
             msg = data.get("error", "unhelpful response from arborist")
             self.logger.error(
-                "could not revoke policies from user `{}`: {}`".format(
-                    username, msg
-                )
+                "could not revoke policies from user `{}`: {}`".format(username, msg)
             )
             return None
         self.logger.info("revoked all policies from user `{}`".format(username))
         return data
 

--- fence/sync/sync_users.py
+++ blackened
@@ -1012,11 +1012,13 @@
         if not self.arborist_client:
             self.logger.warn("no arborist client set; skipping arborist sync")
             return False
         if not self.arborist_client.healthy():
             # TODO (rudyardrichter, 2019-01-07): add backoff/retry here
-            self.logger.error("arborist service is unavailable; skipping main arborist sync")
+            self.logger.error(
+                "arborist service is unavailable; skipping main arborist sync"
+            )
             return False
 
         # Set up the resource tree in arborist
         resources = user_yaml.rbac.get("resources", [])
         for resource in resources:
```
</details></br>

fence/blueprints/data/indexd.py
```diff
@@ -270,6 +272,23 @@ def set_acls(self):
...
+            raise ValueError("index record missing `rbac`")
+        request = {
+            "user": {
+                "jwt": get_jwt(),
+            },
+            "request": {
+                "resource": self.index_document["rbac"],
+                "action": {
+                    "service": "fence",
+                    "method": action,
```
Avantol13: seems like this is either `upload` or `download`, I thought we were going to stick with CRUD and `read_storage`, `write_storage` as the arborist actions? it would nice to have the arborist library do the creation of this json too, like just pass in the jwt, resource, and action into a function

fence/jwt/token.py
```diff
@@ -481,7 +479,6 @@ def generate_id_token(
 
     # If not provided, assume auth time is time this ID token is issued
     auth_time = auth_time or iat
-    policies = [policy.id for policy in user.policies]
```
Avantol13: I think we want to keep these in the id token, right? just remove from access token?

fence/sync/sync_users.py
```diff
@@ -997,9 +983,14 @@ def _sync(self, sess):
         else:
             self.logger.info("No resources specified; skipping arborist sync")
 
-    @staticmethod
-    def _reset_user_access(session):
-        session.execute(users_to_policies.delete())
+    def _reset_user_access(self, session):
+        all_users = session.query(User).all()
+
+        if self.arborist_client:
```
Avantol13: could have this check before this function is event called to avoid the db query for users if no arborist

fence/sync/sync_users.py
```diff
@@ -997,9 +983,14 @@ def _sync(self, sess):
...
+            usernames = {user.username for user in all_users}
+            for username in usernames:
+                self.arborist_client.revoke_all_policies_for_user(username)
```
Avantol13: wait, does this completely wipe everything in arborist and then later try to recreate everything for every usersync? can we just PUT the new stuff and it'll overwrite ?

rudyardrichter: Yes, it wipes out the stuff in arborist to overwrite it. The idea is that if syncing from the user.yaml, exactly what's in the user.yaml should be what ends up in arborist.

Avantol13: I guess this would have to be more involved. It would have to diff b/t current arborist status like the dbgap syncing is doing. And then determine what is the same, what needs to be updated, and what needs to be removed

fence/blueprints/data/indexd.py
rudyardrichter: The `feat/download-auth` branch has the "real" data blueprint changes, which are using the old actions.

fence/sync/sync_users.py
```diff
@@ -783,6 +767,8 @@ def _upsert_userinfo(self, sess, user_info):
                 u = User(username=username)
                 sess.add(u)
 
+            self.arborist_client.create_user({"name": username})
```
Avantol13: should check `if self.arborist_client`

fence/sync/sync_users.py
```diff
@@ -987,7 +973,7 @@ def _sync(self, sess):
             self.logger.info("No users for syncing")
 
         if user_yaml.rbac:
-            self.logger.info("Synchronizing arborist")
+            self.logger.info("Synchronizing arborist...")
             success = self._update_arborist(sess, user_yaml)
```
Avantol13: I think you need to check `if self.arborist_client` before this as well

fence/sync/sync_users.py
Avantol13: yeah, this is just going to be really expensive for dbgap syncing

Avantol13: ideally we'd do something like the current sync process where there's a diff stage b/t the request and current state, and then we only make the necessary db/api calls for changes

fence/blueprints/data/indexd.py
Avantol13: okay, so to be clear, we will be using `read_storage` and `write_storage` in the final version?

fence/sync/sync_users.py
Avantol13: b/c there are a *lot* of users and projects in dbgap.. and if we have to delete and then recreate all those policies and users every time, usersync is going to take forever and be a real strain with all those API calls. at least, I assume it will

Avantol13: moar thoughts: this also means that for a brief moment, someone might not have access when they should. And depending on how long the sync takes, that could cause some annoyance to end-users with random auth rejections. Probably a small window, but still possible

fence/rbac/client.py
```diff
@@ -357,3 +369,82 @@ def create_policy(self, policy_json, skip_if_exists=True):
...
+        data = {
+            "name": name,
+            "users": users,
+            "policies": policies,
+        }
+        if description:
+            data["description"] = description
+        response = requests.post(self._group_url, json=data)
+        data = _request_get_json(response)
+        if response.status_codee != 201:
```
Avantol13:
```suggestion
        if response.status_code != 201:
```

fence/rbac/client.py
```diff
@@ -357,3 +369,82 @@ def create_policy(self, policy_json, skip_if_exists=True):
...
+        policies = policies or []
+        data = {
+            "name": name,
+            "users": users,
+            "policies": policies,
+        }
+        if description:
+            data["description"] = description
+        response = requests.post(self._group_url, json=data)
+        data = _request_get_json(response)
```
Avantol13:
```suggestion
        data = _request_get_json(response)
        if response.status_code == 409:
            # already exists
            return
```

fence/blueprints/data/indexd.py
rudyardrichter: yep!

Avantol13:
was looking through code, seems like `_parse_yaml` from `UserSyncer` isn't called from anywhere, can we remove?

fence/sync/sync_users.py
rudyardrichter: for posterity: the requests have been ordered so that an individual user's updates happen in sequence, so the moment that a user is missing their expected access will be negligible.

fence/jwt/token.py
rudyardrichter: I think we want to remove all the policies from everything; they can be retrieved from the user info endpoint now

PR #637, opened 2019-05-31
fix(usersync): allow not passing in user.yaml, better logging and sft…

Avantol13:
`fence-create sync --sync_from_dbgap True` raises exception if `sync_from_local_yaml_file` is None. it is None by default. so if you're trying to just do a dbgap sync without providing a local user yaml, it will die. apparently commons get around this by providing an empty user.yaml

Additionally, if you provide just a numeric string in sftp config (for example `password` as `555`), it will get coerced to an int before it gets to the library call (which expects a string). fixed by casting right before the call to what the library expects

### Bug Fixes
- can now successfully call `fence-create sync --sync_from_dbgap True` without providing a valid `user.yaml` (e.g. just a dbgap sync)

### Improvements
- Better logging for sftp connections for dbgap syncing

### DIFF
--- a/fence/blueprints/login/__init__.py
+++ b/fence/blueprints/login/__init__.py
@@ -22,6 +22,15 @@ from fence.config import config
 
 logger = get_logger(__name__)
 
+# Mapping from IDP ID to the name in the URL on the blueprint (see below).
+IDP_URL_MAP = {
+    "fence": "fence",
+    "google": "google",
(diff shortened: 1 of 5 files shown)


PlanXCyborg:

rudyardrichter:
one note, may want to add a log pointing out that the sync didn't load a file in that case

PR #649, opened 2019-06-25
feat(privacy-policy): add new endpoint

rudyardrichter:
### New Features

- Add `/privacy-policy` endpoint, which serves the text of the new privacy policy in either markdown or HTML depending on the content-type headers in the request.


PlanXCyborg:

<details>

```diff
--- fence/__init__.py
+++ blackened
@@ -16,11 +16,11 @@
 from fence.oidc.server import server
 from fence.rbac.client import ArboristClient
 from fence.resources.aws.boto_manager import BotoManager
 from fence.resources.openid.google_oauth2 import GoogleOauth2Client as GoogleClient
 from fence.resources.openid.microsoft_oauth2 import (
-    MicrosoftOauth2Client as MicrosoftClient
+    MicrosoftOauth2Client as MicrosoftClient,
 )
 from fence.resources.openid.orcid_oauth2 import OrcidOauth2Client as ORCIDClient
 from fence.resources.storage import StorageManager
 from fence.resources.user.user_session import UserSessionInterface
 from fence.error_handler import get_error_response
@@ -105,12 +105,11 @@
 
     google_blueprint = fence.blueprints.google.make_google_blueprint()
     app.register_blueprint(google_blueprint, url_prefix="/google")
 
     app.register_blueprint(
-        fence.blueprints.privacy.blueprint,
-        url_prefix="/privacy-policy",
+        fence.blueprints.privacy.blueprint, url_prefix="/privacy-policy"
     )
 
     fence.blueprints.misc.register_misc(app)
 
     @app.route("/")

--- setup.py
+++ blackened
@@ -41,11 +41,9 @@
         "Werkzeug>=0.12.2,<1.0.0",
         "storageclient",
         "pyyaml~=5.1",
     ],
     scripts=["bin/fence-create"],
-    package_data={
-        "fence": ["static/privacy_policy.md"],
-    },
+    package_data={"fence": ["static/privacy_policy.md"]},
     packages=find_packages(),
 )
 
```
</details></br>

fence/blueprints/privacy/blueprint.py
```diff
@@ -0,0 +1,18 @@
+import pkgutil
+
+import flask
+from markdown import Markdown
+
+blueprint = flask.Blueprint("privacy-policy", __name__)
+
+
+PRIVACY_POLICY_MD = pkgutil.get_data("fence", "static/privacy_policy.md")
+PRIVACY_POLICY_HTML = Markdown().convert(PRIVACY_POLICY_MD)
```
Avantol13: this should probably have some kind of wrapper to catch errors and log them? also, can we just do this conversion in the endpoint function instead of top level? we could still have a global and just lazy load it / cache it? I guess it's probably fine, happens once on startup :man_shrugging:

fence/blueprints/privacy/blueprint.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@blueprint.route("/", methods=["GET"])
+def privacy_policy():
+    if str(flask.request.accept_mimetypes) == "text/markdown":
```
Avantol13: this will only catch exact case with no additional arguments. which is maybe okay? but things that are technically valid won't result in markdown. like `text/markdown;charset=UTF-8`. I was digging through werkzeug, it looks like to determine the type they do a `"text/blah" in accept_mimetypes` like: https://github.com/pallets/werkzeug/blob/d824659abe95ed31b1f9c355f88c4741da5a6e5f/src/werkzeug/datastructures.py#L1850 I think we can match that pattern and perhaps even `.lower()` the input so `text/Markdown` works too? that last part isn't strictly necessary. I think it's safe for us to just ignore any arguments passed the mimetype, https://stackoverflow.com/a/7367532, it's not forbidden but doesn't seem to actually officially do anything. `Accept-Charset` would be used for it.

rudyardrichter: Good point, changed in fce697f

fence/blueprints/privacy/blueprint.py
```diff
@@ -0,0 +1,45 @@
+"""
+This blueprint defines the /privacy-policy enpdoint to serve a privacy policy from
+fence, which will be linked to in the OAuth2 consent screens.
+
+The logic as to where to get this thing from is as follows.
+
+- If there is a PRIVACY_POLICY_URL environment variable (which should come from the
+  manifest, exposed here through the fence deploy file---see manifest and
+  `cloud-automation/kube/services/fence/`), then redirect to that URL.
+- If the file `fence/static/privacy_policy.md` exists, then return that either as
```
Avantol13: should this path be configurable? it seems more robust to just dump this into a cfg var in the fence-cfg and default to this

fence/config-default.yaml
```diff
@@ -317,6 +317,10 @@ MAX_ACCESS_TOKEN_TTL: 3600
 #                               OPTIONAL CONFIGURATIONS                                #
 ########################################################################################
 
+# For displaying a privacy policy to users, we can either link to the URL specified by
+# PRIVACY_POLICY, or default to the `static/privacy_policy.md` file in fence.
+PRIVACY_POLICY: null
```
Avantol13: I don't think this is used anywhere anymore?

fence/blueprints/privacy/blueprint.py
```diff
@@ -0,0 +1,45 @@
...
+import flask
+from markdown import Markdown
+
+from fence import config
+from fence.errors import NotFound
+
+blueprint = flask.Blueprint("privacy-policy", __name__)
+
+
+PRIVACY_POLICY_URL = os.environ.get("PRIVACY_POLICY_URL")
```
Avantol13: I sorta feel that this should exist in the fence-config file as the sole location of configuration. You can set the default to `None` and do some post processing to actually default to getting from the env here: https://github.com/uc-cdis/fence/blob/cabc362dbf32b1450cd72cc5e96a89252e76e984/fence/config.py#L15 . just make sure to note that in the configuration comments. I just think simplifying and unifying all the config in one file is a good thing to try to enforce, so we don't have all these additional env vars people need to know about setting if they're deploying fence themselves

fence/static/privacy_policy.md
```diff
@@ -0,0 +1,290 @@
+# Gen3/DCFS Privacy Policy
```
Avantol13: if this file is going to live in the manifest and get mounted into the fence pod, I think it's safe to not include it in fence repo at all?

fence/blueprints/privacy/blueprint.py
rudyardrichter: actually should not be configurable because the path has to be included in `setup.py` if it's bundled with the rest of the application. neither is really relevant to our normal use case though

fence/config-default.yaml
rudyardrichter: yeah, this should have been in the blueprint. fixed

fence/blueprints/privacy/blueprint.py
rudyardrichter: yeah this is what the PRIVACY_POLICY in the config was supposed to be for—I updated this

fence/static/privacy_policy.md
rudyardrichter: :ok_hand:

rudyardrichter: to support the case for other people to bundle it with the application, and have it included in `setup.py` (assuming we want to support that) this file has to at least exist; it could be empty?

fence/config-default.yaml
```diff
@@ -317,6 +317,10 @@ MAX_ACCESS_TOKEN_TTL: 3600
 #                               OPTIONAL CONFIGURATIONS                                #
 ########################################################################################
 
+# For displaying a privacy policy to users, we can either link to the URL specified by
+# PRIVACY_POLICY, or default to the `static/privacy_policy.md` file in fence.
```
Avantol13:
```suggestion
# PRIVACY_POLICY_URL, or default to the `static/privacy_policy.md` file in fence.
```

fence/static/privacy_policy.md
Avantol13: yeeah, if we have to have it I'd say we leave it empty and leave it up to people who deploy to overwrite with actual privacy policy if they need it

PR #651, opened 2019-06-25
fix(sa-registration): PATCH that restricts existing SA access should …

Avantol13:
### Bug Fixes
- fix issue with service account registration where validation checks new access *and* previous access when trying to update an SA (so if updating to restrict to a subset of previous access, validation may fail and not allow the update)

### Improvements
- service account validation will only verify that users on the Google Project have access to the projects the *active* registered service account have access to (in other words, expired service account project access won't be checked since they don't actually have access to data)


PlanXCyborg:

PR #653, opened 2019-06-27
feat(public-files): check authz/acl field, then hit arborist as anon …

Avantol13:
### Improvements
- fence will now consider authz field on indexd record to determine if a file is public or not (for different signed url behavior). previously only checked acl field

### DIFF
--- a/fence/blueprints/data/indexd.py
+++ b/fence/blueprints/data/indexd.py
@@ -44,7 +44,6 @@ ACTION_DICT = {
     "gs": {"upload": "PUT", "download": "GET"},
 }
 
-ACTION_TO_ARBORIST_METHOD = {"upload": "write-storage", "download": "read-storage"}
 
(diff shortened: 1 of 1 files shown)


PlanXCyborg:

fence/blueprints/data/indexd.py
```diff
@@ -365,7 +367,7 @@ def metadata(self):
 
     @cached_property
     def public(self):
-        return check_public(self.set_acls)
+        return check_public(self.authz)
```
rudyardrichter: a couple thoughts:
- `check_public` only used here so this would be a good opportunity to inline
- should keep the ACLs check here too I think—we can use `authz` if exists or default to acl

Avantol13: oh, this is perhaps misleading, the cached property isn't just authz, it'll get acl field if authz doesn't exist. I will inline check_public tho

PR #656, opened 2019-07-05
Fix/dbgap combination

Avantol13:
### Bug Fixes
- fix exception when dbgap resources and user.yaml have a matching root level resource without any subresources and create regression test


PlanXCyborg:

PR #659, opened 2019-07-12
fix/privacy-policy: update for python3

rudyardrichter:
### Bug Fixes

- Fix loading privacy policy file in python3
- Fix caching

### DIFF
--- a/fence/blueprints/privacy.py
+++ b/fence/blueprints/privacy.py
@@ -18,6 +18,7 @@ import os
 
 import flask
 from markdown import Markdown
+from werkzeug.contrib.cache import SimpleCache
 
 from fence import config
 from fence.errors import NotFound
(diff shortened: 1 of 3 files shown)


fence/blueprints/privacy.py
```diff
@@ -18,33 +18,45 @@
...
+        return response
     # Check if we want to redirect out for the privacy policy.
+    PRIVACY_POLICY_URL = (
+        config.get("PRIVACY_POLICY_URL")
+        or os.environ.get("PRIVACY_POLICY_URL")
+    )
     if PRIVACY_POLICY_URL:
-        return flask.redirect(PRIVACY_POLICY_URL)
+        response = flask.redirect(PRIVACY_POLICY_URL)
+        cache.add("privacy-policy", response, timeout=0)
```
Avantol13: the forever cache is probably fine, but I would feel more comfortable if this was set to something, even if it's a long time like 30 days. in case fence is running and the data in the URL we point to changes. In practice, since we roll/update fence pretty often it's probably fine but this would make me feel a little better

fence/blueprints/privacy.py
```diff
@@ -18,33 +18,45 @@
...
 
 
 @blueprint.route("/", methods=["GET"])
 def privacy_policy():
+    global cache
+    response = cache.get("privacy-policy")
+    if response:
+        return response
     # Check if we want to redirect out for the privacy policy.
+    PRIVACY_POLICY_URL = (
```
Avantol13: why did this get moved inside the function? these configurations should never change, right? seems like they should stay global on module load.

fence/blueprints/privacy.py
```diff
@@ -18,33 +18,45 @@
...
 
-
-PRIVACY_POLICY_URL = config.get("PRIVACY_POLICY_URL") or os.environ.get(
-    "PRIVACY_POLICY_URL"
-)
-PRIVACY_POLICY_MD = pkgutil.get_data("fence", "static/privacy_policy.md")
-PRIVACY_POLICY_HTML = None
-if PRIVACY_POLICY_MD:
-    PRIVACY_POLICY_HTML = Markdown().convert(PRIVACY_POLICY_MD)
+cache = SimpleCache()
```
Avantol13: the naming convention for mutable globals trips me up. I always want it to be CAPS_CASE but I think that's only for constants

fence/blueprints/privacy.py
rudyardrichter: running it on module init caused some weirdness because of `app.config` not existing, hence the new thing with `cache`

fence/blueprints/privacy.py
rudyardrichter: it's actually only caching the redirect, not the entire response from the target URL. so if the destination is updated then it will still work

fence/blueprints/privacy.py
```diff
@@ -18,33 +18,45 @@
...
 
 
 @blueprint.route("/", methods=["GET"])
 def privacy_policy():
+    global cache
+    response = cache.get("privacy-policy")
```
paulineribeyre: fix to differentiate md vs html in cache

fence/blueprints/privacy.py
Avantol13: ohh, interesting. didn't consider that :+1:

fence/blueprints/privacy.py
Avantol13: oh, right :man_facepalming: . this is cool then :+1:

fence/blueprints/privacy.py
```diff
@@ -31,32 +31,29 @@
...
+    if not cache.has("privacy-policy-md"):
+        file_contents = pkgutil.get_data("fence", "static/privacy_policy.md").decode("utf-8")
+        if not file_contents:
+            raise NotFound("this endpoint is not configured")
+        cache.add("privacy-policy-md", file_contents)
+
     if "text/markdown" in str(flask.request.accept_mimetypes).lower():
-        if not PRIVACY_POLICY_MD:
+        if not cache.has("privacy-policy-md"):
             raise NotFound("this endpoint is not configured")
```
paulineribeyre: i think this `if` is unnecessary now that the previous block checks the cache

tests/privacy-policy/test_privacy_policy.py
```diff
@@ -1,16 +1,17 @@
...
-    fence.blueprints.privacy.cache.add("privacy-policy", privacy_policy_md)
+    fence.blueprints.privacy.cache.add("privacy-policy-md", privacy_policy_md)
     response = client.get("/privacy-policy", headers={"Accept": "text/markdown"})
     assert response.status_code == 200
     assert response.data.decode("utf-8") == privacy_policy_md
     fence.blueprints.privacy.cache.clear()
 
 
 def test_html(client, monkeypatch, privacy_policy_html):
-    fence.blueprints.privacy.cache.add("privacy-policy", privacy_policy_html)
```
paulineribeyre: why do you remove this line?

fence/blueprints/privacy.py
rudyardrichter: true, not necessary. overly defensive programming :stuck_out_tongue:

tests/privacy-policy/test_privacy_policy.py
rudyardrichter: the conftest automatically mocks `get_data` instead

rudyardrichter: so when it goes to load the markdown, it'll get the `privacy_policy_md` fixture

PR #661, opened 2019-07-19
chore(flask-vulnerability): bump flask to 1.1.1

### LINKED ISSUE #689: Upgrade Flask due to DOS
https://snyk.io/vuln/SNYK-PYTHON-FLASK-451637
https://nvd.nist.gov/vuln/detail/CVE-2019-1010083

There is a known exploit in the wild and it's fairly easy to DOS Fence with it.

paulineribeyre:
Fixed #689 

### Dependency updates
- Flask 1.1.1, Werkzeug 0.15.6


tests/data/test_data.py
```diff
@@ -137,8 +137,7 @@ def test_indexd_download_file_no_jwt(client, auth_client):
     assert response.status_code == 401
 
     # response should not be JSON, should be HTML error page
-    with pytest.raises(ValueError):
```
vpsx: You are right that in the commit you sent, the versions don't match up... I don't think it's that one, but anyway the point here was not really to check for the ValueError. To better reflect the stated point of the test, could we just check the mimetype or something instead of (or in addition to) this negative assert?

PR #670, opened 2019-07-29
[PXP-3384] fix(patch-sa): add all granting_project_ids

vpsx:
The `google-delete-expired-service-account` job does not appear to clean out the entries in fence db's `service_account_access_privilege`, so the Google SA will get removed from the GBAG when the link expires, but in `patch_user_service_account()`, the set difference will be empty. Therefore the PATCH call will not re-add the SA to the GBAG. 

So use `granting_project_ids` instead of `to_add` when re-adding SAs to GBAGs. (But continue to use `to_add` in the fence db.) Update tests accordingly. 

### Bug Fixes
Re-add SAs to all GBAGs in PATCH endpoint

### DIFF
--- a/tests/google/test_access_utils.py
+++ b/tests/google/test_access_utils.py
@@ -298,7 +298,7 @@ def test_update_user_service_account_success(cloud_manager, db_session, setup_da
     """
     (
         cloud_manager.return_value.__enter__.return_value.add_member_to_group.return_value
-    ) = {}
+    ) = {"email": "test@gmail.com"}
     (
         cloud_manager.return_value.__enter__.return_value.remove_member_from_group.return_value
     ) = {}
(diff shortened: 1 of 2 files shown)


tests/google/test_access_utils.py
```diff
@@ -413,9 +413,10 @@ def test_update_user_service_account_success3(cloud_manager, db_session, setup_d
     )
     patch_user_service_account("test", "test@gmail.com", ["test_auth_1", "test_auth_2"])
 
-    assert not (
-        cloud_manager.return_value.__enter__.return_value.add_member_to_group.called
-    )
+    # No longer true as we are now adding to every project/GBAG every time
```
Avantol13: you can just remove this code block then

PR #677, opened 2019-08-14
PXP-3717 Fix/token

vpsx:
PXP-3717
take this hack for STAGE https://github.com/uc-cdis/fence/compare/fix/3.2.1.x#diff-7b2574a05fa5500b0a4b10b7b7f23fa0
but make number of projects configurable with default 15 and add log

### Bug Fixes
limit projects in token to avoid token being too large

### DIFF
--- a/fence/jwt/token.py
+++ b/fence/jwt/token.py
@@ -373,6 +373,21 @@ def generate_signed_access_token(
         "azp": client_id or "",
     }
 
+    # NOTE: "THIS IS A TERRIBLE STOP-GAP SOLUTION SO THAT USERS WITH
+    #       MINIMAL ACCESS CAN STILL USE LATEST VERSION OF FENCE
+    #       WITH VERSIONS OF PEREGRINE/SHEEPDOG THAT DO NOT CURENTLY
+    #       SUPPORT AUTHORIZATION CHECKS AGAINST ARBORIST (AND INSTEAD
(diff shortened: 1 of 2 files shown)


fence/jwt/token.py
```diff
@@ -373,6 +373,21 @@ def generate_signed_access_token(
...
+    # NOTE: "THIS IS A TERRIBLE STOP-GAP SOLUTION SO THAT USERS WITH
+    #       MINIMAL ACCESS CAN STILL USE LATEST VERSION OF FENCE
+    #       WITH VERSIONS OF PEREGRINE/SHEEPDOG THAT DO NOT CURENTLY
+    #       SUPPORT AUTHORIZATION CHECKS AGAINST ARBORIST (AND INSTEAD
+    #       RELY ON THE PROJECTS IN THE TOKEN). If the token is too large
+    #       everything breaks. I'm sorry" --See PXP-3717
+    if len(dict(user.project_access)) < config["TOKEN_PROJECTS_CUTOFF"]:
+        claims["context"]["user"]["projects"] = dict(user.project_access)
+    else:
+        logger.warn(
```
Avantol13: `Note There is an obsolete method warn which is functionally identical to warning. As warn is deprecated, please do not use it - use warning instead.`

Avantol13: https://docs.python.org/3/library/logging.html#logging.Logger.warning

PR #682, opened 2019-08-21
fix(upload): allow upload on existing record

rudyardrichter:
### Improvements

- Change data API to allow making a new S3 URL for an index record which already exists. (Mostly for internal/testing purposes.)

### DIFF
--- a/fence/blueprints/data/indexd.py
+++ b/fence/blueprints/data/indexd.py
@@ -335,6 +335,14 @@ class IndexedFile(object):
             except IndexError:
                 raise NotFound("Can't find any file locations.")
 
+        if action == "upload" and not self.index_document.get("urls"):
+            if protocol == "s3":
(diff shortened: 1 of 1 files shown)


fence/blueprints/data/indexd.py
```diff
@@ -335,6 +335,14 @@ def _get_signed_url(
             except IndexError:
                 raise NotFound("Can't find any file locations.")
 
+        if action == "upload" and not self.index_document.get("urls"):
+            if protocol == "s3":
+                location = S3IndexedFileLocation(self.file_id)
```
paulineribeyre: does this work? shouldn't it be something like `S3IndexedFileLocation("s3://{}/{}/{}".format(bucket, self.file_id, file_name))`?

rudyardrichter: you're correct—that should be the whole path

PR #683, opened 2019-08-21
replace fence ArboristClient with gen3authz 

vpsx:
Fence implements its own ArboristClient; rip it out and use gen3authz instead. 

sister PR in gen3authz: https://github.com/uc-cdis/gen3authz/pull/5

Requires Arborist PR 100 https://github.com/uc-cdis/arborist/pull/100 which changes Arborist's `PUT /policy` route to `PUT /policy/{policyID}`. This Fence uses the new gen3authz ArboristClient, which changes the route per above. 

### Improvements
removes duplicated ArboristClient code and instead imports gen3authz's ArboristClient


fence/sync/sync_users.py
```diff
@@ -30,9 +30,10 @@
     Client,
 )
 from fence.sync import utils
-from fence.rbac.client import ArboristClient, ArboristError
 from fence.resources.storage import StorageManager
 
+from gen3authz.client.arborist.client import ArboristError
```
fantix: (trivial) personal taste: sort imports into 3 sections: system import, library import, local (relative) import. Sort alphabetically within each section, but "import xxx" goes before "from xxx import xxx".

tests/rbac/test_client.py
```diff
@@ -6,47 +6,44 @@
 from unittest import mock
```
fantix: Should this file go into gen3auth package?

requirements.txt
```diff
@@ -15,6 +15,7 @@ Flask-CORS==3.0.3
 Flask_OAuthlib==0.9.4
 flask-restful==0.3.6
 Flask_SQLAlchemy_Session==1.1
+gen3authz==0.2.0
```
fantix: maybe a newer version here with the fixes

PR #684, opened 2019-08-22
Feat/id token projects

Avantol13:
### Breaking Changes
- projects field (user access) removed from id_token for implicit client flow

### DIFF
--- a/fence/jwt/token.py
+++ b/fence/jwt/token.py
@@ -190,6 +190,7 @@ def generate_signed_id_token(
     auth_time=None,
     max_age=None,
     nonce=None,
+    projects=None,
     **kwargs
 ):
     """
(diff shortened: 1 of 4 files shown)


fence/jwt/token.py
```diff
@@ -190,6 +190,7 @@ def generate_signed_id_token(
     auth_time=None,
     max_age=None,
     nonce=None,
+    projects=None,
```
paulineribeyre:
```suggestion
    include_project_access=True,
```

fence/jwt/token.py
```diff
@@ -218,6 +219,7 @@ def generate_signed_id_token(
         user,
         expires_in,
         client_id,
+        projects=projects,
```
paulineribeyre:
```suggestion
        include_project_access=include_project_access,
```

PR #686, opened 2019-08-23
Test/improve-coverage

callenmacphee:
Add unit tests to increase coverage.

### Improvements
Added unit tests to cover multipart complete upload, fence_create, and openid connect.

### DIFF
--- a/tests/scripting/test_fence-create.py
+++ b/tests/scripting/test_fence-create.py
@@ -9,6 +9,7 @@ from cirrus.google_cloud.errors import GoogleAuthError
 
 from fence.config import config
 from fence.jwt.validate import validate_jwt
+from fence.utils import create_client
 from fence.models import (
     AccessPrivilege,
     Project,
@@ -33,6 +34,8 @@ from fence.scripting.fence_create import (
     verify_bucket_access_group,
     _verify_google_group_member,
     _verify_google_service_account_member,
+    list_client_action,
+    modify_client_action,
 )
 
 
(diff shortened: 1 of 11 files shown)


tests/scripting/test_fence-create.py
```diff
@@ -871,3 +876,57 @@ def test_delete_expired_service_account_keys_both_user_and_client(
...
+    to auto_approve = False
+    """
+
+
+def test_create_group(db_session):
+    group_name = "test_group_123"
+    data = {
+        "groups": {
+            group_name: {
+                "projects": [{"auth_id": "test_project_123", "privilege": "read"}]
```
paulineribeyre: so i think there is indeed a bug in the group creation. if you don't specify any projects, the group doesn't get created. if you do, it does.

the reason it gets created if you specify projects is that instances of AccessPrivilege get created during `grant_project_to_group_or_user`
https://github.com/uc-cdis/fence/blob/4.2.2/fence/scripting/fence_create.py#L352

this block in the AccessPrivilege code specifies `cascade="all"` , which includes the `cascade=save-update` option.
https://github.com/uc-cdis/userdatamodel/blob/2.0.1/userdatamodel/user.py#L292-L298

that's when the group gets created: https://docs.sqlalchemy.org/en/13/orm/cascades.html#save-update "all the objects associated with it via this relationship() should also be added to that same Session"

paulineribeyre: let's also test the group creation without projects, and fix the code.

PR #690, opened 2019-09-03
feat(signed-urls): don't rely on access in token, hit db

Avantol13:

### New Features
- rely on db for signed url authz checks, don't rely on incomplete token information

### Improvements
- truncate user access in token instead of completely removing

### DIFF
--- a/fence/jwt/token.py
+++ b/fence/jwt/token.py
@@ -367,7 +367,6 @@ def generate_signed_access_token(
                 "name": user.username,
                 "is_admin": user.is_admin,
                 "google": {"proxy_group": user.google_proxy_group_id},
-                "projects": {},
             }
         },
         "azp": client_id or "",
@@ -383,13 +382,17 @@ def generate_signed_access_token(
         claims["context"]["user"]["projects"] = dict(user.project_access)
     else:
         # truncate to configured number of projects in token
-        claims["context"]["user"]["projects"] = {
-            k: dict(user.project_access)[k]
-            for k in list(dict(user.project_access))[: config["TOKEN_PROJECTS_CUTOFF"]]
-        }
+        projects = dict(user.project_access)
+        for key in list(projects.keys())[1:]:
+            del projects[key]
+        claims["context"]["user"]["projects"] = projects
         logger.warning(
             "NOT including project_access = {} in claims for user {} because there are too many projects for the token\n".format(
-                user.project_access, user.username
+                {
+                    k: dict(user.project_access)[k]
+                    for k in set(dict(user.project_access)) - set(projects)
+                },
+                user.username,
             )
         )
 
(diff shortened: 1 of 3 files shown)


fence/jwt/token.py
```diff
@@ -381,6 +382,11 @@ def generate_signed_access_token(
     if len(dict(user.project_access)) < config["TOKEN_PROJECTS_CUTOFF"]:
         claims["context"]["user"]["projects"] = dict(user.project_access)
     else:
+        # truncate to configured number of projects in token
+        claims["context"]["user"]["projects"] = {
+            k: dict(user.project_access)[k]
+            for k in list(dict(user.project_access))[: config["TOKEN_PROJECTS_CUTOFF"]]
```
paulineribeyre: i found this which i think is easier to read and would let you remove the if/else
```
import itertools
n = config["TOKEN_PROJECTS_CUTOFF"]
truncated_projects = dict(itertools.islice(user.project_access.items(), n))
claims["context"]["user"]["projects"] = truncated_projects
```

fence/jwt/token.py
```diff
@@ -367,6 +367,7 @@ def generate_signed_access_token(
                 "name": user.username,
                 "is_admin": user.is_admin,
                 "google": {"proxy_group": user.google_proxy_group_id},
+                "projects": {},
```
paulineribeyre: i don't think you need this

paulineribeyre: actually we could just do `"projects": truncated_projects ` here

fence/jwt/token.py
Avantol13: I did see that as an option, my opinion is that the dictionary comprehension is a bit clearer but I suppose that's up for debate. I could remove the if else with the current version too actually... :thinking:

paulineribeyre:
```
projects = dict(user.project_access)
for key in list(projects.keys())[config["TOKEN_PROJECTS_CUTOFF"]:]
  del projects[key]
```
maybe?

fence/jwt/token.py
```diff
@@ -381,9 +381,18 @@ def generate_signed_access_token(
     if len(dict(user.project_access)) < config["TOKEN_PROJECTS_CUTOFF"]:
         claims["context"]["user"]["projects"] = dict(user.project_access)
     else:
+        # truncate to configured number of projects in token
+        projects = dict(user.project_access)
+        for key in list(projects.keys())[1:]:
```
paulineribeyre: why 1?
```suggestion
        for key in list(projects)[config["TOKEN_PROJECTS_CUTOFF"]:]:
```

Avantol13: Whoops, applied your fix

PR #695, opened 2019-09-12
Rename rbac

vpsx:
PXP-3071 

### Improvements
Rename rbac to authz/abac depending on context. In code, docs, useryamls.


fence/config-default.yaml
```diff
@@ -451,7 +451,7 @@ HTTP_PROXY:
 
 # //////////////////////////////////////////////////////////////////////////////////////
 # MICROSERVICE PATHS
-#   - Support `/data` endpoints & `/rbac` endpoints
+#   - Support `/data` endpoints & authz functionality
 # //////////////////////////////////////////////////////////////////////////////////////
```
Avantol13: in the test config there was another change for `role-based access control`, does that exist in this file?

fence/blueprints/data/blueprint.py
```diff
@@ -10,7 +10,7 @@
 )
 from fence.errors import Forbidden, InternalError, UserError
 from fence.utils import is_valid_expiration
-from fence.rbac import check_arborist_auth
```
Avantol13: this was really the only place where that function got imported?

fence/config-default.yaml
vpsx: Oho! Good catch thanks!

fence/blueprints/data/blueprint.py
vpsx: seems like it 😛

PR #706, opened 2019-10-08
feat(userinfo): add arborist auth mapping to the userinfo endpoint

Avantol13:
### New Features
- add arborist resource mapping to userinfo endpoint


Avantol13:
Failing jenkins test is a known issue and can be ignored `Export default manifest, mount it and check manifest name`. tests/portal is being updated to fix it, this PR should not have changed anything in a way that would cause this issue

PR #715, opened 2019-10-10
fix(synapse): fix Synapse issues

fantix:
### Improvements
- Use separate users for Synapse with suffix
- Up to date Synapse claims
- Fixed bug that `additional_info` not updating
- Fixed bug that users not in `user.yaml` don't have Arborist users

### DIFF
--- a/fence/blueprints/login/synapse.py
+++ b/fence/blueprints/login/synapse.py
@@ -19,20 +19,29 @@ class SynapseLogin(DefaultOAuth2Login):
 class SynapseCallback(DefaultOAuth2Callback):
     def __init__(self):
         super(SynapseCallback, self).__init__(
-            idp_name=IdentityProvider.synapse, client=flask.current_app.synapse_client
+            idp_name=IdentityProvider.synapse,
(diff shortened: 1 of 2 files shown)


fence/blueprints/login/synapse.py
```diff
@@ -19,20 +19,29 @@ def __init__(self):
...
+        user.email = token_result["email_verified"]
         user.display_name = "{given_name} {family_name}".format(**token_result)
-        if user.additional_info is None:
-            user.additional_info = {}
-        user.additional_info.update(token_result)
+        info = {}
+        if user.additional_info is not None:
+            info.update(user.additional_info)
+        info.update(token_result)
+        info.pop("fence_username", None)
```
vpsx: is this supposed to be `username_field`?

fantix: Yes it works here if:
```python
info.pop(self. username_field, None)
```
But after further looking, I feel like we shouldn't do that - we don't want to pop `username_field` if it is in the future no longer an artificial field.

PR #718, opened 2019-10-15
Adds non-aws s3 support e.g. ceph, swift, minio 

bwalsh:
Enables ceph support via ceph's s3 gateway. Adds support for s3 endpoint, which allows fence to connect to s3 instances other than AWS.

### New Features

Following new properties read  from fence-config.yaml:

  * S3_BUCKETS._bucket_
    * `signature_version`   # v4: ['s3v4' , 'v4' , None, ''] v2: ['s3'] 
      *  if v2, uses existing fence/resources/aws/boto_manager:: presigned_url
    * `server_side_encryption`  # signed url includes encryption, default True
      *  set to false to turn off s3 v2 server side encryption
  * AWS_CREDENTIALS._credential_
     * `endpoint_url` # non aws s3 endpoint
       * controls s3 url format

### Breaking Changes

None

###  Bug Fixes

None

### Improvements

Enables ceph support via ceph's s3 gateway

### Dependency updates

None

### Deployment changes

None


fence/blueprints/data/indexd.py
```diff
@@ -603,6 +637,12 @@ def get_signed_url(
             self.bucket_name(), aws_creds, expires_in
         )
 
+        # format url for non-aws s3 endpoint
+        if 'endpoint_url' in credential:
+            http_url = "{}/{}".format(
+                credential['endpoint_url'], self.bucket_name()
```
paulineribeyre: has this been tested with public data? see `return http_url` a few lines below

fence/config-default.yaml
```diff
@@ -433,6 +437,12 @@ S3_BUCKETS:
   bucket4:
     cred: 'CRED1'
     role-arn: 'arn:aws:iam::role1'
+  bucket5:
+    cred: 'CRED3'
+    region: default
+    signature_version: s3    # v4: ['s3v4' , 'v4' , None, ''] v2: ['s3']
```
paulineribeyre: why not use `signature_version: v2`?

bwalsh: @paulineribeyre  Thanks so much for the feedback.  I considered `v2`, however, I elected to conform with the underlying boto package that uses [s3](https://github.com/boto/botocore/blob/3576ba67f2e51d197f278ef4cd90d1c54473b11a/botocore/data/endpoints.json#L2743)

fence/blueprints/data/indexd.py
bwalsh: changes made.

fence/blueprints/data/indexd.py
```diff
@@ -588,6 +588,40 @@ def get_bucket_region(self):
...
 
+    def get_bucket_signature_version(self):
+        """Returns signature_version, None if not set."""
+        s3_buckets = get_value(
+            config, "S3_BUCKETS", InternalError("buckets not configured")
+        )
+        if len(s3_buckets) == 0:
+            return None
+
+        bucket_cred = s3_buckets.get(self.bucket_name())
```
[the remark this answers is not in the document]
bwalsh: changes made.

bwalsh:
@paulineribeyre :  Thanks for your comments. Suggested changes made.

paulineribeyre:
Hi @bwalsh! First of all, apologies for taking so long to answer this. This is a good feature to have but we are not going to merge this PR right now, because supporting Ceph is not currently on our raodmap so we are not ready to support it. Merging this would involve doing some manual testing with Ceph and adding some automated testing, which we don't have the bandwidth for right now. We do want to keep the PR around for later though - thank you again for contributing

bwalsh:
Hi @paulineribeyre, thanks for the feedback.

One idea - make the "create signed url" function a separate service or fence plugin.   That way, fence could ship with it's standard functionality and users could customize and deploy their own url signing strategy.

This would allow better integration without encumbering the project with ever increasing test infrastructure.

Thanks again,
-b

paulineribeyre:
@bwalsh Thank you for the suggestion, we'll consider it as part of our long term roadmap!

PR #723, opened 2019-10-22
PXP-4312 oauth2/token JSON instead of HTML error

paulineribeyre:
Fence now returns JSON instead of using the HTML error template for the `oauth2/token` endpoint in case of error (such as expired or blacklisted refresh token) to conform to the oauth2 spec (https://tools.ietf.org/html/rfc6749#section-5.2) - example:
```
{
    "error": "invalid_grant",
    "error_description": "Authentication Error: Signature has expired"
}
```

There is no need to handle `OAuth2Error` here because `create_token_response()` is already handling it (Authlib 0.11):
- https://github.com/lepture/authlib/blob/v0.11/authlib/oauth2/rfc6749/authorization_server.py#L173-L189
- https://github.com/lepture/authlib/blob/v0.11/authlib/flask/oauth2/authorization_server.py#L100-L103

### Improvements
- Fence oauth2/token endpoint now returns a JSON response in case of expired or blacklisted refresh token


fence/blueprints/oauth2.py
```diff
@@ -282,7 +289,20 @@ def get_token(*args, **kwargs):
     See the OpenAPI documentation for detailed specification, and the OAuth2
     tests for examples of some operation and correct behavior.
     """
-    return server.create_token_response()
+    try:
+        response = server.create_token_response()
+    except Exception as e:
+        details, status_code = get_error_details_and_status(e)
+        error_id = _get_error_identifier()
```
Avantol13: since we're using this function outside the original module, can we remove the leading `_`?

fence/blueprints/oauth2.py
```diff
@@ -282,7 +289,20 @@ def get_token(*args, **kwargs):
     See the OpenAPI documentation for detailed specification, and the OAuth2
     tests for examples of some operation and correct behavior.
     """
-    return server.create_token_response()
+    try:
+        response = server.create_token_response()
+    except Exception as e:
```
Avantol13: are there specific exceptions you can catch for known potential issues? I'd feel more comfortable _not_ catching any issues (which may include Python errors we didn't expect)

fence/jwt/token.py
```diff
@@ -376,29 +376,29 @@ def generate_signed_access_token(
...
-          logger.warning(
-              "NOT including project_access = {} in claims for user {} because there are too many projects for the token\n".format(
-                  {
-                      k: dict(user.project_access)[k]
-                      for k in set(dict(user.project_access)) - set(projects)
-                  },
-                  user.username,
-              )
-          )
+        # NOTE: "THIS IS A TERRIBLE STOP-GAP SOLUTION SO THAT USERS WITH
```
Avantol13: how was this working? I thought Python 3 enforced same indentation within a module..

paulineribeyre: i have no idea 😆

fence/blueprints/oauth2.py
paulineribeyre: well previously _all_ exceptions were displayed in the html template, now _all_ exceptions return JSON
I think it's fine to catch all exceptions here, because we want this endpoint to always return JSON, even in case of internal error

paulineribeyre: i can log the traceback though so that we can debug if there is a 500 error

vpsx: return JSON in a `finally` clause without catching exception? Not sure--but an idea

vpsx: https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions

paulineribeyre:
```
# OAuth2Error
{
    "error": "invalid_client"
}

# JWTError
{
    "error": "invalid_request",
    "error_description": "token is blacklisted"
}

# JWTExpiredError
{
    "error": "invalid_request",
    "error_description": "Authentication Error: Signature has expired"
}
```

fence/blueprints/oauth2.py
```diff
@@ -282,7 +286,22 @@ def get_token(*args, **kwargs):
...
+    except OAuth2Error as e:
+        body = dict(error.get_body())  # returns list of tuples
+        response = flask.Response(
+            json.dumps(body), mimetype="application/json", status=error.status_code
+        )
+        response.headers = error.get_headers()
+    except (JWTError, JWTExpiredError) as e:
+        # fence.jwt.errors.JWTError: blacklisted refresh token.
+        # JWTExpiredError (cdiserrors.AuthNError subclass): expired token.
+        body = {"error": "invalid_request", "error_description": e.message}
```
paulineribeyre: could also be `access_denied` - what do you think?

fence/blueprints/oauth2.py
```diff
@@ -282,7 +286,22 @@ def get_token(*args, **kwargs):
     See the OpenAPI documentation for detailed specification, and the OAuth2
     tests for examples of some operation and correct behavior.
     """
-    return server.create_token_response()
+    try:
+        response = server.create_token_response()
+    except OAuth2Error as e:
+        body = dict(error.get_body())  # returns list of tuples
+        response = flask.Response(
+            json.dumps(body), mimetype="application/json", status=error.status_code
```
Avantol13: does mimetype and status not end up in headers? e.g. is the next line just overwriting this information?

paulineribeyre: in my postman test i'm getting the correct mimetype and code

fence/blueprints/oauth2.py
```diff
@@ -282,7 +286,22 @@ def get_token(*args, **kwargs):
...
+        )
+        response.headers = error.get_headers()
+    except (JWTError, JWTExpiredError) as e:
+        # fence.jwt.errors.JWTError: blacklisted refresh token.
+        # JWTExpiredError (cdiserrors.AuthNError subclass): expired token.
+        body = {"error": "invalid_request", "error_description": e.message}
+        response = flask.Response(
+            json.dumps(body), mimetype="application/json", status=e.code
```
Avantol13: I think the status is supposed to _always_ be 400. OIDC: `The HTTP response body uses the application/json media type with HTTP response code of 400`. We should probably hard-code that here, I'm sure authlib handles that for their errors

fence/blueprints/oauth2.py
Avantol13: we should probably handle both of these as `invalid_grant` error. 

from the oidc spec: `The parameters of the Token Error Response are defined as in Section 5.2 of OAuth 2.0 [RFC6749]`
from the OAuth 2 spec: https://tools.ietf.org/html/rfc6749#section-5.2

```
         invalid_grant
               The provided authorization grant (e.g., authorization
               code, resource owner credentials) or refresh token is
               invalid, expired, revoked, does not match the redirection
               URI used in the authorization request, or was issued to
               another client.
```

fence/blueprints/oauth2.py
Avantol13: :+1:

PR #728, opened 2019-10-29
Support ORCID login

paulineribeyre:
ORCID login is set up at login.bionimbus so we need to specify `idp=orcid` in the multi-tenant fence login URL. Until now we were only handling `idp=shibboleth`. We need a new config field to specify which provider to use

### Improvements
- Add "fence_idp" configuration parameter for LOGIN_OPTIONS to specify which multi-tenant Fence provider to use


vpsx:
lgtm but Alex should definitely look at it :P

fence/config-default.yaml
```diff
@@ -209,24 +209,32 @@ SESSION_ALLOWED_SCOPES:
...
 # Multiple options can be configured with the same idp.
-# "desc" and "secondary" are optional.
-# If the idp is fence, a list of "shib_idps" can be configured for
-# InCommon login.
+# - if provider_id is "fence", "fence_idp" can be any of the providers
+# supported by the other Fence. If not specified, will default to NIH login.
+# - if provider_id is "fence" and fence_idp is "shibboleth", a list of
+# "shib_idps" can be configured for InCommon login. If not specified, will
+# default to NIH login.
+# - "desc" and "secondary" are optional.
```
Avantol13: what are they though? desc = dscription? what is secondary?

Avantol13:
for posterity: I don't like having `secondary` config for front-end buttons in fence configuration. It doesn't change fence behavior so it shouldn't belong in the fence configuration (my opinion). The front-end should have some configuration of what idps it wants to support and it has to be a subset of what fence has configured. This is just coupling the front-end and fence, which I generally don't agree with. Anyway, it wasn't added in this PR but this is the first I saw it. cc @fantix

PR #733, opened 2019-11-05
PXP-4641 Feat/username limit

mfshao:
This PR fulfills https://ctds-planx.atlassian.net/browse/PXP-4641

Lift up the length limit for username filed, which is causing issues now for users with long email addresses

Works with https://github.com/uc-cdis/userdatamodel/pull/66

### Improvements
- Lift length limit for username filed from 40 to 255 chars

### Dependency updates
- Bumps userdatamodel to 2.2.0
- Bumps setuptools to at least 40.3.0

### DIFF
--- a/fence/models.py
+++ b/fence/models.py
@@ -576,8 +576,6 @@ def migrate(driver):
             % (User.__tablename__, str(User.username.type))
         )
         with driver.session as session:
-            session.execute(to_timestamp)
-        with driver.session as session:
             session.execute(
                 'ALTER TABLE "{}" ALTER COLUMN username TYPE {};'.format(
                     User.__tablename__, str(User.username.type)
--- a/fence/models.py
+++ b/fence/models.py
@@ -567,6 +567,23 @@ def migrate(driver):
                 )
             )
 
+    # username limit migration
+
+    table = Table(User.__tablename__, md, autoload=True, autoload_with=driver.engine)
+    if str(table.c.username.type) != str(User.username.type):
+        print(
+            "Altering table %s column username type to %s"
+            % (User.__tablename__, str(User.username.type))
+        )
+        with driver.session as session:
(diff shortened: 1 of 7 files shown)


paulineribeyre:
do we need a db migration?

mfshao:
> do we need a db migration?

oh good catch! I'll add some scripts into `migrate`

requirements.txt
```diff
@@ -35,7 +35,7 @@ setuptools==36.6.0
 six==1.11.0
 SQLAlchemy==1.3.3
 temps==0.3.0
-userdatamodel==2.1.1
+git+https://github.com/uc-cdis/userdatamodel.git@aad4a5100ebcf905c5fe21b4bc2d1e2a0a0cb62e
```
mfshao: TODO: change after releasing `userdatamodel`

requirements.txt
```diff
@@ -31,11 +31,11 @@ pytest==3.2.3
 python_dateutil==2.6.1
 PyJWT==1.5.3
 requests>=2.18.0<3.0.0
-setuptools==36.6.0
+setuptools>=40.3.0
```
mfshao: note: have to do this to get docker build to pass

fence/models.py
```diff
@@ -567,6 +567,23 @@ def migrate(driver):
...
+    # username limit migration
+
+    table = Table(User.__tablename__, md, autoload=True, autoload_with=driver.engine)
+    if str(table.c.username.type) != str(User.username.type):
+        print(
+            "Altering table %s column username type to %s"
+            % (User.__tablename__, str(User.username.type))
+        )
+        with driver.session as session:
+            session.execute(to_timestamp)
```
vpsx: why are we doing this?

mfshao: should remove! thanks

PR #738, opened 2019-11-19
PXP-4319 Feat/cert audit log

mfshao:
This PR is a part of https://ctds-planx.atlassian.net/browse/PXP-4319

Get a trigger function for `certificate` table, to record when a user has submitted a ToU/PP cert

This PR denpends on `userdatamodel` PR https://github.com/uc-cdis/userdatamodel/pull/67

### Improvements
- New trigger function for `certificate` table, to record when a user has submitted a ToU/PP cert

### Dependency updates
- Bump userdatamodel to 2.3.0


fence/models.py
```diff
@@ -746,6 +746,45 @@ def migrate(driver):
...
+            INSERT INTO cert_audit_logs (timestamp, operation, user_id, username,  old_values, new_values)
+            SELECT now(), 'UPDATE', "User".id, "User".username, row_to_json(OLD), row_to_json(NEW)
+            FROM certificate INNER JOIN application ON certificate.application_id = application.id
+            INNER JOIN "User" ON application.user_id = "User".id;
+            RETURN NEW;
+        ELSIF (TG_OP = 'INSERT') THEN
+            INSERT INTO cert_audit_logs (timestamp, operation, user_id, username, new_values)
+            SELECT now(), 'INSERT', "User".id, "User".username, row_to_json(NEW)
+            FROM certificate INNER JOIN application ON certificate.application_id = application.id
+            INNER JOIN "User" ON application.user_id = "User".id;
```
vpsx: I don't think this is right 🤔 --the `INNER JOIN` result set will grow as more users accept more ToU/PPs, and then each time this is triggered, the `SELECT` will just insert all of the rows in the result set as new audit log entries, but all with the same `new_values`. Then there is not only an exponentially growing audit log, but the log entries reports changes that didn't occur.

vpsx: Probably same situation with the other JOINS but I haven't tested those yet because I can't get the triggers to fire with direct `psql` updates 😂 🙃

mfshao: umm no. The `FOR EACH ROW` in https://github.com/uc-cdis/fence/blob/c09b013ccf809995664e9ab510b3087b7a063019/fence/models.py#L785 means this trigger will only be called once on rows that were affected by that operations (`DELETE`, `UPDATE`, `INSERT`), so only changes will be captured

vpsx: Right, the trigger is only called once, but the function itself inserts multiple rows, because the select from inner join returns multiple rows. 

You can try it yourself--it's very simple: 
1. Clear relevant tables
2. Log in as user A and accept cert
3. Log in as user B and accept cert
4. Check cert_audit_logs

PR #739, opened 2019-11-27
(PXP-4777) Feat/synapse userid

mfshao:
Minor changes for Synapse login (BRAIN Dream challenge)

### Bug Fixes
- Fixed an error causing Synapse users' email being incorrectly populated

### Improvements
- Request numerical `userid` from Synapse
-  Use `userid` from Synapse as part of username

### DIFF
--- a/fence/resources/openid/synapse_oauth2.py
+++ b/fence/resources/openid/synapse_oauth2.py
@@ -15,10 +15,9 @@ class SynapseOauth2Client(Oauth2ClientBase):
 
     """
 
-    REQUIRED_CLAIMS = {"given_name", "family_name", "email", "email_verified"}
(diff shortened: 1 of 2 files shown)


fence/resources/openid/synapse_oauth2.py
```diff
@@ -125,7 +124,7 @@ def get_user_id(self, code):
                         return dict(error="Required claim {} not found".format(claim))
                 else:
                     rv[claim] = value
-            rv["fence_username"] = rv["email"] + " (via Synapse)"
```
paulineribeyre: according to ticket requirements it should be `12345 (Synapse ID)`

mfshao:
> from the ticket requirements: `We should store the user's email in fence, so we can include it in the user export`
> not sure if that's the case?

yes so the user's email is populated from the `email` field in the token

PR #744, opened 2019-12-18
Feat/cleversafe url

Avantol13:
### New Features
- New config to specify an endpoint url for an s3 bucket (to support cleversafe)


fence/blueprints/data/indexd.py
```diff
@@ -594,13 +594,24 @@ def get_signed_url(
...
         )
-
-        http_url = "https://{}.s3.amazonaws.com/{}".format(
-            self.parsed_url.netloc, self.parsed_url.path.strip("/")
+        s3_buckets = get_value(
+            config, "S3_BUCKETS", InternalError("buckets not configured")
         )
 
+        bucket = self.bucket_name()
+        current_bucket = s3_buckets.get(self.bucket_name())
```
paulineribeyre:
```suggestion
        bucket_name = self.bucket_name()
        bucket = s3_buckets.get(bucket_name)
```

fence/blueprints/data/indexd.py
```diff
@@ -605,13 +605,24 @@ def get_signed_url(
...
-            self.parsed_url.netloc, self.parsed_url.path.strip("/")
+        s3_buckets = get_value(
+            config, "S3_BUCKETS", InternalError("buckets not configured")
         )
 
+        bucket_name = self.bucket_name()
+        bucket = s3_buckets.get(bucket_name)
+
+        if current_bucket and current_bucket.get("endpoint_url"):
+            http_url = current_bucket["endpoint_url"].strip("/") + "/{}/{}".format(
```
paulineribeyre: `current_bucket` -> `bucket`
sorry

fence/config-default.yaml
```diff
@@ -487,6 +487,8 @@ AWS_CREDENTIALS:
 S3_BUCKETS:
   bucket1:
     cred: 'CRED1'
+    # optionally you can manually specify an s3-compliant endpoint for this bucket
+    endpoint_url: 'https://cleversafe.example.com/'
```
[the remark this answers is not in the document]
Avantol13: I don't think so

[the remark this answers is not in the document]
paulineribeyre: it looks like the code is using the region, so i'm not sure how it's working in prod... maybe they configured it with a fake region field. if not, it would fail trying to get the region from aws

fence/blueprints/data/indexd.py
```diff
@@ -638,11 +638,15 @@ def get_signed_url(
...
 
-        region = self.get_bucket_region()
-        if not region:
-            region = flask.current_app.boto.get_bucket_region(
-                self.parsed_url.netloc, credential
-            )
+        # only attempt to get the region when we're not specifying an
+        # s3-compatible endpoint URL (ex: no need for region when using cleversafe)
+        region = None
+        if not current_bucket.get("endpoint_url"):
```
paulineribeyre:
```suggestion
        if not bucket.get("endpoint_url"):
```

PR #745, opened 2019-12-18
Chore/dbgap backup

diw2:
This PR is for https://ctds-planx.atlassian.net/browse/PXP-4695
Storing backups of telemetry files in S3. 

- Implemented 
fence-create dbgap-download-access-files 
fence-create sync


fence/sync/sync_users.py
```diff
@@ -966,24 +970,48 @@ def sync(self):
...
+            self._download(self.session)
+        else:
+            with self.driver.session as s:
+                self._download(s)
+
+    def _download(self, sess):
+        """
+        Download files from dbgap server.
+        """
+        folderdir = os.path.join("/mnt/shared/", str(self.folder))
```
frickjack: why are we hard coding a `/mnt/shared/` prefix rather than just use the folder passed by the user directly?

fence/sync/sync_users.py
```diff
@@ -995,7 +1023,7 @@ def _sync(self, sess):
             dict(list(zip(dbgap_file_list, permissions))), encrypted=True, sess=sess
         )
         try:
-            shutil.rmtree(tmpdir)
+            shutil.rmtree(folderdir)
```
frickjack: Do we still want to delete folderdir after the sync?

bin/fence-create
```diff
@@ -412,6 +434,19 @@ def main():
...
+    elif args.action == "dbgap-download-access-files":
+        download_dbgap_files(
+            dbGaP,
+            STORAGE_CREDENTIALS,
+            DB,
+            projects=args.project_mapping,
+            is_sync_from_dbgap_server=str2bool(args.sync_from_dbgap),
+            sync_from_local_csv_dir=args.csv_dir,
+            sync_from_local_yaml_file=args.yaml,
+            folder=args.folder,
```
frickjack: why do we need all these extra flags for `download_dbgap_files()` ?

bin/fence-create
```diff
@@ -153,6 +154,27 @@ def parse_arguments():
...
+        "--sync_from_dbgap", help="sync from dbgap server True/False", default="False"
+    )
+    dbgap_download.add_argument(
+        "--arborist",
+        help="the base URL for the arborist service to sync to",
+        default=None,
+    )
+    dbgap_download.add_argument(
+        "--folder", required=False, help="destination where file will be saved", default=None,
+    )
```
frickjack: why does dbgap_download need all the non-folder stuff?

bin/fence-create
```diff
@@ -153,6 +154,27 @@ def parse_arguments():
         help="the base URL for the arborist service to sync to",
         default=None,
     )
+    dbgap_sync.add_argument(
+        "--folder", required=False, help="destination where file will be saved", default=None,
```
paulineribeyre:
```suggestion
        "--folder", required=False, help="destination where dbGaP whitelist files are saved", default=None,
```

bin/fence-create
```diff
@@ -153,6 +154,27 @@ def parse_arguments():
...
+    dbgap_download.add_argument(
+        "--sync_from_dbgap", help="sync from dbgap server True/False", default="False"
+    )
+    dbgap_download.add_argument(
+        "--arborist",
+        help="the base URL for the arborist service to sync to",
+        default=None,
+    )
+    dbgap_download.add_argument(
+        "--folder", required=False, help="destination where file will be saved", default=None,
```
paulineribeyre:
```suggestion
        "--folder", required=False, help="destination where dbGaP whitelist files will be saved", default=None,
```

fence/sync/sync_users.py
```diff
@@ -276,6 +276,7 @@ def __init__(
         sync_from_local_csv_dir=None,
         sync_from_local_yaml_file=None,
         arborist=None,
+        folder="temp_dbgap_files",
```
paulineribeyre: you should do `folder=None` here and `folder = folder or "temp_dbgap_files"` inside the `__init__` function, because something like `UserSyncer(..., folder=None)` can break this

docs/usersync.md
```diff
@@ -159,3 +159,4 @@ The [gen3users CLI](https://github.com/uc-cdis/gen3users) includes a user.yaml v
 pip install gen3users
 gen3users validate user.yaml
 ```
+
```
paulineribeyre: it'd be a good idea to mention the new manifest config var in this doc

fence/sync/sync_users.py
```diff
@@ -966,24 +970,48 @@ def sync(self):
...
-            try:
-                if self.protocol == "sftp":
-                    self._get_from_sftp_with_proxy(tmpdir)
-                else:
-                    self._get_from_ftp_with_proxy(tmpdir)
-                dbgap_file_list = glob.glob(os.path.join(tmpdir, "*"))
-            except Exception as e:
-                self.logger.error(e)
-                exit(1)
+            if self.folder and os.path.exists(folderdir):
```
paulineribeyre: `self.folder` has a default value so it will never be None
```suggestion
            if os.path.exists(folderdir):
```

requirements.txt
```diff
@@ -1,6 +1,7 @@
 Authlib==0.11
 addict==2.2.1
 authutils>=4.0.0<5.0.0
+awscli
```
paulineribeyre: i don't think you use that?

diw2: It is used to setup the env to be able to run aws s3 cp in usersync jobs

fence/sync/sync_users.py
```diff
@@ -993,29 +1005,27 @@ def _process_dbgap_files(self, dbgap_config, sess):
...
+        folderdir = os.path.join(
+            "/var/www/fence/", str(self.folder), str(hostname), str(username)
+        )
+
         try:
-            if protocol == "sftp":
-                self._get_from_sftp_with_proxy(server, tmpdir)
+            if os.path.exists(folderdir):
+                # add logic to list everything for hosts subfolders os.walk
+                # to support backwards compatible
```
paulineribeyre: that's not needed anymore right? let's remove the comment?

requirements.txt
diw2: This was added in fence because this is used in this container and to keep consistent with what we have for awshelper https://github.com/uc-cdis/cloud-automation/blob/master/Docker/awshelper/Dockerfile#L42.  In fence we have two ways, one is adding in fence dockerfile same as awshelper or adding in the requirement doc. 😄

docs/usersync.md
```diff
@@ -165,6 +171,7 @@ Usersync gives users "read" and "read-storage" permissions to the dbGaP studies.
 
 > Note: The dbGaP authorization files contain consent codes that can be parsed by usersync: [more details here](dbgap_info.md). This simplified example does not include consent code parsing.
 
+More deployment details can be found here: https://github.com/uc-cdis/cdis-wiki/blob/master/ops/DBGAP-backup-deployment.md
```
paulineribeyre, line 174: this line should be moved to the other section

docs/usersync.md
```diff
@@ -36,6 +36,12 @@ dbGaP:
 
 An example can be found in the config used for unit testing [tests/test-fence-config.yaml](https://github.com/uc-cdis/fence/blob/master/tests/test-fence-config.yaml)
 
+### Enable dbGaP backup to s3:
+Add below to manifest.json to global block:
```
paulineribeyre, line 40: could you just add a couple sentences about what this does - "copies dbGaP authorization files to <dbgap_backup_bucket> with the following folder structure......."
including what happens if multiple sftp servers are configured

PR #747, opened 2020-01-02
fix(split): create a secret template for cloud automation to refer

qingyashu:
This PR is to add a template config file for secret configs, so that the config helper script in cloud-automation could refer it, and split the config into public and secret part. 
See [this PR](https://github.com/uc-cdis/cloud-automation/pull/1035) for detail

### Improvements
- add a secret config template so that cloud-automation script could refer 
- add warning in default config to avoid inline comments, because those might be wiped by cloud-automation's config helper script


fence/config-default.yaml
```diff
@@ -6,6 +6,8 @@
 # - This is initially configured for minimal local development with reasonable defaults.
 # - Descriptions for each of the configurations (if any) will be *above* the variable as
 #   comments.
+# - Please avoid using inline comments, because inline comments will be wiped by cloud 
+#   automation's config helper script
```
paulineribeyre: maybe need to add a comment about lists and dicts, see https://github.com/uc-cdis/cloud-automation/pull/1035#discussion_r362994371

paulineribeyre:
closing - done in https://github.com/uc-cdis/cloud-automation/pull/1067

PR #749, opened 2020-01-08
Fix/make region required for s3 bucket config

mjmartinson:
### Improvements
- Decrease response time when getting S3 presigned URLs by calling "get_bucket_region" at startup instead of when processing requests

### Deployment changes
- In fence-config, the region field of S3_BUCKETS is now required: if not specified, Fence will print a warning and try to call GetBucketLocation. Fence will error at startup if that doesn't work.


fence/config-default.yaml
```diff
@@ -489,7 +489,7 @@ S3_BUCKETS:
     cred: 'CRED1'
   bucket2:
     cred: 'CRED2'
-    region: 'us-east-1' #optional but if specified avoids a call to GetBucketLocation which you may lack the AWS ACLs for.
+    region: 'us-east-1' #REQUIRED because it will avoids a call to get_bucket_region which increase response time by 50%
   bucket3:
     cred: '*'
   bucket4:
```
paulineribeyre: let's add the region for all buckets and move the explanatory comment to before the `S3_BUCKETS:` line

fence/__init__.py
```diff
@@ -165,6 +165,31 @@ def public_keys():
...
 
+def check_s3_bucket():
+    """
+    Check the s3_bucket config
+    """
+    s3_buckets = get_value(
+        config, "S3_BUCKETS", InternalError("buckets not configured")
+    )
+    if len(s3_buckets) == 0:
+        return None
```
paulineribeyre: we don't need this - if the list is empty, the function will just terminate without running the `for` loop below

fence/__init__.py
```diff
@@ -165,6 +165,31 @@ def public_keys():
             {"keys": [(keypair.kid, keypair.public_key) for keypair in app.keypairs]}
         )
 
+def check_s3_bucket():
```
paulineribeyre: don't forget to call this function in `app_config()` :p
and please rename it `_check_s3_buckets()` because it won't be used in other files

fence/__init__.py
```diff
@@ -165,6 +165,31 @@ def public_keys():
...
+    """
+    Check the s3_bucket config
+    """
+    s3_buckets = get_value(
+        config, "S3_BUCKETS", InternalError("buckets not configured")
+    )
+    if len(s3_buckets) == 0:
+        return None
+
+    for i, bucket in enumerate(config["S3_BUCKETS"]):
```
paulineribeyre: let's use `for bucket_name, bucket_details in s3_buckets.items()` instead - we don't need `i` because providing the name of the bucket in error messages is more useful than providing the index

fence/__init__.py
```diff
@@ -165,6 +165,31 @@ def public_keys():
...
+        cred = bucket["cred"]
+        region == bucket["region"]
```
paulineribeyre: if `cred` is not specified in `bucket`, this line will break before reaching the `if not cred: raise` block.
you need to do `cred = bucket.get("cred")` instead to avoid errors.
same for `region`

fence/__init__.py
```diff
@@ -165,6 +165,31 @@ def public_keys():
...
+        if not cred:
+            raise ValueError(
+                "No cred field for S3_BUCKET: {}. The cred field is required.".format(i))
+            if cred != "*" and cred not in config["AWS_CREDENTIALS"]:
+                raise ValueError(
+                    "No credentials for S3-BUCKET: {} in AWS_CREDENTIALS".format(i))
+        if not region or region == "":
```
paulineribeyre: `not region` catches all "null" values, so it already includes `""` (as well as `None`, `0`, `[]`, `{}`, etc)

fence/__init__.py
```diff
@@ -165,6 +165,31 @@ def public_keys():
...
+            print("WARNING: no region field for S3_BUCKET: {}. Providing the region field will increase"
```
paulineribeyre: let's use `logger.warning` instead of `print` here

fence/__init__.py
```diff
@@ -165,6 +165,31 @@ def public_keys():
...
+                  " response time and avoid a call to GetBucketLocation which you make lack the AWS ACLs for.")
+            region = flask.current_app.boto.get_bucket_region(
+                bucket, credential
```
paulineribeyre: `bucket` is a dictionary here, not the bucket name, so this won't work

fence/config-default.yaml
```diff
@@ -484,16 +484,21 @@ AWS_CREDENTIALS:
     aws_access_key_id: ''
     aws_secret_access_key: ''
 
+# NOTE: the region is optonal for s3_buckets, however it should be specified to avoid a 
+# call to GetBucketLocation which you make lack the AWS ACLs for.
```
paulineribeyre: 👌

tests/test-fence-config.yaml
```diff
@@ -355,17 +355,24 @@ AWS_CREDENTIALS:
...
+    region: 'us-east-1' 
   bucket3:
     cred: 'CRED1'
+    region: 'us-east-1' 
   bucket4:
     cred: '*'
+    region: 'us-east-1' 
   bucket5:
     cred: 'CRED2'
+    region: 'us-east-1'
```
paulineribeyre: let's remove the `region` for one of the test buckets to make sure that the app doesn't crash. the call to `get_bucket_region` is already mocked in the tests, so it won't try to actually get the region

mjmartinson: Remove it for just one? What about in config-deault.yaml?

paulineribeyre: yes one is enough, more if you want to, but not all

config-default should keep all of them to push people in the right direction when they setup fence

fence/blueprints/data/indexd.py
```diff
@@ -678,9 +683,14 @@ def generate_presigne_url_for_part_upload(self, uploadId, partNumber, expires_in
...
-            region = flask.current_app.boto.get_bucket_region(
-                self.parsed_url.netloc, credential
+            raise ValueError(
+                "No region specified. The region field is required for S3_BUCKETS"
             )
+            # region = flask.current_app.boto.get_bucket_region(
+            #     self.parsed_url.netloc, credential
+            # )
+            # This request takes very long, so to improve efficiency region is required for s3 buckets
+            # Calling get_bucket_region increases resonse time by 50%
```
paulineribeyre: revert this

fence/__init__.py
```diff
@@ -166,6 +168,40 @@ def public_keys():
...
+    for bucket_name, bucket_details in buckets.items():
+        cred = bucket_details.get("cred")
+        region = bucket_details.get("region")
+        if not cred:
+            raise ValueError(
+                "No cred field for S3_BUCKET: {}. The cred field is required.".format(
+                    bucket_name
+                )
+            )
+            if cred != "*" and cred not in aws_creds:
```
paulineribeyre: this code will never be reached, it's in the `if not cred` block after the exception raising

fence/__init__.py
```diff
@@ -170,29 +171,30 @@ def public_keys():
...
                 )
             )
-            if cred != "*" and cred not in aws_creds:
-                raise ValueError(
-                    "No credentials for S3-BUCKET: {} in AWS_CREDENTIALS".format(
-                        bucket_name
-                    )
+        if cred not in aws_creds and cred != "*":
+            raise ValueError(
+                "Credential (cred) for S3_BUCKET {} is not defined in AWS_CREDENTIALS".format(
```
paulineribeyre: `cred` should be the variable too :p

tests/test-fence-config.yaml
```diff
@@ -360,6 +360,7 @@ AWS_CREDENTIALS:
 S3_BUCKETS:
   bucket1:
     cred: 'CRED1'
+    region: 'us-east-1'
```
paulineribeyre: no need to add it back! we need to make sure nothing breaks when the region is missing

requirements.txt
```diff
@@ -24,7 +24,7 @@ markdown==3.1.1
 python-jose==2.0.2
 oauthlib==3.0.0
 requests_oauthlib==1.2.0
-psycopg2==2.8.3
+# psycopg2==2.8.3
```
paulineribeyre: we'll need this back :p

mjmartinson: Oh ya sorry

PR #750, opened 2020-01-09
PXP-5038 Fix usersync override

paulineribeyre:
- Update Shibboleth doc (fix mistake in diagram)
- Refactor `UserYAML.from_file()`: 1 loop over `users.items()` instead of 2 - update `UserYAML.projects` and `UserYAML.user_abac` at the same time
- Fix bug: user.yaml authz overrides sftp authz, when the user.yaml contains the same non-lowercase username as the sftp files, because `user_yaml.user_abac` was merged into `user_projects` without lowering the username so the access was not combined

### Bug Fixes
- Fix user.yaml authz information overriding other authz information


Avantol13:
tested in my dev env:
* Only user.yaml sync :heavy_check_mark: 
* user.yaml + dbgap sync :heavy_check_mark:

PR #751, opened 2020-01-13
feat(userinfo): add google primary service account email to userinfo …

Avantol13:
…for requester pays support

### New Features
- added `google_primary_service_account` email to userinfo endpoint (to support Google requester pays buckets, need to know email of service account that signed the url)

### DIFF
--- a/docs/google_architecture.md
+++ b/docs/google_architecture.md
@@ -83,7 +83,7 @@ In order to fully understand the options for requester pays support, it's import
 The easiest option for supporting requester pays is to simply bill a Google Project you already own for all access to the bucket instead of requiring end-users to supply a project to bill. This essentially makes the requester pays bucket a non-requester pays bucket, since you'll be paying for all the access. This may be a necessary solution in cases where:
 
 1) you want to serve data from a bucket you don't fully control (in other words, can't just turn "requester pays" off)
-2) you don't want end-users to have to do manual configuration in Google Cloud Platform to enable billing their project
+2) you don't want end-users (or client applications) to have to do configuration in Google Cloud Platform to enable billing their own project
 3) you/end-users don't want to have to give your application IAM permissions in a project the end-user owns to automatically enable billing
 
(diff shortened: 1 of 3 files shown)


docs/google_architecture.md
```diff
@@ -110,18 +110,25 @@ Whether you bill your own project, or require end-users to specify a billing pro
...
 * `Project IAM Admin`: to update the project's policy to give the necessary service account(s) billing permission
 * `Role Administrator`: for creating a custom role that only provides billing permission to the project
 
 > NOTE: The custom role that Fence creates contains the single permission in Google `serviceusage.services.use`.
 
+If you want to create a client application that will be able to give the right service accounts
+billing permission (option 3) then there is some additional information you should know about. The
+userinfo endpoint supplies the user's `primary_google_service_account` which is the email address of the Google Service Account attached to that user for the creation of Signed URLs.
+
+> NOTE: Users' Primary Google Service Accounts are created lazily, so it is possible that `primary_google_service_account` is `None`/`null` if the user has not previously requested data via a Google Data Access method. There must be an API call previous to reading userinfo to access data in Google that the user has access to.
```
vpsx: Is the assumption that this is OK because the client application wouldn't be trying to grant billing permission to the user's SA unless the user was already trying to access data? Or is there some kind of error message telling the user to go hit the API? Haha

fence/resources/user/__init__.py
```diff
@@ -90,6 +91,11 @@ def get_user_info(current_session, username):
         "message": "",
     }
 
+    # Note that client_id is None, which is how we store the user's SA
```
vpsx: (Nitpick) I was confused on first read--sounded like, "In order to store the user's SA, we pass client_id None here"
may I suggest replacing "store" with "retrieve"? Or saying
"User SAs are stored in db with client_id = None" 

Up to you though. Could be crystal clear to everyone else :grin:

docs/google_architecture.md
Avantol13: It's an edge case we have to discussion with whoever is implementing the client application. Basically they have to know that the primary_google_service_account only shows up _after_ an attempt to download data

fence/resources/user/__init__.py
Avantol13: oh, you're right, I'll fix it

PR #755, opened 2020-01-30
PXP-4582 user sync merges multiple dbgap sftp servers

jiaqi216:
DCF will be checking 2 sftp servers for telemetry files and merge user credentials from both. This PR allows for the configuration of any number of additional dbGaP servers in fence's config file by including a list of dbGaP objects. 

In order to have this work on DCF, we will have to modify DCF's fence config file to include the new sftp server credentials in `additional_dbGaP`

### DIFF
--- a/fence/config-default.yaml
+++ b/fence/config-default.yaml
@@ -387,50 +387,50 @@ ITRUST_GLOBAL_LOGOUT: 'https://auth.nih.gov/siteminderagent/smlogout.asp?mode=ni
 # User syncing and access can also be done throught a User Access file. See
 # fence's README for more information
 dbGaP:
-  info:
-    host: ''
-    username: ''
-    password: ''
-    port: 22
-    proxy: ''
-    proxy_user: ''
-  protocol: 'sftp'
-  decrypt_key: ''
-  # parse out the consent from the dbgap accession number such that something
-  # like "phs000123.v1.p1.c2" becomes "phs000123.c2".
-  #
-  # NOTE: when this is "false" the above would become "phs000123"
-  parse_consent_code: true
-  # A consent of "c999" can indicate access to that study's "exchange area data"
-  # and when a user has access to one study's exchange area data, they
-  # have access to the parent study's "common exchange area data" that is not study
-  # specific. The following config is whether or not to parse/handle "c999" codes
-  # for access to the common exchange area data
-  #
-  # NOTE: When enabled you MUST also provide a mapping to the
-  # `study_common_exchange_areas` from study -> parent common exchange area resource
-  enable_common_exchange_area_access: false
-  # The below configuration is a mapping from studies to their "common exchange area data"
-  # Fence project name a user gets access to when parsing c999 exchange area codes (and
-  # subsequently gives access to an Arborist resource representing this common area
-  # as well)
-  study_common_exchange_areas:
-    'example': 'test_common_exchange_area'
-    # 'studyX': 'test_common_exchange_area'
-    # 'studyY': 'test_common_exchange_area'
-    # 'studyZ': 'test_common_exchange_area'
-  # A mapping from the dbgap study / Fence project to which authorization namespaces the
-  # actual data lives in. For example, `studyX` data may exist in multiple organizations, so
-  # we need to know how to map authorization to all orgs resources
-  study_to_resource_namespaces:
-    '_default': ['/']
-    'test_common_exchange_area': ['/dbgap/']
-    # above are for default support and exchange area support
-    # below are further examples
+  - info:
+      host: ''
+      username: ''
+      password: ''
+      port: 22
+      proxy: ''
+      proxy_user: ''
+    protocol: 'sftp'
+    decrypt_key: ''
+    # parse out the consent from the dbgap accession number such that something
+    # like "phs000123.v1.p1.c2" becomes "phs000123.c2".
     #
-    # 'studyX': ['/orgA/', '/orgB/']
-    # 'studyX.c2': ['/orgB/', '/orgC/']
-    # 'studyZ': ['/orgD/']
+    # NOTE: when this is "false" the above would become "phs000123"
+    parse_consent_code: true
+    # A consent of "c999" can indicate access to that study's "exchange area data"
+    # and when a user has access to one study's exchange area data, they
+    # have access to the parent study's "common exchange area data" that is not study
+    # specific. The following config is whether or not to parse/handle "c999" codes
+    # for access to the common exchange area data
+    #
+    # NOTE: When enabled you MUST also provide a mapping to the
+    # `study_common_exchange_areas` from study -> parent common exchange area resource
+    enable_common_exchange_area_access: false
+    # The below configuration is a mapping from studies to their "common exchange area data"
+    # Fence project name a user gets access to when parsing c999 exchange area codes (and
+    # subsequently gives access to an Arborist resource representing this common area
+    # as well)
+    study_common_exchange_areas:
+      'example': 'test_common_exchange_area'
+      # 'studyX': 'test_common_exchange_area'
+      # 'studyY': 'test_common_exchange_area'
+      # 'studyZ': 'test_common_exchange_area'
+    # A mapping from the dbgap study / Fence project to which authorization namespaces the
+    # actual data lives in. For example, `studyX` data may exist in multiple organizations, so
+    # we need to know how to map authorization to all orgs resources
+    study_to_resource_namespaces:
+      '_default': ['/']
+      'test_common_exchange_area': ['/dbgap/']
+      # above are for default support and exchange area support
+      # below are further examples
+      #
+      # 'studyX': ['/orgA/', '/orgB/']
+      # 'studyX.c2': ['/orgB/', '/orgC/']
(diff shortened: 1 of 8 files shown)


fence/sync/sync_users.py
```diff
@@ -959,59 +973,111 @@ def _get_or_create(sess, model, **kwargs):
...
-                if self.protocol == "sftp":
-                    self._get_from_sftp_with_proxy(tmpdir)
-                else:
-                    self._get_from_ftp_with_proxy(tmpdir)
-                dbgap_file_list = glob.glob(os.path.join(tmpdir, "*"))
-            except Exception as e:
-                self.logger.error(e)
-                exit(1)
-
+        server = dbGaP["Info"]
```
paulineribeyre: capital letter?

fence/sync/sync_users.py
```diff
@@ -959,59 +973,111 @@ def _get_or_create(sess, model, **kwargs):
...
+        user_projects, user_info = self._get_user_permissions_from_csv_list(
+            dbGaP, dbgap_file_list, encrypted=True, session=sess
         )
         try:
             shutil.rmtree(tmpdir)
         except OSError as e:
             self.logger.info(e)
             if e.errno != errno.ENOENT:
                 raise
+        user_projects = self._parse_projects(user_projects)
```
paulineribeyre: `parse_projects` not `_parse_projects`?

fence/sync/sync_users.py
```diff
@@ -959,59 +973,111 @@ def _get_or_create(sess, model, **kwargs):
...
+        """
+        Collect files from dbgap server(s), sync csv and yaml files to storage
+        backend and fence DB
+        """
+
+        # get all dbgap files
+        user_projects = {}
+        user_info = {}
+        if self.is_sync_from_dbgap_server:
+            if self.additional_dbGaP:
```
paulineribeyre: i think the if/else is not necessary, you could do `dbgap_servers = [self.dbGaP] + self.additional_dbGaP` and `_merge_multiple_dbgap_sftp` all the time. `additional_dbGaP` just needs to be an empty list by default instead of None

fence/sync/sync_users.py
```diff
@@ -1044,6 +1106,9 @@ def _sync(self, sess):
         self.sync_two_phsids_dict(user_projects_csv, user_projects)
         self.sync_two_phsids_dict(user_yaml.projects, user_projects)
 
+        # Note: if there are multiple dbgap sftp servers configured
+        # this parameter is always from the config for the first dbgap sftp server
+        # not any additional ones
         if self.parse_consent_code:
```
paulineribeyre: this is an important note, it should also be in the config default comments

if we merge `additional_dbGaP` into `dbGaP`, i think `parse_consent_code` should be pulled out of the `dbGaP` section

fence/config-default.yaml
```diff
@@ -431,6 +431,18 @@ dbGaP:
     # 'studyX': ['/orgA/', '/orgB/']
     # 'studyX.c2': ['/orgB/', '/orgC/']
     # 'studyZ': ['/orgD/']
+# additional dbGap SFTP servers to sync with
+# Follows the same format as dbGaP entries
+# accepts a list of objects
+# example:
+# additional_dbGaP:
```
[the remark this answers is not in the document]
jiaqi216: I don't know if it'd still be backwards compatible since we'd have to go back and add `-` in front of all the existing dbGaP configs so it parses as a list.

fence/sync/sync_users.py
jiaqi216: I think I was trying to optimize for least amount of logical changes to existing dbgap syncs in case `_merge_multiple_dbgap_sftp` had any side effects that we couldn't detect for during testing.

paulineribeyre:
rn `fence-create` is not creating `UserSyncer` with the `additional_dbGaP` parameter... i'm a bit concerned about why the integration tests are not failing

fence/config-default.yaml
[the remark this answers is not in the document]
paulineribeyre: this line https://github.com/uc-cdis/fence/blob/4.14.0/bin/fence-create#L354
could be replaced by
```
dbGaP = os.environ.get("dbGaP") or config.get("dbGaP")
if not isinstance(dbGaP, list):
    dbGaP = [dbGaP]
```
it's a trade off between making the least amount of changes to the code and reducing config complexity/human error 🤷‍♀

jiaqi216:
> rn fence-create is not creating UserSyncer with the additional_dbGaP parameter... i'm a bit concerned about why the integration tests are not failing

Yep completely missed that. I think it makes sense that integration tests are not failing since we don't have any integration tests for multiple sftp support and the rest of the code should work as is without `additional_dbGaP`

fence/config-default.yaml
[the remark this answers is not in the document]
jiaqi216: Yeah, I agree that this is the better approach. I just don't feel super confident in the testing of the merging of the files from 2 sftps and like having the separation of this new feature so it has minimal impact on commons not using it. But if you are confident in the integration tests for user sync to validate that everything works as normal with one sftp, then I'm happy going this route - but will probably for sanity sake want to add more integration tests for 2 sftps which might be challenging...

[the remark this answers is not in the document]
paulineribeyre: yeah i think the integration tests are enough to make sure 1 sftp still works, but we need to test merging 2 sftps, at least manually - the first step would be to try configuring our test sftp server twice and see what happens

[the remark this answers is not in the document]
jiaqi216: Sounds good, I'll make the code changes. I'm testing 2 sftps on my dev environment.

fence/sync/sync_users.py
```diff
@@ -968,50 +1080,34 @@ def sync(self):
...
-        user_projects_csv, user_info_csv = self._parse_csv(
-            dict(list(zip(local_csv_file_list, permissions))),
+        # if syncing from local csv dir dbgap configurations
+        # come from the first dbgap instance in the fence config file
+        user_projects_csv, user_info_csv = self._get_user_permissions_from_csv_list(
+            local_csv_file_list,
             encrypted=False,
-            sess=sess,
+            session=sess,
+            dbgap_config=self.dbGaP[0],
```
jiaqi216: @paulineribeyre this is kind of awkward - not sure if we use this local csv file logic outside of testing but when fetching from local files, it will use the parameters from dbGaP[0] now that dbGaP is a list.

paulineribeyre: yeah it's fine, it's just for testing, but maybe we can add it as a note for the `--csv_dir` argument in the fence-create script?

paulineribeyre:
```suggestion
            dbgap_config=self.dbGaP[0] if self.dbGaP else {},
```

fence/sync/sync_users.py
```diff
@@ -293,15 +293,8 @@ def __init__(
...
-            self.server = dbGaP["info"]
-            self.protocol = dbGaP["protocol"]
-            self.dbgap_key = dbGaP["decrypt_key"]
-        self.parse_consent_code = dbGaP.get("parse_consent_code", True)
-        self.enable_common_exchange_area_access = dbGaP.get(
-            "enable_common_exchange_area_access", False
-        )
-        self.study_common_exchange_areas = dbGaP.get("study_common_exchange_areas", {})
+        self.dbGaP = dbGaP
+        self.parse_consent_code = dbGaP[0].get("parse_consent_code", True)
```
paulineribeyre:
```suggestion
        self.parse_consent_code = dbGaP[0].get("parse_consent_code", True) if dbGaP else False
```

tests/test-fence-config.yaml
```diff
@@ -276,71 +276,78 @@ ITRUST_GLOBAL_LOGOUT: 'https://auth.nih.gov/siteminderagent/smlogout.asp?mode=ni
...
+      port: 22
+      proxy: ''
+      proxy_user: ''
+    protocol: 'sftp'
+    decrypt_key: ''
+    # parse out the consent from the dbgap accession number such that something
+    # like "phs000123.v1.p1.c2" becomes "phs000123.c2".
+    #
+    # NOTE: when this is "false" the above would become "phs000123"
+    parse_consent_code: true
```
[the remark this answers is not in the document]
jiaqi216: It's not ignored, It's still used everywhere in this function https://github.com/uc-cdis/fence/blob/feat/sync/fence/sync/sync_users.py#L397 which is called once per dbgap sftp config.

But specifically on this [line](https://github.com/uc-cdis/fence/blob/feat/sync/fence/sync/sync_users.py#L1142) (which happens after all the dbgap files from all dbgap sftps are merged together), it's only using the parse_consent_code from the first dbgap config.

fence/sync/sync_users.py
```diff
@@ -1023,13 +1119,9 @@ def _sync(self, sess):
             self.logger.error("aborting early")
             return
 
-        user_projects_csv = {
-            key.lower(): value for key, value in user_projects_csv.items()
-        }
-        user_projects = {key.lower(): value for key, value in user_projects.items()}
```
paulineribeyre: i think we need this line

PR #759, opened 2020-03-02
chore(config): don't leave fake s3 buckets and creds as the default c…

Avantol13:
…onfig

### Improvements
- default config no longer gives fake buckets and creds for s3 (avoids an s3 call to get missing bucket region and reduces confusion)

### DIFF
--- a/fence/__init__.py
+++ b/fence/__init__.py
@@ -174,8 +174,8 @@ def _check_s3_buckets(app):
     Function to ensure that all s3_buckets have a valid credential. 
     Additionally, if there is no region it will produce a warning then trys to fetch and cache the region. 
     """
-    buckets = config.get("S3_BUCKETS", {})
-    aws_creds = config.get("AWS_CREDENTIALS", {})
+    buckets = config.get("S3_BUCKETS") or {}
+    aws_creds = config.get("AWS_CREDENTIALS") or {}
 
     for bucket_name, bucket_details in buckets.items():
         cred = bucket_details.get("cred")
--- a/fence/config-default.yaml
+++ b/fence/config-default.yaml
@@ -449,8 +449,8 @@ DBGAP_ACCESSION_WITH_CONSENT_REGEX: '(?P<phsid>phs[0-9]+)(.(?P<version>v[0-9]+))
 #               access for users.
 # //////////////////////////////////////////////////////////////////////////////////////
 # Configuration for various storage systems for the backend
-# NOTE: Remove the [] and supply backends if needed. Example in comments below
-STORAGE_CREDENTIALS: []
+# NOTE: Remove the {} and supply backends if needed. Example in comments below
+STORAGE_CREDENTIALS: {}
 # Google Cloud Storage backend
 #
 #  'google':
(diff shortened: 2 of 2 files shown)


fence/config-default.yaml
```diff
@@ -476,30 +476,32 @@ STORAGE_CREDENTIALS: []
...
+# NOTE: Remove the [] and supply creds if needed. Example in comments below
+#   'CRED1':
+#     aws_access_key_id: ''
+#     aws_secret_access_key: ''
+#   'CRED2':
+#     aws_access_key_id: ''
+#     aws_secret_access_key: ''
 
 # NOTE: the region is optonal for s3_buckets, however it should be specified to avoid a
 # call to GetBucketLocation which you make lack the AWS ACLs for.
```
vpsx: Does this comment still apply? (Do we still recommend supplying a region as the way to avoid this GetBucketLocation call?)

Avantol13: yeah, it still applies

fence/__init__.py
```diff
@@ -174,8 +174,8 @@ def _check_s3_buckets(app):
     Function to ensure that all s3_buckets have a valid credential. 
     Additionally, if there is no region it will produce a warning then trys to fetch and cache the region. 
     """
-    buckets = config.get("S3_BUCKETS", {})
-    aws_creds = config.get("AWS_CREDENTIALS", {})
+    buckets = config.get("S3_BUCKETS") or {}
```
jiaqi216: any particular reason why you prefer this syntax? Functionally, it's the same right?

vpsx: I think because the config default is [] not {}, so this way you always get {}, since [] is.. false-y. Maybe to make it more explicit we can just do it the dumb way: do the config.get and then `if thing is None or thing is [] thing = {}`

fence/config-default.yaml
```diff
@@ -476,30 +476,32 @@ STORAGE_CREDENTIALS: []
...
 #   - Support `/data` endpoints
 # //////////////////////////////////////////////////////////////////////////////////////
-AWS_CREDENTIALS:
-  'CRED1':
-    aws_access_key_id: ''
-    aws_secret_access_key: ''
-  'CRED2':
-    aws_access_key_id: ''
-    aws_secret_access_key: ''
+AWS_CREDENTIALS: []
```
paulineribeyre: why `[]` and not `{}`?

fence/__init__.py
Avantol13: yeah, it's to allow `[]` or `null` but not fail on the line where we try to do `.items()` below

fence/config-default.yaml
Avantol13: I can do that too, the weird thing is that all the other dicts here also have `[]` as a default (for example, see `STORAGE_CREDENTIALS`). so we're handling it like this elsewhere, I thought I'd keep it consistent.

fence/__init__.py
Avantol13: so it is functionally different. Before, it would allow `null` or `[]` and raise an exception on line 180

fence/config-default.yaml
paulineribeyre: it doesn't really make sense to me, sounds like other dicts being initialized to `[]` is a mistake, not a decision?

paulineribeyre: ah i see you already updated it 🎉 🎉

PR #772, opened 2020-04-24
PXP-5867 Update OIDC endpoints in swagger

jiaqi216:
Updated Swagger Docs for `/user` endpoint and added information about `/.well-known/openid-configuration`

Would probably be easier to review / look at in swagger http://petstore.swagger.io/?url=https://raw.githubusercontent.com/uc-cdis/fence/chore/update-swagger/openapis/swagger.yaml

I didn't include too much detail on the fields in either of these endpoints as they are pretty standard in OIDC and users should refer to the OIDC spec for additional information.

### Breaking Changes
None

### Bug Fixes
Fix docs

### Improvements
Updated documentation to reflect Fence API behavior

### DIFF
--- a/openapis/swagger.yaml
+++ b/openapis/swagger.yaml
@@ -6,7 +6,7 @@ info:
   title: Fence OpenAPI Specification
   version: 0.1.0
   description: >-
-    Access management for CDIS data commons. Code is available on
+    Access management for Gen3 data commons. Code is available on
     [GitHub](https://github.com/uc-cdis/fence).
   termsOfService: 'http://cdis.uchicago.edu/terms/'
   contact:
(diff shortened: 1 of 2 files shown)


openapis/swagger.yaml
```diff
@@ -1975,40 +1996,85 @@ components:
...
+        is_admin:
+          type: boolean
+          description: 'Boolean value stating if end-user is an admin or not'
+        role:
           type: string
           description: ''
-        resources_granted:
+        project_access:
+          type: object
+          description: ''
```
paulineribeyre, line 2047:
```suggestion
          description: 'This value is deprecated in favor of authz and resources.'
```
or `list of projects the user has access to` if we don't want to say it's deprecated

openapis/swagger.yaml
```diff
@@ -1975,40 +1996,85 @@ components:
...
+        authz:
+          type: object
+          description: 'end user authorization information'
+        resources:
+          type: array
+          items:
+            type: string
+          description: 'list of resource paths the user has access to'
+        tags:
+          type: object
```
paulineribeyre, line 2074:
```suggestion
        tags:
          type: object
          description: User's tags
```
we actually use this during dbgap whitelists usersync - alex can probably provide a better description if needed

Avantol13: it stores user's PI and some other metadata

PR #774, opened 2020-04-29
add assume_role cache

fantix:
The result from AWS call `assume_role` is now cached both in database and in-memory by ARN with a TTL that is by default 30 minutes longer than the requested expiry.

### New Features
- Added `assume_role` cache for download/upload


fence/blueprints/data/indexd.py
```diff
@@ -500,16 +502,46 @@ def assume_role(cls, bucket_cred, expires_in, aws_creds_config, boto=None):
...
-        assumed_role = boto.assume_role(role_arn, expires_in, aws_creds_config)
+        expiry = time.time() + expires_in
+
+        # try to retrieve from local in-memory cache
+        rv, expires_at = cls._assume_role_cache.get(role_arn, (None, 0))
+        if expires_at > expiry:
+            return rv
+
+        # try to retrieve from database cache
+        from ...models import AssumeRoleCache
```
Avantol13, line 516: I'd prefer we have all the imports at the top, even if it's only used here, just for consistency

fence/blueprints/data/indexd.py
```diff
@@ -500,16 +502,46 @@ def assume_role(cls, bucket_cred, expires_in, aws_creds_config, boto=None):
...
+                )
+                cls._assume_role_cache[role_arn] = rv, cache.expires_at
+                return rv
+
+        # retrieve from AWS, with additional 30 minutes buffer for cache
+        boto = boto or flask.current_app.boto
+        assumed_role = boto.assume_role(
+            role_arn,
+            expires_in
+            + int(flask.current_app.config.get("ASSUME_ROLE_CACHE_SECONDS", 1800)),
```
Avantol13, line 538: instead of a config default here, could you just add this to the config-default.yaml with 1800 seconds and then you can safely from config import config, config["ASSUME_ROLE_CACHE_SECONDS"]?

tests/data/test_data.py
```diff
@@ -608,6 +612,114 @@ def test_indexd_download_with_uploader_authorized(
...
+    google_signed_url,
+):
+    """
+    Test ``GET /data/download/1`` with authorized user (user is the uploader).
+    """
+
+    assume_role_called = 0
+
+    def mock_assume_role(self, role_arn, duration_seconds, config=None):
+        nonlocal assume_role_called
```
Avantol13, line 633: `nonlocal` :sunglasses:

tests/data/test_data.py
```diff
@@ -608,6 +612,114 @@ def test_indexd_download_with_uploader_authorized(
...
+        "hashes": {},
+        "metadata": {},
+        "form": "",
+        "created_date": "",
+        "updated_date": "",
+    }
+    mock_index_document = mock.patch(
+        "fence.blueprints.data.indexd.IndexedFile.index_document", index_document
+    )
+    mock_index_document.start()
```
Avantol13, line 669: I think you need to `.stop()` this later for cleanup

fantix:
I know why the Jenkins keeps failing now.
* AWS has a limit on the maximum time limit ([`MaxSessionDuration`](https://docs.aws.amazon.com/IAM/latest/APIReference/API_UpdateRole.html)) for assume-role access tokens, it is 1 hour by default. Requesting access tokens with TTL greater than this limit will fail.
* In Fence, we have a config `MAX_PRESIGNED_URL_TTL` whose default value is also 1 hour. If a user requests a pre-signed URL with longer TTLs, it will be capped. 
* Before this PR, the two 1-hours match. Every pre-signed URL will use a different access token with the same TTL. Everything works.
* After this PR, different pre-signed URLs will share some access tokens. Thus, the access tokens must have larger TTL than the requested pre-signed URLs. The difference is the new config `ASSUME_ROLE_CACHE_SECONDS`, default is 30 minutes.
* That means, we're requesting for assume-role tokens with TTL 1.5 hours, exceeding the limit. Increasing the AWS account limit `MaxSessionDuration` to 5,400 seconds will fix the issue. That is to say, we need to change this limit to `MAX_PRESIGNED_URL_TTL + ASSUME_ROLE_CACHE_SECONDS` for any relevant AWS accounts. If it's not possible to do so, we'll have to sacrifice some seconds from `MAX_PRESIGNED_URL_TTL` to `ASSUME_ROLE_CACHE_SECONDS` for cache.

fantix:
... unless the pre-signed URL could over-live the assume-role access token?? I didn't find any AWS docs for this, but I think we could always give it a try.

fence/config-default.yaml
```diff
@@ -766,6 +766,7 @@ SYNAPSE_URI: 'https://repo-prod.prod.sagebase.org/auth/v1'
 SYNAPSE_JWKS_URI:
 SYNAPSE_DISCOVERY_URL:
 SYNAPSE_AUTHZ_TTL: 86400
+ASSUME_ROLE_CACHE_SECONDS: 1800
```
themarcelor, line 769: Looks like this value here is not valid 🤔 
The pod is failing due to:
> The requested DurationSeconds exceeds the MaxSessionDuration set for this role.

cc: @Avantol13

themarcelor: We might need @uc-cdis/planx-devops to help us figure this out:
https://docs.aws.amazon.com/cli/latest/reference/iam/update-role.html

> --max-session-duration (integer)
> The maximum session duration (in seconds) that you want to set for the specified role. If you do not specify a value for this setting, the default maximum of one hour is applied. This setting can have a value from 1 hour to 12 hours.

fence/config-default.yaml
```diff
@@ -807,6 +807,12 @@ SYNAPSE_URI: 'https://repo-prod.prod.sagebase.org/auth/v1'
 SYNAPSE_JWKS_URI:
 SYNAPSE_DISCOVERY_URL:
 SYNAPSE_AUTHZ_TTL: 86400
+
+# Role caching for generating presigned urls if max role session increase is true
+# then we can increase the amount of time that a session is valid for
+MAX_ROLE_SESSION_INCREASE: true
```
Avantol13, line 813: I don't think we want all this enabled by default. This will break in places where role owner hasn't increased on the AWS side

fence/__init__.py
```diff
@@ -294,7 +294,8 @@ def app_config(
 
     _setup_oidc_clients(app)
 
-    _check_s3_buckets(app)
+    with app.app_context():
```
Avantol13, line 297: what's the reasoning behind this?

Avantol13 (changes requested):

fence/__init__.py
MichaelLukowski: Not sure about this. I think this was part of fantix's changes we can revert this

fence/blueprints/data/indexd.py
```diff
@@ -942,27 +1010,90 @@ def _generate_google_storage_signed_url(
...
+            db_entry["gcp_proxy_group_id"] = proxy_group_id
+            db_entry["gcp_private_key"] = str(private_key)
+            db_entry["gcp_key_db_entry"] = str(key_db_entry)
+            db_entry["expires_at"] = expiration_time
+
+            if hasattr(flask.current_app, "db"):  # we don't have db in startup
+                with flask.current_app.db.session as session:
+                    session.execute(
+                        """\
+                        INSERT INTO gcp_assume_role_cache (
```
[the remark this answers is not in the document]
MichaelLukowski: There is no mechanism for clearing right now but when the creds are expired we generate new creds. Then we update the table with the new creds on the same `proxy_group_id` which essentially rotates the creds. Do we still need a way to remove the entry from the table if it's expired?

fence/models.py
```diff
@@ -532,6 +532,25 @@ class ServiceAccountToGoogleBucketAccessGroup(Base):
     )
 
 
+class AssumeRoleCache(Base):
```
paulineribeyre, line 535: how about `AssumeRoleCacheAWS` for clarity and to be consistent with `AssumeRoleCacheGCP`?

fence/config-default.yaml
MichaelLukowski: agreed, I think I enabled it because we were having flakey unit tests when it was false. But we can just keep it enabled in the test config and disable it in the default.

fence/blueprints/data/indexd.py
```diff
@@ -597,6 +599,8 @@ class S3IndexedFileLocation(IndexedFileLocation):
     An indexed file that lives in an AWS S3 bucket.
     """
 
+    _assume_role_cache = {}
+
     @classmethod
     def assume_role(cls, bucket_cred, expires_in, aws_creds_config, boto=None):
         """
```
paulineribeyre, line 606: maybe this docstring could use an update with details about the new logic

fence/blueprints/data/indexd.py
```diff
@@ -608,16 +612,53 @@ def assume_role(cls, bucket_cred, expires_in, aws_creds_config, boto=None):
...
+
+        # try to retrieve from database cache
+        if hasattr(flask.current_app, "db"):  # we don't have db in startup
+            with flask.current_app.db.session as session:
+                cache = (
+                    session.query(AssumeRoleCache)
+                    .filter(AssumeRoleCache.arn == role_arn)
+                    .first()
+                )
+                if cache and cache.expires_at and cache.expires_at > expiry:
```
paulineribeyre, line 633: in which case would `cache.expires_at` not be populated?

fence/blueprints/data/indexd.py
```diff
@@ -597,6 +599,8 @@ class S3IndexedFileLocation(IndexedFileLocation):
     An indexed file that lives in an AWS S3 bucket.
     """
 
+    _assume_role_cache = {}
```
paulineribeyre, line 602: i'd comment the expected structure (`{ role_arn: (rv, expires_at) }`)

fence/blueprints/data/indexd.py
```diff
@@ -864,6 +935,8 @@ class GoogleStorageIndexedFileLocation(IndexedFileLocation):
     An indexed file that lives in a Google Storage bucket.
     """
 
+    _assume_role_cache_gs = {}
```
paulineribeyre, line 938: i'd comment the expected structure (`{ proxy_group_id: (private_key, key_db_entry) }`)

fence/blueprints/data/indexd.py
```diff
@@ -947,27 +1020,86 @@ def _generate_google_storage_signed_url(
...
+            private_key, key_db_entry = self._assume_role_cache_gs.get(proxy_group_id)
+            is_cached = True
+        elif hasattr(flask.current_app, "db"):
+            with flask.current_app.db.session as session:
+                cache = (
+                    session.query(AssumeRoleCacheGCP)
+                    .filter(AssumeRoleCacheGCP.gcp_proxy_group_id == proxy_group_id)
+                    .first()
+                )
+                if cache and cache.expires_at and cache.expires_at > expiration_time:
```
paulineribeyre, line 1038: same question about `cache.expires_at`

fence/blueprints/data/indexd.py
```diff
@@ -947,27 +1020,86 @@ def _generate_google_storage_signed_url(
...
+                    )
+                    self._assume_role_cache_gs[proxy_group_id] = rv
+            if (
+                proxy_group_id in self._assume_role_cache_gs
+                and self._assume_role_cache_gs[proxy_group_id] != None
+            ):
+                private_key, key_db_entry = self._assume_role_cache_gs.get(
+                    proxy_group_id
+                )
+                is_cached = True
```
paulineribeyre, line 1051: why are we doing this twice?

fence/models.py
MichaelLukowski: agree, updated to this.

fence/blueprints/data/indexd.py
MichaelLukowski: added

fence/blueprints/data/indexd.py
MichaelLukowski: I think you're right, there probably isn't a time where it would not be populated

fence/blueprints/data/indexd.py
MichaelLukowski: added

fence/blueprints/data/indexd.py
MichaelLukowski: added

fence/blueprints/data/indexd.py
MichaelLukowski: updated

fence/blueprints/data/indexd.py
paulineribeyre: you removed the check for GCP but not for AWS 🤷‍♀️

paulineribeyre:
lgtm

fence/blueprints/data/indexd.py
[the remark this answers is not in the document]
Avantol13: ahh, I see the on conflict stuff. We just are overwriting what's there. okay this is fine then, I just wanted to make sure we weren't creating a new entry every time one expires

Avantol13 (approved):

PR #776, opened 2020-05-05
Feat/delete flow authz

ZakirG:
For the purpose of https://ctds-planx.atlassian.net/browse/PXP-5435

Partner PR: https://github.com/uc-cdis/cirrus/pull/81

### New Features
- DELETE /user/file_id endpoint now uses authz field instead of uploader field to authenticate user


ZakirG:
wool

fence/blueprints/data/blueprint.py
```diff
@@ -36,15 +37,63 @@ def delete_data_file(file_id):
...
+
+    authz = record.index_document.get("authz")
+    arborist_check = None
+    if authz:
+        arborist_check = flask.current_app.arborist.auth_request(
+            jwt=get_jwt(), service="fence", methods="delete", resources=authz
+        )
+
+    # If authz is not empty, use *only* arborist to check if user can delete
+    # (don't fall back on uploader, this stops that security issue from happening).
```
Avantol13, line 49: I would be more specific here about the case it's preventing. like "this prevents users from escalating from edit to delete permissions by changing the uploader field to their own username (b/c users only have edit access through arborist/authz)"

fence/blueprints/data/blueprint.py
```diff
@@ -36,15 +37,63 @@ def delete_data_file(file_id):
...
+    if authz and arborist_check:
+        logger.info("deleting record and files for {}".format(file_id))
+        try:
+            record.delete_files(delete_all=True)
+        except Exception as exc:
+            return (
```
Avantol13, line 55: should log the exception too and traceback too with something like `logging.error(exc, exc_info=True)`

fence/blueprints/data/blueprint.py
```diff
@@ -36,15 +37,63 @@ def delete_data_file(file_id):
         file_id (str): GUID of file to delete
     """
     record = IndexedFile(file_id)
-    # check auth: user must have uploaded the file (so `uploader` field on the record is
-    # this user)
+
+    authz = record.index_document.get("authz")
+    arborist_check = None
+    if authz:
+        arborist_check = flask.current_app.arborist.auth_request(
```
Avantol13, line 44: I would add a debug log here, something like "trying to ask arborist if user can delete in fence for {authz}"

fence/blueprints/data/blueprint.py
```diff
@@ -36,15 +37,63 @@ def delete_data_file(file_id):
         file_id (str): GUID of file to delete
     """
     record = IndexedFile(file_id)
-    # check auth: user must have uploaded the file (so `uploader` field on the record is
-    # this user)
+
+    authz = record.index_document.get("authz")
+    arborist_check = None
```
Avantol13, line 42: nitpick: I think a name like "has_authz" would be more clear. at least to me. For booleans (or booleans that can be None too :P ) I try to use names that start with "has" or "is" so it's clear it's a boolean. anyway, mostly a personal preference, not a huge deal

fence/blueprints/data/blueprint.py
```diff
@@ -36,15 +37,63 @@ def delete_data_file(file_id):
...
+    if authz:
+        arborist_check = flask.current_app.arborist.auth_request(
+            jwt=get_jwt(), service="fence", methods="delete", resources=authz
+        )
+
+    # If authz is not empty, use *only* arborist to check if user can delete
+    # (don't fall back on uploader, this stops that security issue from happening).
+    if authz and arborist_check:
```
Avantol13, line 50: opinion: this and the next block (the case where the check failed) could be nested in the first check above for tiny efficiency boost when authz doesn't exist (b/c you won't need these extra conditional checks). So instead you end up with something like:

```
if authz:
  has_authz = do_arborist_check()
  if has_authz:
    # do the thing (delete)
  else:
    # 403

# do uploader logic
```

fence/blueprints/data/indexd.py
```diff
@@ -437,6 +437,8 @@ def delete_files(self, urls=None, delete_all=True):
         for location in locations_to_delete:
             bucket = location.bucket_name()
             flask.current_app.boto.delete_data_file(bucket, self.file_id)
+            with GoogleCloudManager() as gcm:
+                gcm.delete_data_file(bucket, self.file_id)
```
Avantol13, line 441: you might want some error handling and logging around this

fence/blueprints/data/blueprint.py
ZakirG: i agree ~ this variable is not a boolean though, it's gonna contain the list of resource path strings

ZakirG: oh youre right okay

fence/blueprints/data/indexd.py
ZakirG: I wrapped the delete_files call as a whole inside a try / except in the place where its called in blueprint.py. and this function has its own try / except inside of cirrus as well. do you think putting it here is good too?

fence/blueprints/data/blueprint.py
ZakirG: hmm i feel like the separated-out version is more readable to me -- what do u think? i think the efficiency boost in question is quite small

ZakirG: also, the zen of python says "Flat is better than nested." https://www.python.org/dev/peps/pep-0020/

fence/blueprints/data/blueprint.py
```diff
@@ -36,15 +37,70 @@ def delete_data_file(file_id):
...
+        )
+        has_correct_authz = flask.current_app.arborist.auth_request(
+            jwt=get_jwt(), service="fence", methods="delete", resources=authz
+        )
+
+    # If authz is not empty, use *only* arborist to check if user can delete
+    # Don't fall back on uploader -- this prevents users from escalating from edit to
+    # delete permissions by changing the uploader field to their own username
+    # (b/c users only have edit access through arborist/authz)
+    if authz and has_correct_authz:
```
Avantol13, line 55: I still think you should indent this into the authz block along with the 403 check :sweat_smile:

fence/blueprints/data/indexd.py
Avantol13: oh I missed that, this is cool then

fence/blueprints/data/indexd.py
```diff
@@ -437,7 +437,26 @@ def delete_files(self, urls=None, delete_all=True):
             ]
         for location in locations_to_delete:
             bucket = location.bucket_name()
-            flask.current_app.boto.delete_data_file(bucket, self.file_id)
+
+            if isinstance(location, GoogleStorageIndexedFileLocation):
```
Avantol13, line 441: I would actually rather add a "delete" function to classes so you don't have to do this isinstance stuff here. The "IndexedFileLocation" class should have a delete function that raises notImplemented and then the child classes for Google and AWS can implement the functionality you have here, I think  that'd be a cleaner overall solution to add that feature into the classes. and then here, all you have to do is call "location.delete()"

fence/blueprints/data/indexd.py
```diff
@@ -801,6 +821,28 @@ def _generate_anonymous_google_storage_signed_url(
         )
         return final_url
 
+    def bucket_name(self):
```
Avantol13, line 824: I think this logic is the same for AWS? It would nice to actually have this functionality in the parent class `IndexedFileLocation` so you can use these for google or aws. also, fun python thing, you can use a decorator `@property` so these behave like attributes so you can just do `location.bucket_name` instead of `location.bucket_name()`. it's a small thing, but I think it's actually clearer that way. If you don't want to do that, I'd rename these functions `get_bucket_name()`

ZakirG: the logic is different between the functions. the s3 functions have their own config and parsing that someone else wrote. bucket_name is the pre-existing naming format for this function, it's used throughout the S3 code that i didn't write. are you sure you want me to do a find-and-replace that affects a large amount of s3 code?

fence/blueprints/data/indexd.py
```diff
@@ -845,6 +858,28 @@ def _generate_anonymous_google_storage_signed_url(
...
+            logger.error("Unable to get bucket name from resource path. {}".format(exc))
+
+        return bucket_name
+
+    def file_name(self):
+        resource_path = self.get_resource_path()
+
+        file_name = None
+        try:
+            file_name = resource_path.split("/")[1]
```
Avantol13, line 877: I actually think this needs to be 
```suggestion
            file_name = resource_path.split("/")[-1]
```

B/c this won't work if the file is within a couple folders like: "gs://bucket-name/folder1/folder2/file.txt"

fence/blueprints/data/indexd.py
```diff
@@ -481,7 +481,16 @@ def delete_files(self, urls=None, delete_all=True):
             ]
         for location in locations_to_delete:
             bucket = location.bucket_name()
-            flask.current_app.boto.delete_data_file(bucket, self.file_id)
+            file_suffix = self.file_id
+            if not file_suffix:
```
Avantol13, line 485: I think we need to just always get the file_name, b/c file_id is just a GUID and later on we try to delete based on that name in the bucket which causes errors

fence/blueprints/data/blueprint.py
```diff
@@ -11,6 +11,7 @@
 from fence.errors import Forbidden, InternalError, UserError, Forbidden
 from fence.utils import is_valid_expiration
 from fence.authz.auth import check_arborist_auth
+from fence.auth import get_jwt
```
Avantol13, line 14: this is already imported in line 5

fence/blueprints/data/indexd.py
```diff
@@ -898,6 +931,16 @@ def _generate_google_storage_signed_url(
...
 
+    def delete(self, bucket, file_id):
+        """
+        Delete data file in Google Cloud Storage. Note that file_id is unused to satisfy
+        polymorphism -- the s3 subclass needs the file_id argument.
+        """
+        with GoogleCloudManager(
+            creds=config["CIRRUS_CFG"]["GOOGLE_STORAGE_CREDS"]
+        ) as gcm:
+            gcm.delete_data_file(bucket, self.file_name())
```
Avantol13, line 942:
```suggestion
            gcm.delete_data_file(bucket, file_id)
```

I think file_id is already the filename here?

fence/blueprints/data/indexd.py
```diff
@@ -587,6 +601,10 @@ def bucket_name(self):
                 return bucket
         return None
 
+    def file_name(self):
+        file_name = "/".join(self.url.split("/")[3:])
```
Avantol13, line 605: will this always work, no matter the URL / number of folders? Wondering if you should use something like `self.parsed_url.path.strip("/")` and strip of the bucket name from the front

fence/blueprints/data/indexd.py
```diff
@@ -841,6 +867,30 @@ def _generate_anonymous_google_storage_signed_url(
...
+            logger.error("Unable to get bucket name from resource path. {}".format(exc))
+
+        return bucket_name
+
+    def file_name(self):
+        resource_path = self.get_resource_path()
+
+        file_name = None
+        try:
+            # GCS Filenames must be encoded according to https://tools.ietf.org/html/rfc3986#section-3.3
```
Avantol13, line 886: I feel like this requirement should be encapsulated in cirrus, rather than fence. So Fence can just pass the filename and cirrus handles the strange encoding requirement

fence/resources/aws/boto_manager.py
```diff
@@ -33,24 +33,30 @@ def delete_data_file(self, bucket, guid):
             s3_objects = self.s3_client.list_objects_v2(
                 Bucket=bucket, Prefix=guid, Delimiter="/"
             )
+
             if not s3_objects.get("Contents"):
                 # file not found in the bucket
                 self.logger.info(
-                    "tried to delete GUID {} but didn't find in bucket {}".format(
+                    "tried to delete prefix {} but didn't find in bucket {}".format(
```
Avantol13, line 40: Why the change from GUID to prefix?

ZakirG: In that instance, the value is now the filepath (a.k.a. prefix) of the thing being deleted, not its GUID. originally, the value was indeed the GUID -- but the s3 code was failing silently and nothing was ever being deleted. i fixed it =D

fence/blueprints/data/indexd.py
ZakirG: good call, i altered the code there to use parsed_url

Avantol13 (approved):

PR #777, opened 2020-05-06
PXP-2327 update print to logger in fence-create

mjmartinson:
Change print statements in `fence_create.py` to use the logger.

### Improvements
- Use logger instead of print statements in fence-create

### DIFF
--- a/fence/scripting/fence_create.py
+++ b/fence/scripting/fence_create.py
@@ -3,7 +3,6 @@ import os.path
 import time
 from yaml import safe_load
 import json
-import pprint
 
 from cirrus import GoogleCloudManager
 from cirrus.google_cloud.errors import GoogleAuthError
@@ -315,7 +314,6 @@ def create_project(s, project_data):
                     .filter(CloudProvider.name == provider)
                     .first()
                 )
-                logger.info(b)
                 if not b:
                     b = Bucket(name=bucket)
                     b.provider = c_provider
--- a/fence/scripting/fence_create.py
+++ b/fence/scripting/fence_create.py
@@ -60,9 +60,9 @@ def list_client_action(db):
         driver = SQLAlchemyDriver(db)
         with driver.session as s:
             for row in s.query(Client).all():
-                pprint.pprint(row.__dict__)
+                logger.info(row.__dict__)
     except Exception as e:
-        print(str(e))
+        logger.error(str(e))
 
 
 def modify_client_action(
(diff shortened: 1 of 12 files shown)


fence/scripting/fence_create.py
```diff
@@ -60,9 +60,9 @@ def list_client_action(db):
         driver = SQLAlchemyDriver(db)
         with driver.session as s:
             for row in s.query(Client).all():
-                pprint.pprint(row.__dict__)
+                logger.info(row.__dict__)
```
paulineribeyre, line 63: remove the pprint import in this file

fence/scripting/fence_create.py
```diff
@@ -315,12 +315,12 @@ def create_project(s, project_data):
                     .filter(CloudProvider.name == provider)
                     .first()
                 )
-                print(b)
+                logger.info(b)
```
paulineribeyre, line 318: this looks like a forgotten debug log, you can just remove it

fence/scripting/fence_create.py
```diff
@@ -109,13 +108,13 @@ def create_client_action(
     DB, username=None, client=None, urls=None, auto_approve=False, **kwargs
 ):
     try:
-        print(
+        logger.info(
```
paulineribeyre, line 111: This line is actually the result the admin sees when creating an oidc client, it's the client ID and secret key. The integration tests are parsing it to use the client, that's why the tests have been failing on this PR - I think it's fine to change it back to a print statement, it's easier to read for the user

paulineribeyre: Thanks @haraprasadj

fence/scripting/fence_create.py
```diff
@@ -1251,7 +1249,9 @@ def link_external_bucket(db, name):
             current_session, name, bucket_db_entry.id, google_project_id, privileges
         )
 
-    pprint.pprint("bucket access group email: {}".format(access_group.email))
+    logger.info(
+        json.dumps("bucket access group email: {}".format(access_group.email), indent=2)
```
paulineribeyre, line 1253: json.dumps a string doesn't actually make sense, not sure why it was pretty-printed before, let's just log it

fence/scripting/fence_create.py
```diff
@@ -60,9 +59,9 @@ def list_client_action(db):
         driver = SQLAlchemyDriver(db)
         with driver.session as s:
             for row in s.query(Client).all():
-                pprint.pprint(row.__dict__)
+                logger.info(row.__dict__)
```
paulineribeyre, line 62: The other issue made me realize the print statements that return the information the user asked for (like here, listing the clients) should be easy to parse and we should probably not use the logger... So let's revert this here, and I guess the unit test changes too, sorry about that 🤔

fence/scripting/fence_create.py
```diff
@@ -1381,8 +1381,8 @@ def google_list_authz_groups(db):
             )
         ).all()
 
-        print("GoogleBucketAccessGroup.email, Bucket.name, Project.auth_id")
+        logger.info("GoogleBucketAccessGroup.email, Bucket.name, Project.auth_id")
         for item in google_authz:
-            print(", ".join(item[:-1]))
+            logger.info(", ".join(item[:-1]))
```
paulineribeyre, line 1386: Same for the 2 logs here

tests/scripting/test_fence-create.py
```diff
@@ -1,5 +1,6 @@
 import time
 import mock
+import logging
```
paulineribeyre, line 3: remove unused import

PR #780, opened 2020-05-27
PXP-4787 Enable allowed scopes to be specified in fence-create’s client-create command

johnfrancismccann:
Enable an admin running `fence-create client-create` to specify the allowed scopes when creating a client. Previously a client’s allowed scopes were initialized based on `fence.config.config[“CLIENT_ALLOWED_SCOPES”]`. With this PR, an admin can be more restrictive with what scopes are allowed for a client.

### New Features
- Enabled `fence-create client-create` to accept space-separated `allowed-scopes` arguments

### Improvements
- Added tests for `fence.scripting.fence_create.create_client_action`

### DIFF
--- a/fence/utils.py
+++ b/fence/utils.py
@@ -9,6 +9,7 @@ import requests
 from urllib.parse import urlencode
 from urllib.parse import parse_qs, urlsplit, urlunsplit
 
+from cdislogging import get_logger
 import flask
 from userdatamodel.driver import SQLAlchemyDriver
 from werkzeug.datastructures import ImmutableMultiDict
@@ -20,6 +21,7 @@ from fence.config import config
 
 rng = SystemRandom()
 alphanumeric = string.ascii_uppercase + string.ascii_lowercase + string.digits
+logger = get_logger(__name__)
 
 
 def random_str(length):
@@ -65,7 +67,8 @@ def create_client(
             )
         )
     if "openid" not in allowed_scopes:
-        raise ValueError("One of the allowed scopes must be openid")
+        allowed_scopes.append("openid")
+        logger.warning('Adding required "openid" scope to list of allowed scopes.')
     with driver.session as s:
         user = query_for_user(session=s, username=username)
 
--- a/fence/utils.py
+++ b/fence/utils.py
@@ -42,6 +42,7 @@ def create_client(
     confidential=True,
     arborist=None,
     policies=None,
+    allowed_scopes=None
 ):
     client_id = random_str(40)
     if arborist is not None:
@@ -56,6 +57,11 @@ def create_client(
             client_secret.encode("utf-8"), bcrypt.gensalt()
         ).decode("utf-8")
     auth_method = "client_secret_basic" if confidential else "none"
+    allowed_scopes = allowed_scopes or config["CLIENT_ALLOWED_SCOPES"]
+    if not set(allowed_scopes).issubset(set(config["CLIENT_ALLOWED_SCOPES"])):
+        raise ValueError("Each allowed scope must be one of: {}".format(config["CLIENT_ALLOWED_SCOPES"]))
+    if "openid" not in allowed_scopes:
+        raise ValueError("One of the allowed scopes must be openid")
     with driver.session as s:
         user = query_for_user(session=s, username=username)
 
@@ -71,7 +77,7 @@ def create_client(
             client_secret=hashed_secret,
             user=user,
             redirect_uris=urls,
-            _allowed_scopes=" ".join(config["CLIENT_ALLOWED_SCOPES"]),
+            allowed_scopes=allowed_scopes,
             description=description,
             name=name,
             auto_approve=auto_approve,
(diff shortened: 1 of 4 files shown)


tests/scripting/test_fence-create.py
```diff
@@ -151,8 +206,8 @@ def test_delete_user_with_access_privilege(app, db_session):
     db_session.add(access_privilege)
     db_session.commit()
     delete_users(config["DB"], [user.username])
-    remaining_usernames = db_session.query(User.username).all()
-    assert db_session.query(User).count() == 0, remaining_usernames
+    remaining_usernames = db_session.query(User).filter_by(username=user.username).all()
+    assert len(remaining_usernames) == 0, remaining_usernames
```
johnfrancismccann, line 210: I made this change because in the new client creation tests, `create_client_action` creates a new user that gets left in the database, causing `test_delete_user_with_access_privilege` to fail. Not sure if it would be preferable to delete the new user within the client creation tests as an alternative.

tests/scripting/test_fence-create.py
```diff
@@ -51,6 +52,60 @@ def mock_arborist(mock_arborist_requests):
...
+    assert (
+        saved_client.redirect_uri
+        == "https://betawebapp.example/fence\nhttps://webapp.example/fence"
+    )
+    assert saved_client.grant_type == "authorization_code\nrefresh_token\nimplicit"
+    assert saved_client._allowed_scopes == "openid user data"
+
+
+def test_client_create_inits_default_allowed_scopes(db_session):
+    create_client_action(
```
Avantol13, line 75: Your function names are really informational so this is more of a suggestion, but we generally try to add docstrings to any function (even test functions) just to explain what the test is actually doing / testing. We're not 100% consistent about it though and in this case (since these tests are pretty small / isolated) I don't even know if it's necessary. So it's up to you

tests/scripting/test_fence-create.py
Avantol13: yeah, I would actually do cleanup after the new tests you added and try to leave this unchanged if possible

fence/utils.py
```diff
@@ -71,7 +81,7 @@ def create_client(
             client_secret=hashed_secret,
             user=user,
             redirect_uris=urls,
-            _allowed_scopes=" ".join(config["CLIENT_ALLOWED_SCOPES"]),
+            allowed_scopes=allowed_scopes,
```
Avantol13, line 84: looks like this joining is happening anyway b/c of some handling in the `Client.__init__` but unless there's a good reason to change it here, I'd prefer to just keep the `" ".join(allowed_scopes)`. I guess this is a bit more readable, was there some issue with doing it the other way?

Avantol13:
Congrats on first PR :confetti_ball: !!

Avantol13:
Also, for future reference the PXP issue number only has to be in the PR description (not the branch name) to sync with Jira (I think, I could be wrong). so branches are still just `feat/new-thing` or `fix/encoding-bug` etc

fence/utils.py
johnfrancismccann: Sounds good, I don't think there are any issues with leaving it as `_allowed_scopes=" ".join(allowed_scopes)`

tests/scripting/test_fence-create.py
```diff
@@ -51,6 +53,107 @@ def mock_arborist(mock_arborist_requests):
...
+    to_test()
+    driver = SQLAlchemyDriver(db)
+    with driver.session as session:
+        client = session.query(Client).filter_by(name=client_name).first()
+        user = session.query(User).filter_by(username=username).first()
+        if client is not None:
+            session.delete(client)
+        if user is not None:
+            session.delete(user)
+        session.commit()
```
johnfrancismccann, line 87: I was trying to make use of `delete_client_action` and `delete_users` from `fence/scripting/fence_create.py` to perform cleanup here, but running into difficulties. From the perspective of the tests run after `test_create_client_inits_default_allowed_scopes`, the `Client` and `User` records added by `test_create_client_inits_default_allowed_scopes` were still in the database.

johnfrancismccann:
> Congrats on first PR 🎊 !!

Thanks @Avantol13! This should be ready for review again.

fence/utils.py
```diff
@@ -56,6 +57,15 @@ def create_client(
...
     auth_method = "client_secret_basic" if confidential else "none"
+    allowed_scopes = allowed_scopes or config["CLIENT_ALLOWED_SCOPES"]
+    if not set(allowed_scopes).issubset(set(config["CLIENT_ALLOWED_SCOPES"])):
+        raise ValueError(
+            "Each allowed scope must be one of: {}".format(
+                config["CLIENT_ALLOWED_SCOPES"]
+            )
+        )
+    if "openid" not in allowed_scopes:
+        raise ValueError("One of the allowed scopes must be openid")
```
Avantol13, line 68: I think in this case we can actually log a warning and then just append "openid", since this is an error we can correct and there's no case where we want to support *not* having that scope

PR #782, opened 2020-06-05
add program/project info to user.yaml guide

paulineribeyre:
### Improvements
- Add program/project info to user.yaml guide
- Fix "assert 60 <= 59" error in flaky unit tests

### DIFF
--- a/tests/google/test_service_accounts.py
+++ b/tests/google/test_service_accounts.py
@@ -222,7 +222,7 @@ def test_patch_service_account_expires_in(
     ).all()
     for access in service_account_accesses:
         diff = access.expires - int(time.time())
-        assert requested_exp <= diff <= requested_exp + 10
+        assert requested_exp - 2 <= diff <= requested_exp + 10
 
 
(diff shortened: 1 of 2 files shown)


tests/google/test_service_accounts.py
```diff
@@ -222,7 +222,7 @@ def test_patch_service_account_expires_in(
     ).all()
     for access in service_account_accesses:
         diff = access.expires - int(time.time())
-        assert requested_exp <= diff <= requested_exp + 10
+        assert requested_exp - 2 <= diff <= requested_exp + 10
```
vpsx, line 225: Wait... I think the allowance has been on the wrong end of the inequality this whole time????? 🤦 If that checks out to you, can we:
- Do `assert requested_exp - 2 <= diff <= requested_exp` instead
- Change the comment "up to 10 sec" (a few lines prior) to say 2? Or whatever number we want to go with

paulineribeyre: 🤦‍♀️ 🤦‍♀️ 🤦‍♀️ 🤦‍♀️

PR #785, opened 2020-06-12
PXP-5256 Fix broken authorization page login for users with many projects

johnfrancismccann:
### Bug Fixes
Enable a user who has access to a large number of projects to be able to login successfully when loading the Gen3 authorization page. Previously, for users with access to too many projects, attempting to login and load the authorization page resulted in 502 and 400 errors.

### DIFF
--- a/fence/config-default.yaml
+++ b/fence/config-default.yaml
@@ -364,7 +364,7 @@ MAX_ACCESS_TOKEN_TTL: 3600
 # TEMPORARY: The maximum number of projects allowed in token claims.
 # This config var should be removed after sheepdog and peregrine support
 # auth checks against Arborist, and no longer check the token.
-TOKEN_PROJECTS_CUTOFF: 15
+TOKEN_PROJECTS_CUTOFF: 10
 
 
 ########################################################################################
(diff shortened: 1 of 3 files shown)


fence/config-default.yaml
```diff
@@ -364,7 +364,7 @@ MAX_ACCESS_TOKEN_TTL: 3600
 # TEMPORARY: The maximum number of projects allowed in token claims.
 # This config var should be removed after sheepdog and peregrine support
 # auth checks against Arborist, and no longer check the token.
-TOKEN_PROJECTS_CUTOFF: 15
+TOKEN_PROJECTS_CUTOFF: 10
```
johnfrancismccann, line 368: The max that `TOKEN_PROJECTS_CUTOFF` could be configured to and still pass the newly added automated test is 13. I went with 10 here to give a few hundred bytes of a cushion in the header and because it seemed like a natural choice.

fence/resources/openid/google_oauth2.py
```diff
@@ -40,6 +41,8 @@ def get_user_id(self, code):
         """
         Get user id
         """
+        if config.get("MOCK_GOOGLE_AUTH", False):
+            return {"email": "test@gmail.com"}
```
paulineribeyre, line 45: ideally with `MOCK_GOOGLE_AUTH` Fence should grab the fake username from the configured cookie
https://github.com/uc-cdis/fence/blob/2020.06/fence/config-default.yaml#L51-L56
https://github.com/uc-cdis/fence/blob/2020.06/fence/blueprints/link.py#L284-L286

PR #786, opened 2020-06-15
PXP-2799 Access token hash

vpsx:
### New Features
- Include at_hash claim in id tokens where an access token is issued alongside

### Improvements
- Refactor UnsignedIDToken class in order to use both CodeIDToken and ImplicitIDToken from Authlib
- id token validation is now sensitive to present auth flow type (some claims like nonce and at_hash are validated differently in implicit vs code flow)

### DIFF
--- a/tests/oidc/core/token/test_id_token.py
+++ b/tests/oidc/core/token/test_id_token.py
@@ -3,7 +3,11 @@ import time
 
 from authlib.jose.errors import InvalidClaimError
 
-from fence.jwt.token import generate_signed_id_token, UnsignedIDToken
+from fence.jwt.token import (
+    generate_signed_id_token,
+    UnsignedCodeIDToken,
+    UnsignedIDToken,
+)
 from fence.jwt.validate import validate_jwt
 from fence.models import User
 from fence.utils import random_str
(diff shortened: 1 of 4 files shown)


tests/oidc/core/token/test_id_token.py
```diff
@@ -45,7 +49,7 @@ def test_recode_id_token(app, kid, rsa_private_key):
     new_signed_token = original_unsigned_token.get_signed_and_encoded_token(
         kid, rsa_private_key
     )
-    new_unsigned_token = UnsignedIDToken.from_signed_and_encoded_token(
+    new_unsigned_token = UnsignedCodeIDToken.from_signed_and_encoded_token(
```
paulineribeyre, line 52: why do we need to replace UnsignedIDToken -> UnsignedCodeIDToken in this test?

vpsx: Oh good question!
- UnsignedIDToken used to inherit from CodeIDToken and now inherits from IDToken
- CodeIDToken includes `azp` in its `REGISTERED_CLAIMS`; IDToken does not
- The base class JWTClaims `__getattr__` method relies on `azp` being in `REGISTERED_CLAIMS` to make `azp` available as an attr
- So the assert below concerning `azp` started failing 

Basically this test already assumed the token was a CodeIDToken. The new inheritance structure just makes this more explicit.

Lmk if unclear and I will elaborate!

PR #787, opened 2020-06-16
PXP-6047 Support RAS as idp

BinamB:

### New Features
- Implemented RAS as identity provider.


fence/blueprints/login/ras.py
```diff
@@ -0,0 +1,21 @@
...
+            idp_name=IdentityProvider.ras, client=flask.current_app.ras_client
+        )
+
+
+class RASCallback(DefaultOAuth2Callback):
+    def __init__(self):
+        super(RASCallback, self).__init__(
+            idp_name=IdentityProvider.ras,
+            client=flask.current_app.ras_client,
+            username_field="ras",
```
paulineribeyre, line 20: i think something like `username` would make more sense than `ras`. or `ras_username` if you like that better

fence/blueprints/login/__init__.py
```diff
@@ -256,6 +258,10 @@ def provider_info(login_details):
         blueprint_api.add_resource(ORCIDLogin, "/orcid", strict_slashes=False)
         blueprint_api.add_resource(ORCIDCallback, "/orcid/login", strict_slashes=False)
 
+    if "ras" in configured_idps:
+        blueprint_api.add_resource(RASLogin, "/ras", strict_slashes=False)
+        blueprint_api.add_resource(RASCallback, "/ras/callback", strict_slashes=False)
```
paulineribeyre, line 263:
```suggestion
        # note that the callback endpoint is "/ras/callback", not "/ras/login" like other IDPs
        blueprint_api.add_resource(RASCallback, "/ras/callback", strict_slashes=False)
```

fence/config-default.yaml
```diff
@@ -150,6 +150,19 @@ OPENID_CONNECT:
   # Currently supports organizational account only, so when registering a new App in
   # Azure, make sure to select the `Accounts in any organizational directory` for
```
paulineribeyre, line 151: this comment about the azure client should be removed

tests/test-fence-config.yaml
```diff
@@ -174,6 +178,9 @@ LOGIN_OPTIONS:
     idp: microsoft
   - name: 'NIH Login'
     idp: shibboleth
+  - name: 'RAS Login'
+    idp: fence
+    fence_idp: ras
```
paulineribeyre, line 183:
```suggestion
    idp: ras
```

fence/resources/openid/ras_oauth2.py
```diff
@@ -0,0 +1,75 @@
...
+                return {"ras": userinfo["preferred_username"]}
+            elif "UserID" in userinfo and userinfo["UserID"]:
+                return {"ras": userinfo["UserID"]}
+            elif claims["sub"]:
+                return {"ras": claims["sub"]}
+            else:
+                return {"error": "Can't get user's ras"}
+        except Exception as e:
+            self.logger.exception("Can't get user info")
+            return {"error": "Can't get your ras: {}".format(e)}
```
paulineribeyre, line 75: a couple comments here
- we don't want to return the full error to the user (`"Can't get your ras: {}".format(e)`)
- it's good practice to limit the number of return statements; i think this would be easier to read/debug
```
err_msg = "Can't get user's info"
try:
    ...

    username = None
    if userinfo.get("preferred_username"):
        username = userinfo["preferred_username"]
    elif userinfo.get("UserID"):
        username = userinfo["UserID"]
    elif claims.get("sub"):
        username = claims["sub"]
    if not username:
        logger.error("{}, received claims: {} and userinfo: {}".format(err_msg, claims, userinfo))
        return {"error": err_msg}
    return {"username": username}
except Exception as e:
    return {"error": err_msg}
```

docs/fence_shibboleth.md
```diff
@@ -88,6 +88,9 @@ LOGIN_OPTIONS:
     idp: fence
     fence_idp: shibboleth
     shib_idps: '*'
+  - name: 'RAS login'
+    idp: fence
+    fence_idp: ras
```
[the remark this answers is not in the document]
paulineribeyre: if you're updating that section anyway, could you fix the Shibboleth doc link 🤦  it's linking to https://github.com/uc-cdis/fence/blob/master/docs/shibboleth.md instead of https://github.com/uc-cdis/fence/blob/master/docs/fence_shibboleth.md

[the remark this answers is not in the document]
BinamB: oui!

fence/resources/openid/ras_oauth2.py
```diff
@@ -62,14 +62,23 @@ def get_user_id(self, code):
...
 
-            if "preferred_username" in userinfo and userinfo["preferred_username"]:
-                return {"ras": userinfo["preferred_username"]}
-            elif "UserID" in userinfo and userinfo["UserID"]:
-                return {"ras": userinfo["UserID"]}
-            elif claims["sub"]:
-                return {"ras": claims["sub"]}
-            else:
-                return {"error": "Can't get user's ras"}
+            err_msg = "Can't get user's info"
```
paulineribeyre, line 65: this statement should be outside of the `try` block because it's used in the `except` block.

and i'm thinking maybe you should move the `except` block to right after the `get_userinfo` statement, to reduce indentation since the part where you get the username will not throw any unexpected exception

requirements.txt
```diff
@@ -35,7 +35,7 @@ setuptools>=40.3.0
 six==1.11.0
 SQLAlchemy==1.3.3
 temps==0.3.0
-userdatamodel==2.3.2
+git+https://github.com/uc-cdis/userdatamodel.git@feat/ras-login#egg=userdatamodel
```
paulineribeyre, line 38: don't forget to pin this

BinamB: Oh right! I was waiting until the userdatamodel was merged!

fence/__init__.py
```diff
@@ -334,6 +335,14 @@ def _setup_oidc_clients(app):
             logger=logger,
         )
 
+    # Add OIDC client for RAS if configured.
+    if "ras" in oidc:
+        app.ras_client = RASClient(
+            config["OPENID_CONNECT"]["ras"],
```
vpsx, line 341: can we use `oidc["ras"]` instead?

fence/config-default.yaml
```diff
@@ -145,6 +145,14 @@ OPENID_CONNECT:
     # WARNING: DO NOT ENABLE IN PRODUCTION (for testing purposes only)
     mock: false
     mock_default_user: '0000-0002-2601-8132'
+  ras:
+    client_id: ''
+    client_secret: ''
+    redirect_url: '{{BASE_URL}}/login/ras/login/'
```
vpsx, line 151: I think this default should be `/login/ras/callback` instead?

fence/resources/openid/ras_oauth2.py
```diff
@@ -0,0 +1,86 @@
...
+            "authorization_endpoint", ""
+        )
+
+        uri, state = self.session.create_authorization_url(
+            authorization_endpoint, prompt="login"
+        )
+
+        return uri
+
+    def get_userinfo(self, token, userinfo_endpoint, code):
```
vpsx, line 40: Should the `code` parameter be removed?

fence/__init__.py
BinamB: Oh yeah, definitely!

fence/config-default.yaml
BinamB: Yeah! Good catch!

fence/resources/openid/ras_oauth2.py
BinamB: Yeah, completely forgot about removing this one.

PR #788, opened 2020-06-16
PXP-2802 feat(usync-emails)

vpsx:
Q: Mightn't we as well validate actual _provided_ emails including deliverability checks??
A: Not in usersync; if at all, then in useryaml validation

### New Features
- In usersync, populate user email iff email is provided in useryaml OR username is an email addr

### Breaking Changes
- Usersync will no longer populate user's email field in Fence db with the user's username, UNLESS the username is an email address, in which case previous behavior applies (email field is populated with the username). In other words, nothing should change except in cases where usersync was putting invalid emails into user email fields anyway. 

### Dependency updates
- Added dependency email_validator v1.1.1

### DIFF
--- a/fence/sync/sync_users.py
+++ b/fence/sync/sync_users.py
@@ -14,6 +14,7 @@ from stat import S_ISDIR
 
 import paramiko
 from cdislogging import get_logger
+from email_validator import validate_email, EmailNotValidError
 from gen3authz.client.arborist.client import ArboristError
 from gen3users.validation import validate_user_yaml
 from paramiko.proxy import ProxyCommand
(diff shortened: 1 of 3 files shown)


fence/sync/sync_users.py
```diff
@@ -217,12 +218,20 @@ def from_file(cls, filepath, encrypted=True, key=None, logger=None):
...
+                "email": details.get("email", ""),
                 "display_name": details.get("display_name", ""),
                 "phone_number": details.get("phone_number", ""),
                 "tags": details.get("tags", {}),
                 "admin": details.get("admin", False),
             }
+            if not details.get("email"):
+                try:
+                    valid = validate_email(
+                        username, allow_smtputf8=False, check_deliverability=False
```
paulineribeyre, line 230: why `check_deliverability=False`? 😄

vpsx: (Discussed elsewhere--for posterity: No strong opinions either way but we just don't really want to be checking deliverability in the middle of usersync)

PR #792, opened 2020-07-07
PXP-6240 Support ga4gh passport claim in userinfo

vpsx:
~(This needs to be rebased--currently branched off RAS IDP branch. To review, just select the last few commits in the differ)~
~(Not actually doc-only, but want to skip tests until rebase is done)~

### New Features
- Add ga4gh_visa_v1 table to database
- Save user's RAS visas to db upon RAS login
- Include ga4gh_passport_v1 claim in userinfo response

### DIFF
--- a/fence/blueprints/login/ras.py
+++ b/fence/blueprints/login/ras.py
@@ -1,6 +1,8 @@
 import flask
+import jwt
+from flask_sqlalchemy_session import current_session
 
-from fence.models import IdentityProvider
+from fence.models import GA4GHVisaV1, IdentityProvider, User
 
 from fence.blueprints.login.base import DefaultOAuth2Login, DefaultOAuth2Callback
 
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -36,7 +36,7 @@
         "six>=1.11.0,<2.0.0",
         "SQLAlchemy>=1.3.3,<1.4.0",
         "temps>=0.3.0,<1.0.0",
-        "userdatamodel>=2.0.1,<3.0.0",
+        "userdatamodel",
```
jiaqi216, line 39: uh why did this change? because we'll just use whatever requirements.txt gives us?

vpsx: Not sure--that change was not part of my commits--as noted in the description, only the last few commits should be selected 😁

fence/blueprints/login/ras.py
```diff
@@ -19,3 +21,33 @@ def __init__(self):
...
         )
+
+    def post_login(self, user, token_result):
+
+        # TODO: I'm not convinced this code should be in post_login.
+        # Just putting it in here for now, but might refactor later.
+        # This saves us a call to RAS /userinfo, but will not make sense
+        # when there is more than one visa issuer.
+
+        encoded_visas = flask.g.userinfo["ga4gh_passport_v1"]
```
paulineribeyre, line 32: should we use `.get("ga4gh_passport_v1", [])` to be safe?

fence/resources/user/__init__.py
```diff
@@ -128,6 +129,16 @@ def get_user_info(current_session, username):
         optional_info = _get_optional_userinfo(user, requested_userinfo_claims)
         info.update(optional_info)
 
+    # Include ga4gh passport visas
+    # TODO: Respect access token claims (only include if ga4gh_passport_v1 scope present)
```
paulineribeyre, line 133: why not do this right away?

fence/blueprints/login/ras.py
vpsx: Yes thank you! Done~

fence/resources/user/__init__.py
vpsx: [discussed offline; for posterity:] No deliberate reason, but I haven't done it yet 🤷

fence/models.py
```diff
@@ -534,6 +534,27 @@ class ServiceAccountToGoogleBucketAccessGroup(Base):
     )
 
 
+class GA4GHVisaV1(Base):
```
jiaqi216, line 537: @Avantol13 are we happy with these fields? I guess we can always add more later, but these are all the fields I specified in the design doc so far.

fence/resources/user/__init__.py
jiaqi216: I think we'll do this in a future sprint (separate PR) - since we'd like to merge visas ASAP.

PR #795, opened 2020-07-13
feat(google): keep track of user access per group and do a "bulk" upd…

Avantol13:
…ate per group with smarter diff

### New Features
- (new optional configuration) Google Group syncing for usersync can keep track of all groups and users and do one "bulk" update at the end (instead of per user) with a smart diff of expected vs actual group membership in google

### DIFF
--- a/fence/config-default.yaml
+++ b/fence/config-default.yaml
@@ -475,7 +475,15 @@ DBGAP_ACCESSION_WITH_CONSENT_REGEX: '(?P<phsid>phs[0-9]+)(.(?P<version>v[0-9]+))
 #               Also used during User Syncing process to automate managing Storage
 #               access for users.
 # //////////////////////////////////////////////////////////////////////////////////////
+# When true, this modifies usersync (not fence service itself) such that when syncing user
+# access to a Google storage backend happens in "bulk" by doing a diff *per google group* 
+# between what's in Google and what's expected. Then it adds, removes only as necessary.
+# This is in contrast to the default logic which does blind updates per user and ignores
+# 409s from Google.
+# NOTE: This reduces the number of API calls to Google in the general case, but increases
+#       memory usages by usersync (as it has to track all the Google groups and user access)
 GOOGLE_BULK_UPDATES: false
+
 # Configuration for various storage systems for the backend
 # NOTE: Remove the {} and supply backends if needed. Example in comments below
 STORAGE_CREDENTIALS: {}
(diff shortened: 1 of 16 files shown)


fence/resources/google/access_utils.py
```diff
@@ -37,6 +37,40 @@
...
+            for member_email in to_add:
+                logger.info(f"Adding to group {group}: {member_email}")
+                gcm.add_member_to_group(member_email, group)
+
+            # do remove
+            for member_email in to_delete:
+                logger.info(f"Removing from group {group}: {member_email}")
+                gcm.remove_member_from_group(member_email, group)
+
+            logger.info(f"All already in group {group}: {to_update}")
```
vpsx, line 71: Super annoying nitpick but can we move this log message to before the add? I know it says "_already_ in group" but if it's at the end, I think it might look to someone skimming the logs like "here is the full member list now". 

Also if nothing about the intersection members is going to be updated can we rename `to_update` to maybe like `no_change` or something?

fence/resources/storage/__init__.py
```diff
@@ -393,16 +425,36 @@ def _update_access_to_bucket(
...
 
                 bucket_name = bucket_access_group.email
-                self.clients[provider].delete_bucket_acl(bucket_name, storage_username)
 
-                self.logger.info(
-                    "User {}'s Google proxy group ({}) removed or never existed in Google Bucket Access Group {}.".format(
-                        storage_user.email, storage_username, bucket_name
+                if google_bulk_mapping is not None:
+                    google_bulk_mapping.setdefault(bucket_name, [])
+                    while storage_username in google_bulk_mapping[bucket_name]:
```
vpsx, line 431: Just a question: Is this `while` just paranoia or are there normal cases where we expect the storage_username would get appended more than once?

fence/config-default.yaml
```diff
@@ -475,6 +475,7 @@ DBGAP_ACCESSION_WITH_CONSENT_REGEX: '(?P<phsid>phs[0-9]+)(.(?P<version>v[0-9]+))
 #               Also used during User Syncing process to automate managing Storage
 #               access for users.
 # //////////////////////////////////////////////////////////////////////////////////////
+GOOGLE_BULK_UPDATES: false
```
vpsx, line 478: Could you add a short comment (probably just lifted from PR description) saying what this new config does? And maybe explicitly say that it only affects usersync (i.e. it's not config for the GCM or something like that)

Avantol13: for sure :+1:

fence/resources/google/access_utils.py
Avantol13: yeah, that makes sense :+1:

fence/resources/storage/__init__.py
Avantol13: mostly paranoia. `.remove` only removes the first one it finds

PR #797, opened 2020-07-13
fix(client-create-output): Improve output for fence-create client-create

vpsx:
~(I'm pretty sure nothing uses the output of this command programmatically~
Note: gen3-qa and cloud-auto use the fence-create client-create output programmatically, but they take the last line of output. So this PR keeps the last line of output the same. 

### Improvements
Improve output for fence-create client-create (label the id/secret, add instructions to save the creds)

### DIFF
--- a/fence/scripting/fence_create.py
+++ b/fence/scripting/fence_create.py
@@ -109,14 +109,17 @@ def create_client_action(
     DB, username=None, client=None, urls=None, auto_approve=False, **kwargs
 ):
     try:
-        client_id, client_secret = create_client(
-            username, urls, DB, name=client, auto_approve=auto_approve, **kwargs
-        )
         print(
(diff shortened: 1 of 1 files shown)


jiaqi216:
I approved but @vpsx you sure you want to do a print instead of logger? Not a blocker by any means, feel free to merge away as is.

vpsx:
@jiaqi216  Good point but I think the logger prefix (date,  module, etc) would also mess with parsing for the last line of output, and since the last line of output is a print, I think the preceding two should also be. Otherwise it would look pretty weird. In any case this particular output is meant for a human who ran a command at the terminal, so I think print makes sense :D

jiaqi216:
sounds good - let's merge it to make devops' lives easier 😬

PR #800, opened 2020-07-17
Poetry

paulineribeyre:
- replace requirements.txt, dev-requirements.txt and setup.py with pyproject.toml and poetry.lock
- did NOT include in `pyproject.toml` previous dependencies that are not explicitly imported in Fence:
  - from `requirements.txt`: boto, datamodelutils, Flask_OAuthlib, httplib2, oauthlib, requests_oauthlib, pysftp, setuptools, six, temps
  - from `setup.py`: oauth2client, google_api_python_client
- add `awscli` in the Dockerfile instead of pyproject.toml to avoid dependency issues with s3transfer (through storageclient) and botocore. See discussion [here](https://github.com/uc-cdis/fence/pull/745#discussion_r423441492)
- updating `cdiserrors` to 1.x.x in fence or dependencies (gen3config, cdis-python-utils, authutils, gen3authz) creates a lot of conflicts:
  - in Fence, updating cdiserrors to 1.x.x means we need authutils 5.x.x, which [breaks Fence (because of httpx)](https://cdis.slack.com/archives/CLZJVC38B/p1591651699415900). So authutils stays pinned to < 5.0.0
  - to use the latest gen3authz (1.x.x) we need cdiserrors 1.x.x. So gen3authz stays pinned to < 1.0.0 to avoid the authutils issue
  - cdiserrors 1.0.0 is actually [not a breaking change](https://github.com/uc-cdis/cdiserrors/pull/15) so I pinned it to < 2.0.0 to avoid some conflicts
  - => to update everything to the latest versions, we'll need to fix Fence to work with httpx authutils
- rename `fence-create` to `fence_create.py` because poetry imports scripts as python packages
- using git links for dependencies in pyproject.toml is tricky; my updates to the Dockerfile and uwsgi.ini work for the libs I've tested but may not work for everything :(

### Dependency updates
- Fence now uses poetry


.travis.yml
```diff
@@ -11,12 +11,10 @@ addons:
   postgresql: "9.6"
 
 install:
-  - pip uninstall -y six || true # travis installs wrong version
-  - pip uninstall -y userdatamodel || true
-  - pip install -r dev-requirements.txt
-  - pip install -r requirements.txt
+  - curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python
```
BinamB, line 14: So you didnt have the issue with git dependencies with the recent version of poetry?

paulineribeyre: No 🤷‍♀️ I thought I did but it was something else

BinamB (approved):
lgtm!

PR #805, opened 2020-07-27
fix(fence-cleversafe): Always get region to allow cleversafe intergra…

emalinowski:
### Bug Fixes
Always get region to fix cleversafe presigned-urls


fence/blueprints/data/indexd.py
```diff
@@ -685,13 +685,11 @@ def get_signed_url(
 
         # only attempt to get the region when we're not specifying an
         # s3-compatible endpoint URL (ex: no need for region when using cleversafe)
```
paulineribeyre, line 687: please update or remove this comment

PR #807, opened 2020-07-29
Document NIH login issue

paulineribeyre:
Issue found when deploying Fence 2020.05 to KF (which was previously using 4.8.0)

### Improvements
- Document misconfigured NIH login issue


vpsx:
> Issue found when deploying Fence 2020.05 to KF (which was previously using 4.8.0)

Wait I'm confused--the issue arises when we _update_ Fence? I thought this issue was specific to old Fence? (Either way, maybe good to say in the doc which versions have this fixed?)

paulineribeyre:
> Wait I'm confused--the issue arises when we update Fence? I thought this issue was specific to old Fence? (Either way, maybe good to say in the doc which versions have this fixed?)

@vpsx i'm honestly not sure, yes it's related to old Fence, but i think the bug is triggered by one of the orcid login + incommon fixes that happened around versions 4.10.0/4.12.0

PR #810, opened 2020-07-31
task(Observability): Introduce prometheus-flask-exporter and enable Monitoring for PreSigned URLs

themarcelor:
Jira Ticket: [PXP-7422](https://ctds-planx.atlassian.net/browse/PXP-7422)

Exporting metrics from Fence so we can observe events in our Grafana dashboards.

Result:
```
qa-dcp@dev_admin:~$ kc exec -it presigned-url-fence-deployment-664d476489-vt6m4 -- curl http://127.0.0.1/metrics | tail -n3
# HELP pre_signed_url_req_total Multiprocess metric
# TYPE pre_signed_url_req_total counter
pre_signed_url_req_total{file_id="b9c368cd-d681-4d42-b9d1-1778c41fc1f7",username="marceloc@email.com"} 4.0
```

1. Adding new dependencies:
```
prometheus_client = "^0.5.0"
prometheus-flask-exporter = "^0.18.1"
```

2. Enabling uwsgi stats http server.

3. Adding a prometheus Counter to the Presigned URL function.
Tracking the username and the file_id for each presigned url request.

Additional info:
- The Disk space usage is negligible. The metrics & labels remain the same, the only thing that changes is the value of each metric (counter / gauges / histograms):

```
$ kc exec -it presigned-url-fence-deployment-594bc9f4cb-pf4d7 -- bash
...
bash-4.4# du -sh /var/www/metrics/metrics.txt
16.0K	/var/www/metrics/metrics.txt
bash-4.4# du -sh /var/tmp/uwsgi_flask_metrics/
56.0K	/var/tmp/uwsgi_flask_metrics/
```

### New Features
- Exporting metrics from Fence so we can observe events in our Grafana dashboards.

This change is linked with https://github.com/uc-cdis/cloud-automation/pull/1348.

This change is blocked by: https://github.com/uc-cdis/cloud-automation/pull/1498


dockerrun.bash
```diff
@@ -18,3 +18,14 @@ if [ -f /fence/jwt-keys.tar ]; then
...
+
+# add nginx status config
+nginx_status_conf="\ \ \ \ location /nginx_status {\n\ \ \ \ \ \ stub_status;\n\ \ \ \ \ \ allow 127.0.0.1;\n\ \ \ \ \ \ deny all;\n\ \ \ \ \ \ access_log off;\n\ \ \ \ }"
+sed -i "/\ \ \ \ error_page\ 502/i ${nginx_status_conf}" /etc/nginx/sites-available/uwsgi.conf
+# add uwsgi status config
+uwsgi_status_conf="\ \ \ \ location /uwsgi_status {\n\ \ \ \ \ \ proxy_pass \"http://127.0.0.1:9191\";\n\ \ \ \ \ \ allow 127.0.0.1;\n\ \ \ \ \ \ deny all;\n\ \ \ \ \ \ access_log off;\n\ \ \ \ }"
+sed -i "/\ \ \ \ error_page\ 502/i ${uwsgi_status_conf}" /etc/nginx/sites-available/uwsgi.conf
+
+# add another access log in a non-json format
+additional_access_log_conf="\ \ \ \ access_log  /var/log/nginx/access_not_json.log main;"
```
frickjack, line 30: Is this log file going to be rotated?

dockerrun.bash
```diff
@@ -18,3 +18,14 @@ if [ -f /fence/jwt-keys.tar ]; then
     fi
   )
 fi
+
+# add nginx status config
+nginx_status_conf="\ \ \ \ location /nginx_status {\n\ \ \ \ \ \ stub_status;\n\ \ \ \ \ \ allow 127.0.0.1;\n\ \ \ \ \ \ deny all;\n\ \ \ \ \ \ access_log off;\n\ \ \ \ }"
```
frickjack, line 23: Should we update the base image rather than do stuff like this?

fence/blueprints/metrics.py
```diff
@@ -0,0 +1,20 @@
...
+This blueprint defines the /metrics enpdoint to serve a placeholder
+to host fence metrics that will be added later on
+"""
+
+import flask
+
+blueprint = flask.Blueprint("metrics", __name__)
+
+
+@blueprint.route("/", methods=["GET"])
```
frickjack, line 11: It looks like these new endpoints will be publicly accessible?  We don't want that do we?

dockerrun.bash
themarcelor: Dunno, how do we handle the regular json-formatted access log today?

dockerrun.bash
themarcelor: Yeah, let me move this logic to the base image.

fence/blueprints/metrics.py
themarcelor: True. We do need Prometheus to be able to communicate with Fence so perhaps we can block this endpoint at the reverse proxy layer, what do you think?

Avantol13: ehh, we shouldn't rely on authz from revproxy in the event someone is deploying framework services w/o using revproxy/our automation. We want to use Gen3 AuthZ through arborist if possible. last ditch we could use Basic Auth

deployment/uwsgi/uwsgi.ini
```diff
@@ -23,8 +23,10 @@ vacuum = true
 pythonpath = /var/www/fence/
 pythonpath = /fence/
 pythonpath = /usr/local/lib/python3.6/site-packages/
-# poetry installs git dependencies at /usr/local/src
-pythonpath = /usr/local/src/*
```
Avantol13, line 27: don't we still need this line?

fence/blueprints/data/indexd.py
```diff
@@ -76,6 +81,10 @@ def get_signed_url_for_file(action, file_id, file_name=None):
         r_pays_project=r_pays_project,
         file_name=file_name,
     )
+
+    # increment counter for gen3-metrics
+    pre_signed_url_req.labels(current_token["context"]["user"]["name"], file_id).inc()
```
Avantol13, line 86: is this async? how does this metrics overhead affect the response time aggregated over 1000s of signed url requests? we want to make sure it's negligible

fence/blueprints/data/indexd.py
```diff
@@ -76,6 +83,14 @@ def get_signed_url_for_file(action, file_id, file_name=None):
         r_pays_project=r_pays_project,
         file_name=file_name,
     )
+
+    # increment counter for gen3-metrics
+    try:
+        current_user = get_current_user().username
```
themarcelor, line 89: Hi @Avantol13 , I noticed that during load tests, this `current_user` label is always set to `unauthorized` (because there's no Flask Session / The K6 load testing operation just sends a HTTP request with an access token).

I could use some help to identify the best function to utilize here to pick up the name of the user that is sending the PreSigned URL request.
I came across a `get_jwt()` function, but that introduced some issues with the unit tests 🤷🏼‍♂️

fence/blueprints/metrics.py
themarcelor: A different approach has been adopted and this is no longer a concern.

deployment/uwsgi/uwsgi.ini
themarcelor: We do. Good catch 👍🏼 
I put the line back in place.

fence/blueprints/data/indexd.py
themarcelor: Turns out this broke our unit tests as the `current_token` is not always available in all scenarios.
So I've ran some successful experiments with the `current_user` function, but for requests that don't originate in a browser with a user session (HTTP GET requests with the access token only), the `current_user` is always `unauthorized`.

PR #820, opened 2020-08-14
Add Client_ID to Audiences

jiaqi216:
GA4GH AAI requires that client ID be in `aud` in the access token if the `aud` claim is provided. Adding client id and also adding a test to check for compatibility with GA4GH spec. 

This might impact our existing validation if we're checking for audiences.

### DIFF
--- a/fence/jwt/token.py
+++ b/fence/jwt/token.py
@@ -27,7 +27,7 @@ SCOPE_DESCRIPTION = {
     "google_credentials": "Receive temporary Google credentials to access data on Google.",
     "google_service_account": "Allow registration of external Google service accounts to access data.",
     "admin": "View and update user authorizations.",
-    "ga4gh_passport_v1": "Retrieve ga4gh passports and visas",
+    "ga4gh_passport_v1": "Retrieve GA4GH Passports and Visas",
 }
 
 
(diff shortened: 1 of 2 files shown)


fence/jwt/token.py
```diff
@@ -377,10 +377,14 @@ def generate_signed_access_token(
                 "must provide value for `iss` (issuer) field if"
                 " running outside of flask application"
             )
+    audiences = scopes.copy()
+    # include client_id in audiences as required by GA4GH
+    if client_id:
+        audiences.append(client_id)
```
vpsx, line 383: The requirement on `aud` is actually not specific to GA4GH AAI but simply [a part of the generic JWT spec](https://tools.ietf.org/html/rfc7519#section-4.1.3). And the requirement is just that _if_ `aud` is supplied then it must include the `client_id`. So I would reword the comment to not include "as required by GA4GH". (I guess that would leave a very redundant comment so maybe just remove...)

Additionally, given the above, it might make more sense to start with `audiences = client_id if client_id else []`, and _then_ append the scopes to `audiences`, and add a comment explaining that we only append scopes for the sake of backwards compatibility. I think going forward that would just make a lot more sense to someone reading the code, versus starting from the scopes 🤪

jiaqi216: Sounds good, made changes as you requested. Should be good to delete the `audiences = audiences + scope` part if/when we remove scopes from aud

PR #823, opened 2020-08-24
Fix/logout encoding

frickjack:
Jira Ticket: [PXP-6631](https://ctds-planx.atlassian.net/browse/PXP-6631)

* fix `next=` query parameter sent to `fence` idp being double encoded

### Bug Fixes
* fix `next=` query parameter sent to `fence` idp being double encoded


fence/auth.py
```diff
@@ -104,7 +104,7 @@ def logout(next_url):
     elif provider == IdentityProvider.fence:
         base = config["OPENID_CONNECT"]["fence"]["api_base_url"]
         safe_url = urllib.parse.quote_plus(next_url)
```
jiaqi216, line 106: if we're no longer using this variable, might be good to remove it?

frickjack: done :+1:

PR #833, opened 2020-09-11
Fix: Consent deny redirect bug

BinamB:
Hitting the 'Deny' button in the consent screen for NIH RAS would redirect you back to the `/authorization` endpoint which in turn would take you back to the RAS Consent Screen when you go through this flow using a different client than the windmill client. 

### Bug Fixes
- Fixed consent Deny redirecting users back to the consent screen.

### DIFF
--- a/fence/blueprints/login/base.py
+++ b/fence/blueprints/login/base.py
@@ -74,7 +74,6 @@ class DefaultOAuth2Callback(Resource):
         # Check if user granted access
         if flask.request.args.get("error"):
 
-           
             request_url = flask.request.url
             received_query_params = parse_qsl(
                 urlparse(request_url).query, keep_blank_values=True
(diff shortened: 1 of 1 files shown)


fence/blueprints/login/base.py
```diff
@@ -80,6 +82,17 @@ def get(self):
             redirect_query_params = parse_qsl(
                 urlparse(redirect_uri).query, keep_blank_values=True
             )
+            if "client_id" in redirect_uri:
+                redirect_query_dict = parse_qs(
```
paulineribeyre, line 86: you can reuse `redirect_query_params`
and use `if "client_id" in redirect_query_params ` instead of `if "client_id" in redirect_uri`

BinamB: oh i just remembered. `redirect_query_params` is a list of tuples so something like `[('response_type', 'code'), ('client_id', 'lalallala')` so we cant really do ` if "client_id" in redirect_query_params`

paulineribeyre: ok then i'd suggest
```
redirect_query_params = parse_qsl(...)
client_id = dict(redirect_query_params).get("client_id")
if client_id:
    client_id = client_id[0]
    ...
```

BinamB: ohh didnt know we could do that. cool cool cool.

fence/blueprints/login/base.py
```diff
@@ -1,11 +1,12 @@
 import flask
 from flask_restful import Resource
-from urllib.parse import urlparse, urlencode, parse_qsl
+from urllib.parse import urlparse, urlencode, parse_qsl, parse_qs
```
paulineribeyre, line 3: we don't need parse_qs anymore :p

paulineribeyre (approved):

fence/blueprints/login/base.py
BinamB: si just saw the unused import

fence/blueprints/login/base.py
BinamB: oh it shouldnt be `client_id[0]`. client_id is not an array.

PR #834, opened 2020-09-15
(PXP-6755) Allow user to authenticate without RAS visa.

BinamB:
### Bug Fixes
- allow user to authenticate to gen3 successfully even if we fail to retrieve RAS visa during login.


fence/resources/openid/ras_oauth2.py
```diff
@@ -46,6 +46,14 @@ def get_userinfo(self, token, userinfo_endpoint):
         access_token = token["access_token"]
         header = {"Authorization": "Bearer " + access_token}
         res = requests.get(userinfo_endpoint, headers=header)
+        if res.status_code != 200:
+            self.logger.info(
+                "Unable to get visa: status_code: {}, message: {}".format(
+                    res.status_code,
+                    str(res.json()),
```
paulineribeyre, line 53: `format` handles the conversion, you shouldn't need to `str()`

fence/resources/openid/ras_oauth2.py
```diff
@@ -46,6 +46,14 @@ def get_userinfo(self, token, userinfo_endpoint):
         access_token = token["access_token"]
         header = {"Authorization": "Bearer " + access_token}
         res = requests.get(userinfo_endpoint, headers=header)
+        if res.status_code != 200:
+            self.logger.info(
```
paulineribeyre, line 50:
```suggestion
            self.logger.error(
```

fence/resources/openid/ras_oauth2.py
paulineribeyre, line 53: i'd do something like `msg = res.text; try: msg = res.json(); except: pass` because you may not get json if there was an error

paulineribeyre (approved):
do we need a unit test?

PR #841, opened 2020-10-01
(PXP-6533): RAS logout

BinamB:
`ERA_GLOBAL_LOGOUT:` is set to fault on default. Set `ERA_GLOBAL_LOGOUT: true` in fence config to use Itrust to logout from RAS.

### Improvements
- Use Itrust to logout from RAS.


fence/__init__.py
```diff
@@ -144,13 +144,14 @@ def root():
     def logout_endpoint():
         root = config.get("BASE_URL", "")
         request_next = flask.request.args.get("next", root)
+        era_global_logout = flask.request.args.get("era_global_logout") == "true"
```
fantix, line 147: This will be set by ISB I guess? Perhaps we should also add this new parameter in the API docs.

fence/auth.py
```diff
@@ -98,7 +98,9 @@ def logout(next_url):
     # propogate logout to IDP
     provider_logout = None
     provider = flask.session.get("provider")
-    if provider == IdentityProvider.itrust:
+    if provider == IdentityProvider.itrust or (
+        provider == IdentityProvider.ras and era_global_logout
+    ):
```
fantix, line 103: Perhaps it makes more sense to rename the `era_global_logout=False` to `force_era_global_logout=False`, and just have this here:

```python
    if force_era_global_logout or provider == IdentityProvider.itrust:
```

fantix (approved):

PR #842, opened 2020-10-06
Update TECHDEBT.md

VJalili:
Is there any simpler/different method for voting on such techdebts than PRs?


paulineribeyre (approved):

paulineribeyre:
thanks for upvoting :smile: we don't have another way for now

PR #843, opened 2020-10-13
Fix/usersync update role

mattgarvin1:
Problem: previously the usersync job would fail to modify existing roles in Arborist -this is because the Arborist API didn't have a PUT /role/{roleID} endpoint

Solution: this PR solves this problem by having Fence hit the newly implemented PUT /role/{roleID} Arborist endpoint

Note: this is a non-destructive code change in that the new usersync logic is: 
- try to overwrite role by hitting the new Arborist endpoint
- if role overwrite doesn't work (because an older version of Arborist is deployed which does not have the new PUT /role/{roleID} endpoint), go to the pre-existing logic flow, which results in creating new roles that don't already exist in Arborist but does NOT result in modifying existing roles

That is to say - ideally in a given environment Fence and Arborist would be updated to these latest versions simultaneously, BUT if for whatever reason that's not possible - it would not cause any errors to only update Fence with this new version without updating Arborist. 

Similarly, it would not cause any errors to only update the Arborist deployment without updating Fence, because the only difference would be that now that Arborist deployment has a new API endpoint that simply doesn't get used anywhere.

### Bug Fixes
- during usersync - update existing roles in Arborist by hitting the new PUT /role/{roleID} endpoint


Avantol13 (dismissed):

fence/sync/sync_users.py
```diff
@@ -1385,12 +1385,17 @@ def _update_arborist(self, session, user_yaml):
...
         for role in roles:
             try:
-                response = self.arborist_client.create_role(role)
+                response = self.arborist_client.update_role(role["id"], role)
                 if response:
                     self._created_roles.add(role["id"])
-            except ArboristError as e:
-                self.logger.error(e)
-                # keep going; maybe just some conflicts from things existing already
+            except:
```
Avantol13, line 1391: LGTM bot did a nice catch, we should have an Exception type here. `ArboristError` if that's guaranteed, or at least the `Exception` which should catch everything else programming-related. If you leave the `except` as-is, Python will catch weird things like keyboard interrupts and system exists which makes debugging hard

mattgarvin1: Gotcha good call, I will confirm that the `ArboristError` is expected and make the exception handler look for that particular error here.

mattgarvin1: Confirmed - `ArboristError` here is guaranteed, and I updated the exception to expect exactly that error.

fence/sync/sync_users.py
```diff
@@ -1385,12 +1385,18 @@ def _update_arborist(self, session, user_yaml):
...
         for role in roles:
             try:
-                response = self.arborist_client.create_role(role)
+                response = self.arborist_client.update_role(role["id"], role)
                 if response:
                     self._created_roles.add(role["id"])
             except ArboristError as e:
-                self.logger.error(e)
-                # keep going; maybe just some conflicts from things existing already
+                self.logger.info("couldn't put role: {}".format(str(e)))
```
paulineribeyre, line 1392:
```suggestion
                self.logger.info("couldn't update role '{}', creating instead".format(str(e)))
```

paulineribeyre (approved):
lgtm

PR #847, opened 2020-11-03
PXP-7074 fix(google): when linking an external bucket, search for existing buc…

Avantol13:
…ket before creating another

### Bug Fixes
- Fix issue where duplicate bucket entries could get created in the database if you accidentally called `fence-create link-external-bucket` more than once

### DIFF
--- a/fence/scripting/fence_create.py
+++ b/fence/scripting/fence_create.py
@@ -1318,6 +1318,38 @@ def link_external_bucket(db, name):
     with db.session as current_session:
         google_cloud_provider = _get_or_create_google_provider(current_session)
 
+        # search for existing bucket based on name, try to use existing group email
+        existing_bucket = current_session.query(Bucket).filter_by(name=name).first()
+        if existing_bucket:
+            access_group = (
+                current_session.query(GoogleBucketAccessGroup)
+                .filter(GoogleBucketAccessGroup.privileges.any("read"))
+                .filter_by(bucket_id=existing_bucket.id)
+                .all()
+            )
+            if len(access_group) > 1:
+                raise Exception(
+                    f"Existing bucket {name} has more than 1 associated "
+                    "Google Bucket Access Group with privilege of 'read'. "
+                    "This is not expected and we cannot continue linking."
+                )
+            elif len(access_group) == 0:
+                raise Exception(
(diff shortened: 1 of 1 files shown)


fence/scripting/fence_create.py
```diff
@@ -1318,6 +1318,38 @@ def link_external_bucket(db, name):
...
+                    f"Existing bucket {name} has more than 1 associated "
+                    "Google Bucket Access Group with privilege of 'read'. "
+                    "This is not expected and we cannot continue linking."
+                )
+            elif len(access_group) == 0:
+                raise Exception(
+                    f"Existing bucket {name} has no associated "
+                    "Google Bucket Access Group with privilege of 'read'. "
+                    "This is not expected and we cannot continue linking."
+                )
```
paulineribeyre, line 1341: is there no edge case where you'd first create a bucket and later link it?

fence/scripting/fence_create.py
```diff
@@ -1318,6 +1318,38 @@ def link_external_bucket(db, name):
...
+
+            access_group = access_group[0]
+
+            email = getattr(access_group, "email", "N/A")
```
paulineribeyre, line 1345: why not just `access_group.email`? is a gbag without an email valid? [we define it as non nullable](https://github.com/uc-cdis/fence/blob/2020.11/fence/models.py#L420)

fence/scripting/fence_create.py
Avantol13: the bucket and group should get created at the same time. So having a google bucket without a group is not something that should be able to happen under normal circumstances https://github.com/uc-cdis/fence/blob/2a3e7346579eb8cf5e628d116b6c4402b4e33c93/fence/scripting/fence_create.py#L1086-L1093

Avantol13: so it's possible for a bucket to exist without being linked to a project, but there should always be a bucket+group combination

fence/scripting/fence_create.py
Avantol13: oh, this was before I checked the length and the result of query could be `None`. I'll change it

paulineribeyre (approved):

PR #848, opened 2020-11-17
Feat/refresh expires in

mattgarvin1:
Ticket: https://ctds-planx.atlassian.net/browse/PXP-6563

Allow user/client to specify a lifetime of a refresh token - specified in seconds as param expires_in at the /authorize endpoint - must be a positive integer, and any request greater than the config default value (30 days) results in a refresh_token lifetime of the default 30 days.

Primarily this feature is for testing purposes, for interop, for developers of other systems and apps who want to test what happens in the case of an expired refresh token.

### New Features
- Implemented support for specifying lifetime of refresh token at the /authorize endpoint via param expires_in

### Deployment changes
- REQUIRES A FENCE DB MIGRATION, since with this update there is now an additional column in the authorization_code table "refresh_token_expires_in"

### DIFF
--- a/fence/oidc/grants/oidc_code_grant.py
+++ b/fence/oidc/grants/oidc_code_grant.py
@@ -10,12 +10,6 @@ import flask
 from fence.utils import get_valid_expiration_from_request
 from fence.config import config
 from fence.models import AuthorizationCode, ClientAuthType, User
-from sqlalchemy.ext.automap import automap_base
-from sqlalchemy import (
-    MetaData,
-    Table,
-)
-
 
 
 class OpenIDCodeGrant(grants.OpenIDCodeGrant):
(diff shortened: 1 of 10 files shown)


fence/oidc/grants/oidc_code_grant.py
```diff
@@ -35,13 +36,24 @@ def create_authorization_code(client, grant_user, request):
         the arguments passed from the OAuth request (the redirect URI, scope,
         and nonce).
         """
+
+        # requested lifetime (in seconds) for the refresh token
+        refresh_token_expires_in = get_valid_expiration_from_request()
```
vpsx, line 41: I am about to be really annoying, but:
- `get_valid_expiration_from_request` checks the query params for "expires_in"
- So the client in this new flow will be making a request at /authorize with a param "expires_in"
- Usually this /authorize request is thought of as a request for an access token (even though /token returns a bunch of stuff)
- So I'm pretty sure that, when this story is no longer fresh in people's minds, someone will get confused and think that the "expires_in" param refers to the access token expiry. This is worsened by the fact that the /token response (the `{"access_token": "blah", "id_token": "blah, "expires_in": 123}` response) itself has a value named exactly "expires_in", that refers to the _access_ token TTL.

So I am wondering what you think about doing something like: (In fence.utils)

```
def get_valid_expiration_from_request(expiration_param="expires_in"):
if expiration_param in flask.request.args: blah blah
```
This way it still defaults to looking for "expires_in" but you can tell it to look for e.g. "refresh_token_expires_in" when you call it from here. Then the client will hit /authorize with the qp "refresh_token_expires_in" and there will be much less potential confusion.

vpsx: This surfaces the fact that, also in fence.utils, there's a problem with `is_valid_expiration`--it ignores its `expires_in` argument and just grabs it from flask.request.args 🤪 So you'd have to make a slight change there too:

```
try:
    expires_in = int(expires_in)
    assert expires_in > 0
```

fence/oidc/grants/oidc_code_grant.py
```diff
@@ -35,13 +36,24 @@ def create_authorization_code(client, grant_user, request):
...
+        if refresh_token_expires_in:
+            refresh_token_expires_in = min(
```
vpsx, line 43: I would add a new optional parameter to `get_valid_expiration_from_request`, maybe called `configured_max` or `check_max` or just `max`, and shift this validation logic there. (Exactly the same logic: If the max arg was provided, take the minimum of the two, else just return the number from the request)

Then from here you could have just:
```refresh_token_expires_in = get_valid_expiration_from_request(check_max=config["REFRESH_TOKEN_EXPIRES_IN"])```  

It makes sense to me to factor validation checks into the `get_valid_x` utility functions, and as a bonus they'll then be reusable. But up to you.

vpsx (approved):
The db migration and the new flow work swimmingly in my local tests 🎉 I have left only nitpicks made with the long term in mind--in both cases, do feel free to disagree and dismiss 👯

vpsx:
Note: Dropped some merge commits and rebased--commit logs were a mess. Only merge commits were dropped.

PR #849, opened 2020-11-23
(PXP-7166): use ras validation endpoint to validate passports

BinamB:
If the passport is not valid then don't add it to the database. 

### New Features
- Validate RAS passports using RAS's `/passport/validation` endpoint.

### DIFF
--- a/fence/resources/openid/ras_oauth2.py
+++ b/fence/resources/openid/ras_oauth2.py
@@ -107,7 +107,16 @@ class RASOauth2Client(Oauth2ClientBase):
             self.logger.info("Using {} field as username.".format(field_name))
 
             # Save userinfo and token in flask.g for later use in post_login
-            flask.g.userinfo = userinfo if validation == "Valid" else {}
+            flask.g.userinfo = {}
+            if validation == "Valid":
+                self.logger.info("Passport validated")
+                flask.g.userinfo = userinfo
+            else:
+                self.logger.error(
+                    "Passport validation failed. Not storing passport in database. Error: {}".format(
+                        validation
+                    )
+                )
             flask.g.tokens = token
 
         except Exception as e:
@@ -146,7 +155,16 @@ class RASOauth2Client(Oauth2ClientBase):
         encoded_visas = (
             userinfo.get("ga4gh_passport_v1", []) if validation == "Valid" else []
         )
-        # TODO: add logs for visa validation
+
+        encoded_visas = []
+        if validation == "Valid":
+            encoded_visas = userinfo.get("ga4gh_passport_v1", [])
+        else:
+            self.logger.error(
+                "Passport validation failed. Not storing passport in database. Error: {}".format(
+                    validation
+                )
+            )
 
         for encoded_visa in encoded_visas:
             try:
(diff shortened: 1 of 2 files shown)


fence/resources/openid/ras_oauth2.py
```diff
@@ -48,6 +49,17 @@ def get_userinfo(self, token, userinfo_endpoint):
         res = requests.get(userinfo_endpoint, headers=header)
         return res.json()
 
+    def validate_passport(self, validation_endpoint, passport):
+        """
+        Validate passport with RAS's validation endpoint.
+        TODO: Remove this once we can locally validate passports
+        NOTE: RAS has an option to query a single visa with the /passport/validate?visa= but not using these since we hit the limit for an http header pretty quick
```
paulineribeyre, line 56:
```suggestion
        NOTE: RAS has an option to validate a single visa using "/passport/validate?visa=". Not using this since visas are too large (we hit the http header limit)
```

paulineribeyre, line 56: and can you mention why we would want to validate a single visa in the first place

fence/resources/openid/ras_oauth2.py
```diff
@@ -93,7 +110,16 @@ def get_user_id(self, code):
             self.logger.info("Using {} field as username.".format(field_name))
```
paulineribeyre, line 110:
```suggestion
            self.logger.debug("Using {} field as username.".format(field_name))
```

fence/resources/openid/ras_oauth2.py
```diff
@@ -93,7 +110,16 @@ def get_user_id(self, code):
...
 
             # Save userinfo and token in flask.g for later use in post_login
-            flask.g.userinfo = userinfo
+            flask.g.userinfo = {}
+            if validation == "Valid":
+                self.logger.info("Passport validated")
```
paulineribeyre, line 115:
```suggestion
                self.logger.debug("Passport validated")
```

fence/resources/openid/ras_oauth2.py
```diff
@@ -58,11 +70,16 @@ def get_user_id(self, code):
...
             )
+            validation_endpoint = (
+                self.get_value_from_discovery_doc("issuer", "") + "/passport/validate"
+            )
 
             token = self.get_token(token_endpoint, code)
             keys = self.get_jwt_keys(jwks_endpoint)
             userinfo = self.get_userinfo(token, userinfo_endpoint)
 
+            validation = self.validate_passport(validation_endpoint, userinfo)
```
paulineribeyre, line 81: this probably slows down the login flow, do we really need to validate a passport freshly obtained from RAS' userinfo endpoint?

fence/resources/openid/ras_oauth2.py
```diff
@@ -120,12 +146,29 @@ def update_user_visas(self, user):
...
             self.logger.exception("{}: {}".format(err_msg, e))
             raise
 
+        validation_endpoint = (
+            self.get_value_from_discovery_doc("issuer", "") + "/passport/validate"
+        )
+        validation = self.validate_passport(validation_endpoint, userinfo)
+        encoded_visas = (
+            userinfo.get("ga4gh_passport_v1", []) if validation == "Valid" else []
+        )
```
paulineribeyre, line 160: remove this 🗑️

PR #869, opened 2021-01-25
(PXP-6471): Update all user visas

BinamB:
Design doc [here](https://docs.google.com/document/d/1Xt_g2eQa4ChHtjUlhw-B-bvbkYeH5qZbVKvXEUYHLFk/edit#heading=h.5e0lej9k5tiv).

Related [PR](https://github.com/uc-cdis/cloud-automation/pull/1513) in cloud-automation for writing a cronjob yaml.

### New Features
- added logic to update all user visas and refresh tokens.


fence/job/visa_update_cronjob.py
```diff
@@ -0,0 +1,162 @@
...
+
+        await asyncio.gather(*producers)
+        self.logger.info("Producers done producing")
+        await queue.join()
+
+        await asyncio.gather(*workers)
+        await semaphore.join()  # blocks until everything in semaphore is complete
+
+        for w in workers:
+            w.cancel()
```
paulineribeyre, line 81: probably don't need this since the workers stop when the queue is empty and you already did `asyncio.gather(*workers)`

fence/job/visa_update_cronjob.py
```diff
@@ -0,0 +1,162 @@
...
+        for u in updaters:
+            u.cancel()
```
paulineribeyre, line 83: probably need to add `await asyncio.gather(*updaters)` after this - wait for tasks to be cancelled

fence/job/visa_update_cronjob.py
```diff
@@ -0,0 +1,162 @@
...
+
+        self.logger.info(
+            "Visa cron job completed in {}".format(
+                datetime.timedelta(seconds=time.time() - start_time)
+            )
+        )
+
+    async def window(self, db_session, queue, window_idx):
```
paulineribeyre, line 91: you don't need the `queue` parameter

fence/job/visa_update_cronjob.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        window function to get chunks of data from the table
+        """
+        start, stop = self.window_size * window_idx, self.window_size * (window_idx + 1)
+        users = db_session.query(User).slice(start, stop).all()
+        return users
+
+    async def producer(self, db_session, queue, window_idx):
```
paulineribeyre, line 99: you don't need `window_idx` to be a parameter since it always starts at 0 and is incremented in the function loop

fence/job/visa_update_cronjob.py
```diff
@@ -0,0 +1,162 @@
...
+        while True:
+            users = await self.window(db_session, queue, window_idx)
+
+            if users == None:
+                break
+            for user in users:
+                self.logger.info("Producer producing user {}".format(user.username))
+                await queue.put(user)
+            if len(users) < window_size:
+                break
```
paulineribeyre, line 113: could you explain this

fence/job/visa_update_cronjob.py
```diff
@@ -0,0 +1,162 @@
...
+            window_idx += 1
+
+    async def worker(self, name, queue, semaphore):
+        """
+        Create tasks to pass tot updater to update visas AND pass updated visas to _verify_jwt_token for verification
```
paulineribeyre, line 118:
```suggestion
        Create tasks to pass to updater to update visas AND pass updated visas to _verify_jwt_token for verification
```

fence/job/visa_update_cronjob.py
```diff
@@ -0,0 +1,162 @@
...
+        self.logger.info("Total concurrency size: {}".format(self.concurrency))
+        self.logger.info("Total thread pool size: {}".format(self.thread_pool_size))
+        self.logger.info("Total buffer size: {}".format(self.buffer_size))
+        self.logger.info("Total numbner of workers: {}".format(self.n_workers))
+
+        queue = asyncio.Queue(maxsize=self.buffer_size)
+        semaphore = asyncio.Queue(maxsize=self.n_workers)
+        loop = asyncio.get_event_loop()
+
+        producers = [
```
paulineribeyre, line 60: i'd suggest adding a big docstring that explains
- what producers/workers/updaters are doing
- what's in the semaphore (i think it could be renamed to be more explicit - since it's not even an actual semaphore?)
- the overall logic

future maintainers won't refer to the design doc

fence/job/visa_update_cronjob.py
```diff
@@ -0,0 +1,162 @@
...
+                )
+
+            semaphore.task_done()
+
+    def _pick_client(self, visa):
+        """
+        Pick oidc client according to the visa provider
+        """
+        if "ras" in visa.type:
+            oidc = config.get("OPENID_CONNECT", {})
```
paulineribeyre, line 152: i would add an error if `ras` is not in `OPENID_CONNECT`

you are initializing a new RASClient every time you call this function - instead you could initialize the RASClient during the class init, or lazy init with a singleton (the downside is that it wouldn't error right away in case of misconfiguration)

fence/job/visa_update_cronjob.py
```diff
@@ -0,0 +1,162 @@
...
+            return RASClient(
+                oidc["ras"],
+                HTTP_PROXY=config.get("HTTP_PROXY"),
+                logger=logger,
+            )
+
+    def _verify_jwt_token(self, visa):
```
paulineribeyre, line 159: this function is never called

tests/ras/test_ras.py
```diff
@@ -331,3 +333,90 @@ def test_update_visa_token_with_invalid_visa(
...
+@mock.patch("fence.resources.openid.ras_oauth2.RASOauth2Client.get_userinfo")
+@mock.patch("fence.resources.openid.ras_oauth2.RASOauth2Client.get_access_token")
+@mock.patch(
+    "fence.resources.openid.ras_oauth2.RASOauth2Client.get_value_from_discovery_doc"
+)
+def test_cronjob(
+    mock_discovery,
+    mock_get_token,
+    mock_userinfo,
+    config,
```
paulineribeyre, line 347: unused `config`

tests/ras/test_ras.py
```diff
@@ -331,3 +333,90 @@ def test_update_visa_token_with_invalid_visa(
...
+        "sub": "abcd-asdj-sajpiasj12iojd-asnoin",
+        "name": "",
+        "preferred_username": "someuser@era.com",
+        "UID": "",
+        "UserID": "admin_user",
+        "email": "",
+    }
+
+    for i in range(n_users):
+        username = "admin_user_{}".format(i)
```
paulineribeyre, line 379:
```suggestion
        username = "user_{}".format(i)
```

tests/ras/test_ras.py
```diff
@@ -331,3 +333,90 @@ def test_update_visa_token_with_invalid_visa(
...
+    ).decode("utf-8")
+
+    userinfo_response["ga4gh_passport_v1"] = [encoded_visa]
+    mock_userinfo.return_value = userinfo_response
+
+    # test "fence-create update-visa"
+    job = Visa_Token_Update()
+    loop = asyncio.get_event_loop()
+    loop.run_until_complete(job.update_tokens(db_session))
+    # asyncio.run(job.update_tokens(db_session))
```
paulineribeyre, line 417: remove commented out line

tests/ras/test_ras.py
```diff
@@ -331,3 +333,90 @@ def test_update_visa_token_with_invalid_visa(
...
+
+    query_visas = db_session.query(GA4GHVisaV1).all()
+
+    for visa in query_visas:
```
paulineribeyre, line 421: maybe also check the total number of visas == `n_users` - i'd also change `n_users_no_visa` to a different number

fence/job/visa_update_cronjob.py
paulineribeyre: btw, why do all producers read the whole db?

fence/job/visa_update_cronjob.py
BinamB: I think you're probably fine. I'll test without

fence/job/visa_update_cronjob.py
BinamB: the `gather` is usually done just for the producers, since we're gathering data. Doing this gave me some error since i think it closed the event loop before some of the updaters were done

fence/job/visa_update_cronjob.py
BinamB: Instead of just the `ga4ghvisa` table? Thats a good question, i cant remember but thats how i started but i ended up doing this? :shrug:

fence/job/visa_update_cronjob.py
BinamB: Say you have 23 users and your window size is 5. There will be 3 users in the last batch of users and thats when we want to stop.

fence/job/visa_update_cronjob.py
BinamB: correct. We dont have logic for verifying jwt token locally yet so just leaving it be for now.

fence/job/visa_update_cronjob.py
BinamB: I think this is because the `update_visa_tokens` takes in user record as a parameter.

fence/job/visa_update_cronjob.py
paulineribeyre: right i mean wouldn't you want to add a call at the right place, so that when you do need it you'll just have to write the function?

bin/fence_create.py
```diff
@@ -342,24 +342,26 @@ def parse_arguments():
...
         "--concurrency",
         required=False,
-        help="number of concurrent users going through the visa update flow",
+        help="number of concurrent users going through the visa update flow. Default value: 5",
     )
     update_visas.add_argument(
         "--thread-pool-size",
         required=False,
-        help="number of Docker container CPU used for jwt verifcation",
+        help="number of Docker container CPU used for jwt verifcation. Default value: 3",
```
paulineribeyre, line 360: the default is 2 in your code

fence/job/visa_update_cronjob.py
```diff
@@ -0,0 +1,170 @@
...
+        self.concurrency = concurrency or 5
+        self.thread_pool_size = thread_pool_size or 3
+        self.buffer_size = buffer_size or 10
+        self.n_workers = self.thread_pool_size + self.concurrency
+        self.logger = logger
+
+        # Initialize visa clients:
+        oidc = config.get("OPENID_CONNECT", {})
+        if "ras" not in oidc:
+            self.logger.error("RAS client not configured")
```
paulineribeyre, line 47:
```suggestion
        if "ras" not in oidc:
            self.logger.error("RAS client not configured")
            self.ras_client = None
```

fence/job/visa_update_cronjob.py
```diff
@@ -0,0 +1,170 @@
...
+                )
+
+            updater_queue.task_done()
+
+    def _pick_client(self, visa):
+        """
+        Pick oidc client according to the visa provider
+        """
+        if "ras" in visa.type:
+            return self.ras_client
```
paulineribeyre, line 165:
```suggestion
        client = None
        if "ras" in visa.type:
            client = self.ras_client
        if not client:
            # raise Exception or write error log - if just a log, then the caller needs to handle a None return value
        return client
```
handles the case where RAS is not configured, and the case where we don't know how to handle this visa type

fence/job/visa_update_cronjob.py
```diff
@@ -0,0 +1,170 @@
...
+                    "User {} doesnt have visa. Skipping . . .".format(user.username)
+                )
+
+            updater_queue.task_done()
+
+    def _pick_client(self, visa):
+        """
+        Pick oidc client according to the visa provider
+        """
+        if "ras" in visa.type:
```
paulineribeyre, line 164: is `visa.type` a list?

fence/job/visa_update_cronjob.py
paulineribeyre: > you don't need `window_idx` to be a parameter since it always starts at 0 and is incremented in the function loop

unresolved

> why do all producers read the whole db?

i just noticed `for _ in range(1)` in the producers list generator. I think that answers my question. If you want to keep a single producer, you should simplify the initialization of the `producers` variable. If you want the option of having more than one, i think each producer should read its own chuck of DB table instead of the whole table

BinamB: I think one producer is fine since all they would be doing is waiting for workers to be done. Also by design its a single producer.

bin/fence_create.py
BinamB: oh i changed it to 3 in the same commit :thinking:

fence/job/visa_update_cronjob.py
BinamB: its a string. Its annoying. I was using a specific string at some point and they changed it but they'll always have ras in it :shrug:

fence/job/visa_update_cronjob.py
```diff
@@ -161,8 +160,12 @@ def _pick_client(self, visa):
         """
         Pick oidc client according to the visa provider
         """
+        client = None
         if "ras" in visa.type:
-            return self.ras_client
+            client = self.ras_client
+        if not client:
+            raise Exception("Visa Client not set up or not avaialable")
```
paulineribeyre, line 167: print the type

fence/job/visa_update_cronjob.py
paulineribeyre: mm how about `if visa.type in ["ras", ...]`?

BinamB: oh so we just add all the things we know they've used or will be using? I think thats solid. But this field will always have "ras" in it. do we still want to do what you suggested?

paulineribeyre: my concern is about getting a future type of visa that has nothing to do with RAS  but contains the substring "ras"

fence/job/visa_update_cronjob.py
```diff
@@ -0,0 +1,177 @@
...
+            buffer_size: max size of queue
+        """
+        self.chunk_size = chunk_size or 10
+        self.concurrency = concurrency or 5
+        self.thread_pool_size = thread_pool_size or 3
+        self.buffer_size = buffer_size or 10
+        self.n_workers = self.thread_pool_size + self.concurrency
+        self.logger = logger
+
+        self.visa_types = config.get("USERSYNC").get("visa_types")
```
paulineribeyre, line 44:
```suggestion
        self.visa_types = config.get("USERSYNC", {}).get("visa_types", [])
```

paulineribeyre:
```suggestion
        self.visa_types = config["USERSYNC"].get("visa_types", {})
```

or this, because the values should never be empty as long as they are in the default config (see https://github.com/uc-cdis/gen3config if you're interested)

paulineribeyre (approved):

PR #870, opened 2021-01-27
Fix/delete with no urls

mattreex:
https://ctds-planx.atlassian.net/browse/PXP-7389

No longer gives an error when attempting to delete records with no urls listed. Deletion proceeds as usual. 

### Bug Fixes
Fix bug caused by an empty url list when attempting to delete a record.


fence/blueprints/data/indexd.py
```diff
@@ -495,6 +495,7 @@ def delete_files(self, urls=None, delete_all=True):
                 )
             )
             return location.delete(bucket, file_suffix)
```
paulineribeyre, line 497: hey, this function doesn't look right, it `returns` after deleting the first `location` 😱

paulineribeyre:
cc @mcannalte

fence/blueprints/data/indexd.py
paulineribeyre: @mattreex what do you think of this?
```
response = ("No URLs to delete", 200)
for location in locations_to_delete:
    response = location.delete(...)
    if something went wrong, break
return response
```

mattreex: Yeah I thought that was strange as well, but I thought someone would surely have pointed it out by now if this only ever deleted one url. But since you pointed it out too I'll look into it.

mattreex: Guess it wasn't that long ago :) https://github.com/uc-cdis/fence/pull/776 
I think the gist will work, but then only the last response is sent. I think we could make it a generator function...

paulineribeyre: i'd say it's fine to only return the last response if they are all 2xx... As long as we don't ignore error responses

paulineribeyre: and you can update the function's docstring `return` line

fence/blueprints/data/indexd.py
```diff
@@ -470,7 +470,7 @@ def delete_files(self, urls=None, delete_all=True):
             urls (Optional[List[str]])
 
         Return:
-            None
+            Response (str, str)
```
paulineribeyre, line 473:
```suggestion
            Response (str: message, int: status code)
```

fence/blueprints/data/indexd.py
```diff
@@ -494,7 +495,11 @@ def delete_files(self, urls=None, delete_all=True):
                     file_suffix, bucket
                 )
             )
-            return location.delete(bucket, file_suffix)
+            response = location.delete(bucket, file_suffix)
+            # check status code not in 200s
+            if response[1][0] != "2":
```
paulineribeyre, line 500: could use `response.raise_for_status()` or `if response > 399`

tests/data/test_data.py
```diff
@@ -1204,3 +1204,39 @@ def json(self):
         response = client.post("/data/multipart/complete", headers=headers, data=data)
 
         assert response.status_code == 200, response
+
+
+@patch("fence.blueprints.data.indexd.IndexedFile.indexed_file_locations")
+def test_delete_files(
```
paulineribeyre, line 1210:
```
def delete_files(self, urls=None, delete_all=True)
```

`delete_files` is never actually called with the `urls` param - the endpoint where we found the bug calls `delete_files(delete_all=True)`.

could you keep these test cases, but add 2 cases for when there are no urls, with the `delete_all` param, and when when there are urls, with the `delete_all` param

mattreex: I believe delete_all is always set to true, but I can still add the test case for when urls = None

paulineribeyre: the parameter is set to true but the code doesn't "delete all" when `urls` is set

tests/data/test_data.py
```diff
@@ -1204,3 +1204,65 @@ def json(self):
...
+        assert status == 200
+
+        # case with error generated
+        mocklocation.reset_mock()
+        mocklocation.delete.return_value = ("bad response", 400)
+        message, status = indx.delete_files(urls)
+        assert 1 == mocklocation.delete.call_count
+        assert status == 400
+
+        # case without error and files
```
paulineribeyre, line 1245:
```suggestion
        # case without error and urls
```

tests/data/test_data.py
```diff
@@ -1204,3 +1204,65 @@ def json(self):
         response = client.post("/data/multipart/complete", headers=headers, data=data)
 
         assert response.status_code == 200, response
+
+
+def test_delete_files(app, client, auth_client, encoded_creds_jwt, user_client):
```
paulineribeyre, line 1209: looks good, just missing test case “`delete_all` without the `urls` parameter when there are urls to delete"

paulineribeyre (approved):

PR #873, opened 2021-02-11
PXP-7529: sign download urls where no_force_sign=False

mcannalte:
### Bug Fixes

- checks the value of no_force_sign query parameter on /user/data/download/\<guid\> and proceeds to sign urls where this value is false


paulineribeyre (approved):
lgtm!

PR #874, opened 2021-02-15
PXP-7696 Fix/renew access token

mfshao:
This PR fulfills https://ctds-planx.atlassian.net/browse/PXP-7696 (the simplest approach)

Also introduced a new config option: `RENEW_ACCESS_TOKEN_BEFORE_EXPIRATION` and default to `false`. When setting to `true` this new access token refresh behavior will be enable. Setting it to `false` will let fence fall back to to old logic (update only when the current one expires). So we can disable it if this is causing performance issues. 

Deployed and tested in https://qa-mickey.planx-pla.net/

### Improvements
- Let fence refreashes access token in cookie before the current one expires
- New config option `RENEW_ACCESS_TOKEN_BEFORE_EXPIRATION` is added, default value is `false`


vpsx (approved):
LGTM! I would just expand the comment to make explicit that you are intentionally no longer checking for an existing access token. Something like:

`# if a user is logged in, refresh the access token or generate a new one, and set in cookie`

Small change but I think it makes the intention to refresh much clearer :D

PR #877, opened 2021-02-25
(PXP-7539): Fix/populate idp for user info

johnfrancismccann:
[PXP-7539](https://ctds-planx.atlassian.net/browse/PXP-7539)

So that the `/user` endpoint is populated with idp info, make modifications to `login_user` function.

Please see ticket above for more details.

### Bug Fixes
- Fix `/user` endpoint so that idp field is populated for the user.

### Improvements
- Add a docstring for `login_user` function


fence/auth.py
```diff
@@ -61,20 +61,32 @@ def build_redirect_url(hostname, path):
...
 
 def login_user(request, username, provider):
-    user = query_for_user(session=current_session, username=username)
+    """
+    Login a user with the given username and provider. Set values in Flask
+    session to indicate the user being logged in. In addition, commit the user
+    and associated idp information to the db.
 
+    Args:
+        request (flask.request): not currently used by this function, TODO
```
johnfrancismccann, line 70: I was wondering if there were any thoughts on removing `request` from the args? It’s not used in `login_user`, but somehow I’d still be cautious removing it given the number of places that supply it (I counted 7 across all of Fence).

paulineribeyre: i'd definitely remove it to clean up the codebase - if something breaks, Jenkins will let us know

paulineribeyre (approved):

fence/auth.py
```diff
@@ -61,20 +61,32 @@ def build_redirect_url(hostname, path):
...
+    idp = (
+        current_session.query(IdentityProvider)
+        .filter(IdentityProvider.name == provider)
+        .first()
+    )
+    if not idp:
+        idp = IdentityProvider(name=provider)
+    user.identity_provider = idp
+    current_session.add(user)
+    current_session.commit()
```
Avantol13, line 88: we only need to do this commit if a) the user is completely new OR B) the user existed but the idp was not previously there. e.g. doing it every time like this I don't think is necessary and will be a bit of a performance hit

fence/auth.py
```diff
@@ -60,27 +60,56 @@ def build_redirect_url(hostname, path):
...
+
+    user = query_for_user(session=current_session, username=username)
+    if user:
+        #  This expression is relevant to those users who already have user and
+        #  idp info persisted to the database. We return early to avoid
+        #  unnecessarily re-saving that user and idp info.
+        if user.identity_provider and user.identity_provider.name == provider:
+            set_flask_session_values(user)
+            return
+    else:
```
Avantol13, line 97: you can actually remove this else statement and un-indent the next line, b/c you're doing an early return above

johnfrancismccann: This `else` is paired with `if user:`, so imagine a flow where a user does exist but doesn't have an identity provider. In that case we wouldn't necessarily want to overwrite the existing user.

Avantol13: :facepalm: you're right, I didn't look closely enough

Avantol13 (approved):

PR #885, opened 2021-03-15
PXP-7447 audit-service integration

paulineribeyre:
Jira Ticket: [PXP-7447](https://ctds-planx.atlassian.net/browse/PXP-7447)

Audit Service integration:
- add audit service client class, configuration to enable audit logs, and create audit logs when a user downloads data or logs in (during `post_login`)
- add `post_login` calls to all IDPs' callback endpoints, and add a unit test to make sure all future IDP implementations call it too
- if audit logs are enabled, Fence will not come up if it can't reach audit-service. However if audit-service crashes while Fence is already up, Fence will crash if a user logs in or attempts to download data. PXP-7805 will improve this
- only recording successful logins and successful download (not upload) presigned URL requests for now

And other changes:
- remove unused `handle_login` function and `mock_get`, `authentication_headers` fixtures
- the `FenceLogin` (`/login/fence`), `FenceCallback` (`/login/fence/login`), `ShibbolethLogin` (`/login/shib`) and `ShibbolethCallback` (`/login/shib/login`) classes now inherit from the same base classes as other IDP endpoints
- the `/login/fence/login` endpoint now ignores mismatched states if `MOCK_AUTH == True` to facilitate testing
- add error handling to the `/login/ras/callback` endpoint so we get logs if `refresh_token` or `idp_token` are missing before crashing
- remove use of deprecated `DEFAULT_LOGIN_URL` in the `/oauth2` endpoint; check `DEFAULT_LOGIN_IDP` and use this IDP's URL instead
- add `fence_idp` and `shib_idp` to userinfo endpoint

### New Features
- Audit Service integration: Fence creates audit logs for data downloads and user logins

### Deployment changes
- Audit logs for user logins and data downloads can be enabled by adding setting `ENABLE_AUDIT_LOGS` to the fence config (see https://github.com/uc-cdis/fence/blob/1c850d07e54f8527dbde182fdd9dd593d18f660e/fence/config-default.yaml#L586-L588)


fence/__init__.py
```diff
@@ -371,6 +373,19 @@ def _setup_arborist_client(app):
         app.arborist = ArboristClient(arborist_base_url=config["ARBORIST"])
 
 
+def _setup_audit_service_client(app):
+    # Initialize the client regardless of whether audit logs are enabled. This
+    # allows us to call `app.audit_service_client.create_x_log()` from
+    # anywhere without checking if audit logs are enabled. The client
+    # checks that for us.
+    service_url = (
+        app.config.get("AUDIT_SERVICE") or app.config["BASE_URL"].rstrip("/") + "/audit"
```
williamhaley, line 382: Maybe a good spot for `urljoin()`? https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin

fence/auth.py
```diff
@@ -72,16 +72,20 @@ def login_user(username, provider):
 
     """
 
-    def set_flask_session_values(user):
+    def set_flask_session_values(user, fence_idp, shib_idp):
```
williamhaley, line 75: Since this is a nested `def`, do we need to pass these parameters? Aren't they available from the outer scope?

ocshawn:
looks good to me

fence/sync/sync_users.py
```diff
@@ -929,7 +929,7 @@ def _upsert_userinfo(self, sess, user_info):
             u.is_admin = user_info[username].get("admin", False)
 
             # do not update if there is no tag
-            if user_info[username]["tags"] == {}:
+            if not user_info[username].get("tags"):
```
williamhaley, line 932: 👍

fence/blueprints/login/base.py
```diff
@@ -101,8 +103,19 @@ def get(self):
             return resp
         raise UserError(result)
 
-    def post_login(self, user, token_result):
-        pass
+    def post_login(self, user=None, token_result=None):
+        create_login_log(self.idp_name)
+
+
+def create_login_log(idp_name):
```
williamhaley, line 110: Just a thought (no need to take any action), but I wonder if we could make our own `@audit` decorator where needed to generically (more or less) allow for auditable functions 🤔

fence/resources/audit_service_client.py
```diff
@@ -0,0 +1,120 @@
...
+def get_request_url():
+    request_url = flask.request.url
+    base_url = config.get("BASE_URL", "")
+    if request_url.startswith(base_url):
+        request_url = request_url[len(base_url) :]
+    return request_url
+
+
+def is_audit_enabled(category=None):
+    enable_audit_logs = config.get("ENABLE_AUDIT_LOGS") or {}
```
williamhaley, line 18: Could we do...
```suggestion
    enable_audit_logs = config.get("ENABLE_AUDIT_LOGS", {})
```

Doesn't really make a difference though. Just curious!

fence/resources/audit_service_client.py
```diff
@@ -0,0 +1,120 @@
...
+    if category:
+        return enable_audit_logs and enable_audit_logs.get(category, False)
+    return enable_audit_logs and any(v for v in enable_audit_logs.values())
```
williamhaley, line 21: Just curious, why would this happen where we might have the `category` set as a value? Seems like we generally assume it would be a key on the `enable_audit_logs` config?

fence/resources/audit_service_client.py
```diff
@@ -0,0 +1,120 @@
...
+                wait_time *= 2
+        raise Exception(
+            f"Audit logs are enabled but audit-service is unreachable at {status_url}: {r.text}"
+        )
+
+    def check_response(self, resp, body):
+        # The audit-service returns 201 before inserting the log in the DB.
+        # This request should only error if the input is incorrect (status
+        # code 422) or if the service is unreachable.
+        if resp.status_code != 201:
```
williamhaley, line 57: If the audit service becomes unavailable for any reason or an unexpected error occurs, is that acceptable for the user story? In that case might we lose some fidelity on auditing?

williamhaley (approved):
Just a few questions, but lgtm! :shipit:

fence/resources/audit_service_client.py
paulineribeyre: the `gen3config` lib should ensure all the fields that are in the default config file will be in `config`. So i could actually do `config["ENABLE_AUDIT_LOGS"]` instead of `config.get("ENABLE_AUDIT_LOGS")`

But if we configure `ENABLE_AUDIT_LOGS:` then this null value will override the default value, so we need `or {}` to be safe

fence/resources/audit_service_client.py
paulineribeyre: see config [here](https://github.com/uc-cdis/fence/blob/5b7fde4912acea189ddc7ec01e049d6a143c6930/fence/config-default.yaml#L586-L588) - categories are keys, but if we call `is_audit_enabled()` without specifying a category, we want to know if any of the categories are enabled, so if any of the values are True

williamhaley: Aah, I see. Thanks!

fence/resources/audit_service_client.py
paulineribeyre: i'm not sure i understand your question but yes, fence crashing in case of audit-service failure is fine for this ticket, there's a separate ticket to improve this

fence/resources/audit_service_client.py
williamhaley: 👍

fence/resources/audit_service_client.py
williamhaley: Sorry, I worded that poorly, and thanks for confirming! I was wondering if there were some cases where maybe we'd send an audit log request before an action like...

```
# 1 Track an audit entry
POST /audit
# 2 Perform some action
POST /login
```

And if it was possible for the `/audit` action to fail, but the `/login` action to succeed. And so maybe we'd log in but not track it properly, and I was wondering if that was a problem. I wasn't sure if it could be a problem that we didn't have 100% confidence on how well we logged all our auditable behavior.

But it sounds like things should crash for now if that happens and then we'll enhance in the future, so nevermind! 😄

paulineribeyre: gotcha, yeah that would be a problem. the current plan is to use a queueing service that Fence can post to and the audit service can read from

PR #890, opened 2021-03-30
Update fence to now include Okta as an OIDC Identity Provider

cginmn:
Added a new profile for Okta to be additional option for an OpenID Connect (OIDC) Identity Provider.

**New Features**
I essentially took the current Azure AD OIDC profile as a template and modified it to work with Okta as the provider.

**Breaking Changes
Bug Fixes**
**Improvements**
Now supports Okta as an OIDC provider

**Dependency updates
Deployment changes**


fence/resources/openid/okta_oauth2.py
```diff
@@ -0,0 +1,49 @@
...
+        try:
+            token_endpoint = self.get_value_from_discovery_doc(
+                "token_endpoint",
+                "",
+            )
+            jwks_endpoint = self.get_value_from_discovery_doc(
+                "jwks_uri",
+                "",
+            )
+            claims = self.get_jwt_claims_identity(token_endpoint, jwks_endpoint, code)
```
vpsx, line 41: If you create an Okta app, create a new "person", and then "assign" that new person to the app, that person then somehow ends up without a username. Then, when they go through the Fence login flow, the code fails on this line with an AuthlibBaseError, "The 'sub' system claim could not be evaluated." Manually updating the user info with a username solves the problem.

I am completely new to Okta, so I am probably just missing context about how Okta directories/auth servers/OIDC claims usually get configured to work together. Does the above seem like a likely scenario to you? If so, I'd add a note somewhere (maybe config-default) about it, but if not, that's also fine.

vpsx (commented):
Hey @cginmn , thanks for your PR and sorry it took so long to review! This LGTM; just one thing I noticed during manual testing--see inline comment.

Other than that, we're working to get this through the integration tests, and then will approve ASAP. Thanks again!

vpsx:
*whoops--integration tests and also a little bit more manual testing!

fence/resources/openid/okta_oauth2.py
cginmn: Okta is essentially exactly the same as Azure, as an IDP. The only difference is the discovery URL is unique to each tenant. So my code was basically copying the azure code and making discover_url a variable. So, the expected behavior  should be the same as what you expect from the claims coming from the Okta IDP and matching on the email claim. So, If i am understanding your comment correctly, (and forgive if i am not..) i do add my email to the users file and then when i login using okta or azure, as long as the email is passed in the claim, they match and i am authenticated.

cginmn: old PR i didn't close when i was trying to figure out how to work with Gen3 has some tasks i need to add to test code to fix some testing issues. Will add them and update this PR. This PR is a replacement of PR 828. I will close that one. I was asked to remove the "generic" idp features so this PR is only Okta IDP related.

vpsx: Thanks for your response! I probably didn't make myself entirely clear--my fault--but I was referring more to the `get_jwt_claims_identity` function and the `sub` claim, which needs to be present in the JWT regardless of whether the `email` claim is present. In any case, after poking around more with Okta I think it's fine.

vpsx (approved):
LGTM! 

Note for posterity: Needs this cloud-auto modification https://github.com/uc-cdis/cloud-automation/pull/1566 for squid to allow the oauth handshake. If Fence complains about a sub claim, make sure your Okta user has a username and Okta is otherwise configured so that the sub claim is populated. 

Thank you for your contribution @cginmn and for your patience! :pray:

PR #893, opened 2021-04-01
(PXP-6693): Fix generation of signed upload urls for public Indexd records for authz-permitted users.

johnfrancismccann:
Jira Ticket: [PXP-6693](https://ctds-planx.atlassian.net/browse/PXP-6693)

With the old logic for generating a signed url for public records, Fence did not allow upload urls to be generated for records with public `authz` or `acl` values ("/open" and "*", respectively). However, users with a `write-storage` permission on "/open" for Fence should be permitted to generate upload urls.

### Bug Fixes
- Allow permitted users to generate a signed upload url for public records (i.e. those records with `authz=[“/open”]`)

### DIFF
--- a/fence/blueprints/data/indexd.py
+++ b/fence/blueprints/data/indexd.py
@@ -93,9 +93,10 @@ def create_presigned_url_audit_log(indexed_file, action, protocol):
         # fall back on ACL
         resource_paths = indexed_file.index_document.get("acl", [])
     if not protocol:
-        # we can assume there are locations since `get_signed_url()`
-        # used the same logic before this code runs
-        protocol = indexed_file.indexed_file_locations[0].protocol
+        protocol = (
+            indexed_file.indexed_file_locations
+            and indexed_file.indexed_file_locations[0].protocol
+        )
     flask.current_app.audit_service_client.create_presigned_url_log(
         username=user_info["username"],
         sub=user_info["user_id"],
@@ -369,7 +370,7 @@ class IndexedFile(object):
             }
             if not self.check_authz(action_to_permission[action]):
                 raise Unauthorized(
-                    f"Authz check failed: you don't have "
+                    f"Either you weren't logged in or you don't have "
                     f"{action_to_permission[action]} permission "
                     f"on {self.index_document['authz']} for fence"
                 )
--- a/fence/blueprints/data/indexd.py
+++ b/fence/blueprints/data/indexd.py
@@ -465,9 +465,6 @@ class IndexedFile(object):
 
     @cached_property
     def public(self):
-        #  authz_resources = list(self.set_acls)
-        #  authz_resources.extend(self.index_document.get("authz", []))
-        #  return "*" in authz_resources or "/open" in authz_resources
         return self.public_acl or self.public_authz
 
     @cached_property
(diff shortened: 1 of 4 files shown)


fence/blueprints/data/indexd.py
```diff
@@ -94,9 +94,10 @@ def create_presigned_url_audit_log(indexed_file, action, protocol):
         # fall back on ACL
         resource_paths = indexed_file.index_document.get("acl", [])
     if not protocol:
-        # we can assume there are locations since `get_signed_url()`
-        # used the same logic before this code runs
-        protocol = indexed_file.indexed_file_locations[0].protocol
+        protocol = (
+            indexed_file.indexed_file_locations
+            and indexed_file.indexed_file_locations[0].protocol
+        )
```
johnfrancismccann, line 100: I added the check because I was getting a 500 error 'list index out of range' for Indexd records with an empty `urls` field (at least that's what I think was the problem). For example, with the old code, setting the `urls` field to `[]` for `INDEXD_RECORD_WITH_PUBLIC_AUTHZ_AND_ACL_POPULATED` and then running `test_public_authz_and_acl_object_upload_file` resulted in a 500 from the list index.

fence/blueprints/data/indexd.py
```diff
@@ -436,8 +454,18 @@ def check_authz(self, action):
...
+
+        header = flask.request.headers.get("Authorization")
+        if header:
+            try:
+                _, token = header.split(" ")
+            except ValueError:
+                raise Unauthorized("authorization header not in expected format")
+        else:
+            token = None
+
```
johnfrancismccann, line 466: `get_jwt` raises a `401` on a non-existent `Authorization` header, so not using `get_jwt` is to allow for anonymous signed download requests.

fence/blueprints/data/indexd.py
johnfrancismccann: When `indexed_file.indexed_file_locations` is falsy, it would probably be better to set `protocol` to either `None` or `“”` (instead of an empty list)?

fence/blueprints/data/indexd.py
```diff
@@ -363,14 +364,31 @@ def get_signed_url(
...
+        else:
+            if self.public_acl and action == "upload":
+                raise Unauthorized(
+                    "Cannot upload on public files while using acl field"
+                )
+            # don't check the authorization if the file is public
+            # (downloading public files with no auth is fine)
+            if not self.public_acl and not self.check_authorization(action):
+                raise Unauthorized(
+                    "You don't have access permission on this file: {}".format(
```
Avantol13, line 387: should use f-string here too, or .format above, up to you but ideally should be consistent

fence/blueprints/data/indexd.py
Avantol13: can we just try/except instead of implementing custom logic here? might be a bit cleaner

```
try:
  token = get_jwt()
except {{ExceptionObject}} as exc:
  token = None
```

fence/blueprints/data/indexd.py
```diff
@@ -93,9 +93,7 @@ def create_presigned_url_audit_log(indexed_file, action, protocol):
     if not resource_paths:
         # fall back on ACL
         resource_paths = indexed_file.index_document.get("acl", [])
-    if not protocol:
-        # we can assume there are locations since `get_signed_url()`
-        # used the same logic before this code runs
+    if not protocol and indexed_file.indexed_file_locations:
         protocol = indexed_file.indexed_file_locations[0].protocol
```
johnfrancismccann, line 97: Please see [earlier comment](https://github.com/uc-cdis/fence/pull/893#discussion_r605763295) for why this change was made

fence/blueprints/data/indexd.py
Avantol13: you may want to add a quick comment for clarication why catching that error is expected and we can continue. something like `raises unauthorized when user is anonymous (no token is available) but we can still evaluate if an anonymous user can access public data, so continue` maybe something less verbose

Avantol13 (approved):

PR #894, opened 2021-04-06
Fix 'object dict can't be used in 'await' expression'

paulineribeyre:
https://github.com/uc-cdis/gen3authz/pull/35

### Bug Fixes
- Fix 'object dict can't be used in 'await' expression' error during usersync

### Dependency updates
- `gen3authz` to version 1.0.5
- Base image `pybase3-1.5.0` (Rust toolchain fix for cryptography dependency)


BinamB:
Does this fix the CI pipeline? This issue doesn't really show up when you run usersync iirc.

MaribelleHGomez (approved):

Dockerfile
```diff
@@ -1,15 +1,16 @@
...
+FROM quay.io/cdis/python-nginx:pybase3-1.5.0
 
 ENV appname=fence
 
-RUN apk update \
-    && apk add postgresql-libs postgresql-dev libffi-dev libressl-dev \
-    && apk add linux-headers musl-dev gcc \
-    && apk add curl bash git vim make lftp \
-    && apk update && apk add openssh && apk add libmcrypt-dev
+RUN pip install --upgrade pip
```
williamhaley, line 8: Maybe out of scope for this work, but I think it'd be ideal if we could do this and our other common deps in the base `pybase` Docker image 🐳

pyproject.toml
```diff
@@ -27,7 +27,7 @@ flask-cors = "^3.0.3"
 flask-restful = "^0.3.6"
 flask_sqlalchemy_session = "^1.1"
 email_validator = "^1.1.1"
-gen3authz = "^1.0.4"
+gen3authz = "^1.0.5"
```
williamhaley, line 30: Could maybe do `^1.0` to automatically get patch versions. Still have to regenerate the lock file, but at least it's automatic 🤖

williamhaley (approved):
lgtm! :shipit: Just a couple notes/thoughts

BinamB (approved):

Dockerfile
paulineribeyre: true

pyproject.toml
paulineribeyre: done!

paulineribeyre:
> Does this fix the CI pipeline? This issue doesn't really show up when you run usersync iirc.

@BinamB i was able to reproduce the issue by running usersync in my dev env. This bug was non-blocking though (both for the CI pipeline and for usersync) since the client does get created before the error happens

PR #906, opened 2021-04-19
(PXP-7565): User registration, plus CSRF changes

vpsx:
Jira Ticket: [PXP-7565](https://ctds-planx.atlassian.net/browse/PXP-7565)

Two pieces to this PR: (1) Add user registration endpoints, and (2) update CSRF handling.

Registration means that a user provides their name, organization, and email, in order to gain some predefined permissions (i.e. to be automatically added to a preconfigured Arborist group). For more information see the docstrings in the diff. This PR adds the registration form itself along with an admin endpoint to list registered users and their information.

The registration form presents some difficulties with existing CSRF protection, so this PR also includes CSRF changes alongside a sister cloud-automation PR https://github.com/uc-cdis/cloud-automation/pull/1580. More info below.

---

**Instructions for testing/using:**
- Set up a useryaml that has a GROUP with a policy (for example: "registered_users" with "data_upload" policy)
- Your test user should not be in this group
- Set up and run your Arborist
- Run usersync
- Put the name of the group in fence config's REGISTERED_USERS_GROUP; also set fence REGISTER_USERS_ON to true.
- Look in the Fence db and confirm that your test user is not registered (he has no registration_info block in his additional_info column)
- Look in Arborist db and confirm that your test user is not part of the group and does not have the group's policies
- Set up and run your Fence, which is talking to your Arborist
- Log in. You should be redirected to the registration form!
- Register.
- Check Arborist db and confirm that now your test user IS part of the group and DOES have the group's policies.
- Check Fence db and confirm that the registration info you entered is now in the additional_info column.
- Log out. Log in again. There should be no redirect to the registration form.
- You can hit /user/register-user directly to re-register.
- Now log in with a user that is an admin. Hit the /user/register-user/list endpoint. You should see your first test user's registration info.

**NOT implemented in this PR:**
- Nicer UI feedback upon successful registration!
- On login and redirect to registration form, if user declines to register, don't ask again on next login. (Do we want this?) Could set flag in additional_info or could do it cookie-based.

**Background on CSRF:**

Prior to this PR, CSRF protection was being implemented at two levels--once in Fence, and the other in the revproxy. Both Fence and the revproxy used the cookie-to-header method for CSRF prevention. There were multiple problems with this.

First problem: HTML forms. The new registration form, like other HTML forms, is not compatible with the cookie-to-header method. (The usual way to implement CSRF protection with HTML forms is using a hidden form field.) Fence's other, existing HTML form, the OAuth consent form, up until recently [was using AJAX](https://github.com/uc-cdis/fence/blob/61be960499434b39db8793f9cde37741f0709b26/fence/templates/oauthorize.html#L170-L219) instead of a plain form submit to get around this. This additionally necessitated some [interesting](https://github.com/uc-cdis/fence/blob/61be960499434b39db8793f9cde37741f0709b26/fence/templates/oauthorize.html#L188-L190) [gymnastics](https://github.com/uc-cdis/fence/blob/61be960499434b39db8793f9cde37741f0709b26/fence/blueprints/oauth2.py#L129-L133) for redirects to work. Then, recently, Fence began to [only issue HttpOnly cookies](https://github.com/uc-cdis/fence/commit/537c9370f92e4c7cfcb4fb946650b683d61c2391) in response to pen testing, and so the AJAX method technically broke--except that the revproxy was rewriting the csrf cookies anyway, see below, and so things happened to keep working.

Second problem: Fence cookie was being overwritten. Fence [has been operating](https://github.com/uc-cdis/fence/blob/61be960499434b39db8793f9cde37741f0709b26/fence/__init__.py#L398-L427) under the assumption that it manages its own CSRF token. But actually the revproxy [rewrites the token](https://github.com/uc-cdis/cloud-automation/blob/c445e42ad69a8a65eb8846e19cb2d09869f71fd6/kube/services/revproxy/nginx.conf#L260-L264) and [performs its own CSRF checking](https://github.com/uc-cdis/cloud-automation/blob/c445e42ad69a8a65eb8846e19cb2d09869f71fd6/kube/services/revproxy/nginx.conf#L331-L345). Apart from being confusing (see above), this also meant that Fence effectively couldn't implement its own CSRF protection (for example, requests with the correct token in a form field would still fail the revproxy CSRF check), while at the same time the revproxy CSRF fails to accommodate verification via HTML hidden form fields. 

A full reconsideration of CSRF protection in Gen3 (involving revproxy, Fence, and all the other services) was outside the scope of this ticket. The present solution makes partial improvements while maintaining [existing behavior](https://github.com/uc-cdis/fence/blob/6055e4bab45c16623ba9f9a7b778b34b23838cbf/fence/__init__.py#L400-L405) around CSRF checks. (NB: This existing behavior is also the reason why this PR does not fully discard all the current CSRF code in favor of built-in WTForms protection. Further work in that direction might want to start [here](https://flask-wtf.readthedocs.io/en/stable/csrf.html#exclude-views-from-protection).) The changes consist of the following:

1. Remove the Fence code that sets the CSRF cookie
2. In Fence, use [Flask-WTF](https://flask-wtf.readthedocs.io/en/stable/index.html), which wraps [WTForms](https://wtforms.readthedocs.io/en/2.3.x/), to generate a CSRF token that is stored in the session
3. Extend the Fence-specific CSRF check to allow verification against this token either via a header or via a hidden form field
4. Exclude the registration endpoints from the revproxy CSRF check.

See https://github.com/uc-cdis/cloud-automation/pull/1580 for the required revproxy change.

### New Features
Add registration endpoints (registration form endpoint and admin-only list endpoint)

### Improvements
Improvements around CSRF handling

### Dependency updates
New dependency Flask-WTF

### Deployment changes

Requires these cloud-automation updates to revproxy: https://github.com/uc-cdis/cloud-automation/pull/1580

### DIFF
--- a/docs/register.md
+++ b/docs/register.md
@@ -1,0 +1,19 @@
+# Description:
+This register functionality enables user to register before downloading from any commons. Registration means that a user provides their name, organization, and email, in order to gain some predefined permissions (i.e. to be automatically added to a preconfigured Arborist group). This functionality adds the registration form itself along with an admin endpoint to list registered users and their information.
+
+# How to enable:
+- Set up a useryaml that has a GROUP with a policy (for example: "registered_users" with "data_upload" policy)
+- Your test user should not be in this group
+- Set up and run your Arborist
+- Run usersync
+- Put the name of the group in fence config's REGISTERED_USERS_GROUP; also set fence REGISTER_USERS_ON to true.
+- Look in the Fence db and confirm that your test user is not registered (he has no registration_info block in his additional_info column)
+- Look in Arborist db and confirm that your test user is not part of the group and does not have the group's policies
+- Set up and run your Fence, which is talking to your Arborist
+- Log in. You should be redirected to the registration form!
+- Register.
+- Check Arborist db and confirm that now your test user IS part of the group and DOES have the group's policies.
+- Check Fence db and confirm that the registration info you entered is now in the additional_info column.
+- Log out. Log in again. There should be no redirect to the registration form.
+- You can hit /user/register directly to re-register.
+- Now log in with a user that is an admin. Hit the /user/register/list endpoint. You should see your first test user's registration info.
--- a/fence/__init__.py
+++ b/fence/__init__.py
@@ -9,6 +9,7 @@ from flask_sqlalchemy_session import flask_scoped_session, current_session
 from authutils.oauth2.client import OAuthClient
 from cdislogging import get_logger
 from gen3authz.client.arborist.client import ArboristClient
+from flask_wtf.csrf import validate_csrf
 from userdatamodel.driver import SQLAlchemyDriver
 from werkzeug.middleware.dispatcher import DispatcherMiddleware
 from azure.storage.blob import BlobServiceClient
@@ -533,27 +534,24 @@ def check_csrf():
         return
     if not config.get("ENABLE_CSRF_PROTECTION", True):
         return
-    # cookie based authentication
     if flask.request.method != "GET":
-        csrf_header = flask.request.headers.get("x-csrf-token")
-        csrf_cookie = flask.request.cookies.get("csrftoken")
-        referer = flask.request.headers.get("referer")
-        logger.debug("HTTP REFERER " + str(referer))
-        if not all([csrf_cookie, csrf_header, csrf_cookie == csrf_header, referer]):
-            raise UserError("CSRF verification failed. Request aborted")
+        try:
+            csrf_header = flask.request.headers.get("x-csrf-token")
+            csrf_formfield = flask.request.form.get("csrf_token")
+            # validate_csrf checks the input (a signed token) against the raw
+            # token stored in session["csrf_token"].
+            # (session["csrf_token"] is managed by flask-wtf.)
+            # To pass CSRF check, there must exist EITHER an x-csrf-token header
+            # OR a csrf_token form field that matches the token in the session.
+            assert (
+                csrf_header
+                and validate_csrf(csrf_header) is None
+                or csrf_formfield
+                and validate_csrf(csrf_formfield) is None
+            )
 
-
-@app.after_request
-def set_csrf(response):
-    """
-    Create a cookie for CSRF protection if one does not yet exist
-    """
-    if not flask.request.cookies.get("csrftoken"):
-        secure = config.get("SESSION_COOKIE_SECURE", True)
-        # response.set_cookie("csrftoken", random_str(40), secure=secure, httponly=True)
-        # TODO ^ this broke oauthorize form but we probs don't want to revert
-        response.set_cookie("csrftoken", random_str(40), secure=secure, httponly=False)
-
-    if flask.request.method in ["POST", "PUT", "DELETE"]:
-        current_session.commit()
-    return response
+            referer = flask.request.headers.get("referer")
+            assert referer, "Referer header missing"
+            logger.debug("HTTP REFERER " + str(referer))
+        except Exception as e:
+            raise UserError("CSRF verification failed: {}. Request aborted".format(e))
--- a/fence/__init__.py
+++ b/fence/__init__.py
@@ -550,7 +550,9 @@ def set_csrf(response):
     """
     if not flask.request.cookies.get("csrftoken"):
         secure = config.get("SESSION_COOKIE_SECURE", True)
-        response.set_cookie("csrftoken", random_str(40), secure=secure, httponly=True)
+        # response.set_cookie("csrftoken", random_str(40), secure=secure, httponly=True)
+        # TODO ^ this broke oauthorize form but we probs don't want to revert
+        response.set_cookie("csrftoken", random_str(40), secure=secure, httponly=False)
 
     if flask.request.method in ["POST", "PUT", "DELETE"]:
         current_session.commit()
--- a/fence/__init__.py
+++ b/fence/__init__.py
@@ -138,9 +138,7 @@ def app_register_blueprints(app):
         fence.blueprints.privacy.blueprint, url_prefix="/privacy-policy"
     )
 
-    app.register_blueprint(
-        fence.blueprints.register.blueprint, url_prefix="/register-user"
-    )
+    app.register_blueprint(fence.blueprints.register.blueprint, url_prefix="/register")
 
     fence.blueprints.misc.register_misc(app)
 
--- a/fence/__init__.py
+++ b/fence/__init__.py
@@ -51,6 +51,7 @@ import fence.blueprints.well_known
 import fence.blueprints.link
 import fence.blueprints.google
 import fence.blueprints.privacy
+import fence.blueprints.register
 
 
 # for some reason the temp dir does not get created properly if we move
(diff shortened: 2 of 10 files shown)


fence/config-default.yaml
```diff
@@ -802,6 +805,11 @@ SYNAPSE_URI: 'https://repo-prod.prod.sagebase.org/auth/v1'
 SYNAPSE_JWKS_URI:
 SYNAPSE_DISCOVERY_URL:
 SYNAPSE_AUTHZ_TTL: 86400
+# MIDRC user registration feature: Ask users to register (provide name/org/email) on login.
```
Avantol13, line 808: I'd rather not mention specific projects, but just call this an optional feature

fence/templates/register_user.html
```diff
@@ -0,0 +1,107 @@
...
+
+
+{% block header %}
+  <h1>{% block title %}Register{% endblock %}</h1>
+{% endblock %}
+
+{% block content %}
+  <form method="post">
+
+    <h1 class="introduction">Register in order to get access to download data</h1>
```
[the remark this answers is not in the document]
vpsx: Very true, but I was thinking that I should keep it simple until the need arises for very configurable registration forms (if it ever does), rather than adding complexity straight away that may never become useful. Also, the different-registration-forms thing would need not only different templates but also different form classes and maybe different field validators..?

In any case I can definitely still push for this if you want 👍 lmk!

fence/__init__.py
```diff
@@ -136,6 +138,10 @@ def app_register_blueprints(app):
         fence.blueprints.privacy.blueprint, url_prefix="/privacy-policy"
     )
 
+    app.register_blueprint(
+        fence.blueprints.register.blueprint, url_prefix="/register-user"
```
Avantol13, line 142: can we change this to just `/register` ?

fence/blueprints/register.py
```diff
@@ -47,7 +47,9 @@ def xor_with_user_email(form, field):
 
 
 class RegistrationForm(FlaskForm):
-    name = StringField(label="Name", validators=[DataRequired()])
+    # name = StringField(label="Name", validators=[DataRequired()])
```
Avantol13, line 50: dead code, please remove

Avantol13 (changes requested):
See comments, also please update openapi docs for new endpoints and preferably add a new markdown file in the docs folder describing this feature and how it works

fence/__init__.py
rnerella92: Updated the endpoint from /register-user to /register

rnerella92:
> See comments, also please update openapi docs for new endpoints and preferably add a new markdown file in the docs folder describing this feature and how it works

fence/blueprints/login/base.py
```diff
@@ -128,6 +128,14 @@ def _login(username, idp_name, email=None):
     redirect.
     """
     login_user(username, idp_name, email=email)
+    assert flask.g.user is not None
```
Avantol13, line 131: why is this required? How could that be None? I generally don't recommend using assertions like this in code outside of unit tests if you're not catching these exceptions and providing a useful message to the user - rather, you should raise an actual exception with a message. In this case though, I'm not convinced you need this at all. If anything, you should handle that case in the block below

rnerella92:
> See comments, also please update openapi docs for new endpoints and preferably add a new markdown file in the docs folder describing this feature and how it works

Added a markdown file with instructions on how to enable register users

fence/blueprints/register.py
```diff
@@ -0,0 +1,148 @@
...
+    form = RegistrationForm()
+
+    if flask.request.method == "GET":
+        return flask.render_template(
+            "register_user.html",
+            user=flask.g.user,
+            form=form,
+        )
+
+    assert flask.request.method == "POST"
```
Avantol13, line 91: This is not necessary, the flask blueprint only routes GET and POST, and GET is captured above

fence/blueprints/register.py
```diff
@@ -0,0 +1,148 @@
...
+            config["REGISTERED_USERS_GROUP"],
+        )
+
+    # Respect session redirect--important when redirected here from login flow
+    if flask.session.get("redirect"):
+        return flask.redirect(flask.session.get("redirect"))
+    return flask.jsonify(registration_info)
+
+
+@blueprint.route("/list", methods=["GET"])
```
Avantol13, line 132: I'm not a fan of using `/list` in a route, I'd rather we try to follow the pattern of existing "restful" endpoints elsewhere in Gen3. Maybe we need to rethink this...

how about `GET, POST /register` (stays the same) and a new, admin endpoint:

- `GET /admin/register` : lists all registrations
- `POST/PUT /admin/register` : could allow manually modifying a registration in the future

Avantol13: that seems a lot cleaner to me, and easier to extend in the future if we need more admin functionality

docs/register.md
```diff
@@ -0,0 +1,19 @@
+# Description:
```
Avantol13, line 1: This is a great start, can you mention:

* This is an optional feature
* The provided user information IS NOT VALIDATED IN ANY WAY currently
* The registration page is not customizable at the moment

This is important information for other people seeing this. I would also add this information in the PR description

Avantol13 (changes requested):
Thanks for the updates, you still need to update https://github.com/uc-cdis/fence/blob/master/openapis/swagger.yaml#L1-L0 with the new endpoints

fence/blueprints/register.py
rnerella92: Updated /list endpoint to /admin/register

docs/register.md
rnerella92: Added these to markdown file

openapis/swagger.yaml
```diff
@@ -1543,6 +1543,20 @@ paths:
               schema:
                 type: string
                 example: "<h1>Gen3 Privacy Policy</h1> ..."
+  /user/register:
+    get:
+      tags:
+        - register-user
```
Avantol13, line 1549:
```suggestion
        - register
```

openapis/swagger.yaml
```diff
@@ -1543,6 +1543,20 @@ paths:
...
+      summary: Register in order to get access to download data
+      description: >-
+        This register functionality enables user to register before downloading from any commons. Registration means that a user provides their firstname, lastname, organization, and email, in order to gain some predefined permissions (i.e. to be automatically added to a preconfigured Arborist group).
```
Avantol13, line 1552: I might rephrase this in your docs. It doesn't necessarily do that if we just consider what's happening in fence. It enables a registration page for users to provide basic information before "being added to a preconfigured group". That group can be given special permissions depending on the use case. So I would try to avoid saying "before downloading from any commons" b/c that's just one example

Avantol13: I would also mention that this GET request redirects to a simple HTML form to be filled out manually

docs/register.md
Avantol13: where did you add this? I don't see it in these docs

rnerella92: Last three lines [here](https://github.com/uc-cdis/fence/blob/feat/register/docs/register.md)

Avantol13: ah, I see. I think this should not be in the "How to enable:" section. It should probably be before that section, since it's not really part of how to enable, it's more information about the feature in general

PR #904, opened 2021-04-16
Use PLANXCYBORG_GITHUB_TOKEN instead of GITHUB_TOKEN

paulineribeyre:
The GITHUB_TOKEN provided automatically in GitHub workflows only has "read" access when a workflows runs on a fork (see https://github.community/t/github-actions-are-severely-limited-on-prs/18179#M9249), so it's blocking our tests on PRs from external contributors. Switch to using `PLANXCYBORG_GITHUB_TOKEN` which will always have write access

PR #907, opened 2021-04-20
fix(ci-token): use GITHUB_TOKEN

williamhaley:
### Improvements

* Use GITHUB_TOKEN for wool in CI

### DIFF
--- a/.github/workflows/wool.yml
+++ b/.github/workflows/wool.yml
@@ -14,4 +14,4 @@ jobs:
 
     - uses: uc-cdis/wool@master
       env:
-        GITHUB_TOKEN: ${{ secrets.PLANXCYBORG_GITHUB_TOKEN }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/wool.yml
```diff
@@ -14,4 +14,4 @@ jobs:
 
     - uses: uc-cdis/wool@master
       env:
-        GITHUB_TOKEN: ${{ secrets.PLANXCYBORG_GITHUB_TOKEN }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
williamhaley, line 17: > With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository. The permissions for the GITHUB_TOKEN in forked repositories is read-only. For more information, see "Authenticating with the GITHUB_TOKEN.

https://docs.github.com/en/actions/reference/events-that-trigger-workflows

> Before each job begins, GitHub fetches an installation access token for the job. The token expires when the job is finished.

https://docs.github.com/en/actions/reference/authentication-in-a-workflow

GITHUB_TOKEN is not a true 'secret' for workflows. It's ok to use in external PRs. It's a _one_time_ secret per-action run and external PRs only have read access - https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token

paulineribeyre (approved):

williamhaley:
Adding a comment for posterity. This reverted https://github.com/uc-cdis/fence/pull/904 and instead we improved wool in https://github.com/uc-cdis/wool/pull/11 to play better with forks

PR #911, opened 2021-05-12
Documentation about public data

paulineribeyre:
### Improvements
- Documentation about public data

### DIFF
--- a/docs/user.yaml_guide.md
+++ b/docs/user.yaml_guide.md
@@ -8,6 +8,9 @@
   - [Notes](#notes)
 - [Deprecated format](#deprecated-format)
   - [For Gen3 Data Commons that do not use Arborist or use the Google Data Access method of Google Service Account Registration](#for-gen3-data-commons-that-do-not-use-arborist-or-use-the-google-data-access-method-of-google-service-account-registration)
+- [Public data](#public-data)
+  - [Public access to resources](#public-access-to-resources)
+  - [The "/open" resource](#the-open-resource)
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


docs/user.yaml_guide.md
```diff
@@ -236,3 +239,71 @@ users:
...
 
 The `user_project_to_resource` section can be used to avoid specifying a resource path for each `users.projects.resource`.
+
+## Public data
+
+> What is involved in making a project "public"; that is, making both the metadata and object files accessible to anyone who visits the Data Commons?
+
+### Public access to resources
+
+Arborist can be configured to apply a policy to _all_ users who visit the system. This is done via the special `user.yaml` field `anonymous_policies`.
```
Avantol13, line 249: you may also want to mentioned authenticated_users group as well, in case they don't want true anon access

Avantol13: a lot of cases still require an audit trail of who's accessing the data

paulineribeyre: mentioned a couple paragraphs below this one :D

docs/user.yaml_guide.md
```diff
@@ -236,3 +239,71 @@ users:
 ```
 
 The `user_project_to_resource` section can be used to avoid specifying a resource path for each `users.projects.resource`.
+
+## Public data
```
Avantol13, line 243: I wondering if this really belongs in Fence's docs, I guess there's not a better place to put it and since it does involve user.yaml I think this is fine

paulineribeyre: Agreed, I asked the user services teams but didn't get answers. Unfortunately I think right now our only option for public docs that are not specific to a service/repo is gen3.org. I think we should start a "public-wiki" repo or a documentation website like we discussed a couple times.
But someone asked me about this and is waiting for an answer, so I just put this doc here for now.

docs/user.yaml_guide.md
```diff
@@ -236,3 +239,71 @@ users:
...
+      - storage_reader
+      resource_paths:
+      - /programs/PUBLIC_PROGRAM
+```
+
+Arborist is very flexible: we could define an open policy per public program, or per public project, or even a single open policy with a list of all open resources.
+
+Note that the same can be done with `all_users_policies` instead of `anonymous_policies` if access should be granted to all authenticated users instead of both authenticated and non-authenticated users.
+
+### The "/open" resource
```
Avantol13, line 283: is this behavior something we really want to support in the future? I think there was some work to move away from this hard-coded logic in some respect, though the non-signed URLs are maybe another thing we'd have to change (maybe add a query param to force_nonsign and depending on the bucket cfg it may or may not actually end up with an accessible URL? This might not be as elegant actually :( 

Ugh, I'd lean towards putting some disclaimer here that this is a hard-coded current state that may change in the future and that wherever possible, people should **not** rely on this behavior for "/open"

paulineribeyre: mm I wasn't aware this was something we wanted to change. I don't know of any code change to move away from this. I can add to the docs that we don't recommend using it but I wasn't aware that we don't :p

docs/user.yaml_guide.md
Avantol13: lol, I commented before I got to that point, thanks

docs/user.yaml_guide.md
Avantol13: I guess I imagined that we had started to move away from this :sweat_smile: I don't like the hard-coded resource path for open data b/c we can't always force consistency in those resources so things get confusing if a project decides to use "/programs/open_access/projects/1000G" for example instead of "/open". the only logic that relies on this seems to be what you describe (e.g. don't "sign" the URL if we expect it to be public) but even the presence of "/open" doesn't necessarily mean we know the bucket itself is public (e.g. our endpoint could return a non-signed URL that doesn't even work b/c the bucket isn't open). 

So I guess most of this has been in my head, but it probably makes sense for us to rethink this and either allow the client/user to ask us to not sign things if they want that instead of trying to be smart about it (which puts the burden on them knowing when the bucket is open :man_shrugging: ). Or even changing this logic to somehow check if a resource exists in a public group (e.g. anon_users or all_users) rather than checking if it's "/open"

Annnnyway, I guess I'd be happier if we put some disclaimer that we may alter the behavior around "/open" in the future so as not to have hard-coded resource logic in Fence

Avantol13 (approved):

PR #913, opened 2021-05-18
chore(observability): Feature flag for Prometheus metrics

themarcelor:
Introducing new ENABLE_PROMETHEUS_METRICS config to allow operators to enable/disable metrics

### Improvements
- Introducing new ENABLE_PROMETHEUS_METRICS config to allow operators to enable/disable metrics

### DIFF
--- a/tests/data/test_data.py
+++ b/tests/data/test_data.py
@@ -105,6 +105,7 @@ def test_indexd_prometheus_presigned_url_counter(
     """
     Test that when a user requests a presigned URL, the prometheus counter is increased.
     """
+    fence.auth.config["ENABLE_PROMETHEUS_METRICS"] = True
     before = (
         app.prometheus_registry.get_sample_value(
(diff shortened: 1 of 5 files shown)


m0nhawk (approved):

tests/data/test_data.py
```diff
@@ -105,6 +105,7 @@ def test_indexd_prometheus_presigned_url_counter(
     """
     Test that when a user requests a presigned URL, the prometheus counter is increased.
     """
+    fence.auth.config["ENABLE_PROMETHEUS_METRICS"] = True
```
paulineribeyre, line 108: i think you don't need this line since you already enabled it in `test-fence-config.yaml`?

themarcelor: True. Thanks 🙏🏼

paulineribeyre (approved):

PR #918, opened 2021-05-25
chore(docs): update some docs, also make sure to use jsonify for corr…

Avantol13:
…ect content headers

### Improvements
- use jsonify for response for correct content headers
- update docs to include new google endpoint
- update docs to clarify header for shib

### DIFF
--- a/fence/blueprints/data/blueprint.py
+++ b/fence/blueprints/data/blueprint.py
@@ -55,8 +55,7 @@ def delete_data_file(file_id):
                 return flask.jsonify({"message": message}), status_code
 
             try:
-                delete_info = record.delete()
-                return flask.jsonify(delete_info)
+                return record.delete()
             except Exception as e:
                 logger.error(e)
                 return (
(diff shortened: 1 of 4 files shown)


fence/blueprints/data/blueprint.py
```diff
@@ -55,7 +55,8 @@ def delete_data_file(file_id):
                 return flask.jsonify({"message": message}), status_code
 
             try:
-                return record.delete()
+                delete_info = record.delete()
+                return flask.jsonify(delete_info)
```
williamhaley, line 59: I can't tell off-hand, but what status do we return on deletes? Just checking we're not returning a 204 with data now.

williamhaley: And do we also need this on line 98?

Avantol13: yeah this ended up breaking some unit tests, I'm taking a closer look

Avantol13: oh I goofed. The delete function when successful is not returning JSON, it's a tuple with empty text and a 204. and when json is returned, it's jsonified in the delete call. Whoops, so I'm removing this change, it's not necessary

williamhaley (approved):
:shipit:

PR #921, opened 2021-05-25
fix(google_sa): new endpoint should replicate exactly what SA signed …

Avantol13:
…URLs are using, correct

### Bug Fixes
- /google/primary_google_service_account was returning the client/user SA, which is NOT what is used for signed urls. this PR corrects the behavior to create and return the user's single primary service account

### DIFF
--- a/fence/blueprints/google.py
+++ b/fence/blueprints/google.py
@@ -36,6 +36,7 @@ from fence.resources.google.utils import (
     get_monitoring_service_account_email,
     get_registered_service_accounts,
     get_project_access_from_service_accounts,
+    get_or_create_primary_service_account_key,
 )
 from fence.models import UserServiceAccount
 from fence.utils import get_valid_expiration_from_request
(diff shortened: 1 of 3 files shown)


fence/blueprints/google.py
```diff
@@ -613,20 +614,20 @@ def post(self):
         relying on lazy creation at first time of Google Data Access.
         """
         user_id = current_token["sub"]
-        client_id = current_token.get("azp") or None
         proxy_group_id = get_or_create_proxy_group_id()
         username = current_token.get("context", {}).get("user", {}).get("name")
+        service_account_email = None
 
-        service_account = get_or_create_service_account(
```
vpsx, line 620: tiny tiny annoying nitpick but i think you can remove the import now

vpsx (approved):

tests/conftest.py
```diff
@@ -1110,15 +1111,40 @@ def primary_google_service_account_google(
     db_session.add(service_account)
     db_session.commit()
 
+    service_account_key_db_entry = models.GoogleServiceAccountKey(
+        key_id=1, service_account_id=service_account.id, expires=int(time.time()) - 3600
```
vpsx, line 1115: Oof sorry, just one more thing, is this supposed to be an expired key?

fence/blueprints/google.py
Avantol13: good call

tests/conftest.py
Avantol13: oh, oof. it's not but I'm not hitting the endpoint that actually checks the expiration. but for clarity, I'll update it

PR #923, opened 2021-05-26
PXP-7805 Push audit logs to an AWS SQS

paulineribeyre:
Jira Ticket: [PXP-7805](https://ctds-planx.atlassian.net/browse/PXP-7805)

goes with https://github.com/uc-cdis/audit-service/pull/2 and https://github.com/uc-cdis/cloud-automation/pull/1603

### New Features
- Fence can be configured to push audit logs to an AWS SQS
- Record all requests instead of only successful requests by creating audit logs right before returning the response

### Improvements
- Audit logs for login events now have the "code" and "state" query parameters redacted
- Fence unit tests for login events now test all available IDPs

### Deployment changes
- For existing deployments with "ENABLE_AUDIT_LOGS" turned ON, to DISABLE pushing audit logs to an AWS SQS, the configuration "PUSH_AUDIT_LOGS_CONFIG.type" must be changed to "api" (default: "aws_sqs"). If audit-service has never been deployed, no action required.
- To ENABLE pushing audit logs to an AWS SQS (REQUIRES audit-service 1.0.0 or more recent), create/update the configuration file's "PUSH_AUDIT_LOGS_CONFIG" field ("type: aws_sqs" (default), and the SQS URL and region can be copied from the audit-service configuration - or run `gen3 sqs info $(gen3 api safe-name audit-sqs)`), see https://github.com/uc-cdis/fence/blob/fec3dd937d68384e86b2ed6f9b53daf0026c2df9/fence/config-default.yaml#L632-L636. Then run `kubectl delete secret fence-config` and `gen3 kube-setup-fence`


fence/config-default.yaml
```diff
@@ -624,6 +624,17 @@ AUDIT_SERVICE: 'http://audit-service'
 ENABLE_AUDIT_LOGS:
   presigned_url: false
   login: false
+# `PUSH_AUDIT_LOGS_CONFIG.type` is one of: [api, aws_sqs].
+# - if type == api: logs are created by hitting the log creation endpoint.
+# - if type == aws_sqs: logs are pushed to an SQS and fields `sqs_url` and
+# `region` are required. Fields `aws_access_key_id` and
+# `aws_secret_access_key` are optional.
+PUSH_AUDIT_LOGS_CONFIG:
+  type: aws_sqs
```
Avantol13, line 633: defaulting to this means that existing deployments with ENABLE_AUDIT_LOGS on will now error until they flip this to `api`, right? We may want to note that in `DEPLOYMENT CHANGES`. I know you note something about enabling them, but may want to explicitly mentioned that even for deployments that DON'T want this, they have to update the cfg (unless you change this default)

fence/blueprints/data/indexd.py
```diff
@@ -84,30 +94,21 @@ def get_signed_url_for_file(action, file_id, file_name=None):
...
+def prepare_presigned_url_audit_log(protocol, indexed_file):
+    """
+    Store in `flask.g.audit_data` the data needed to record an audit log.
+    """
     resource_paths = indexed_file.index_document.get("authz", [])
     if not resource_paths:
         # fall back on ACL
         resource_paths = indexed_file.index_document.get("acl", [])
-    if not protocol and indexed_file.indexed_file_locations:
+    if not protocol and len(indexed_file.indexed_file_locations) > 0:
```
Avantol13, line 108: shouldn't the truthy check above already handle this?

paulineribeyre: which check? if the indexd record `urls` field is empty, we need to check this here - we record no protocol, and then the presigned URL generation function errors 'there are no locations'

fence/__init__.py
```diff
@@ -487,3 +483,58 @@ def set_csrf(response):
...
+            if param in query_params:
+                query_params[param] = "redacted"
+        url_parts = list(parsed_url)  # cast to list to override query params
+        url_parts[4] = urlencode(query=query_params)
+        request_url = urlunparse(url_parts)
+
+    return request_url
+
+
+@app.after_request
```
Avantol13, line 505: is this called after every request? :grimacing: I'm a little nervous of the overhead of this. I like the idea of consolidating and making it cleaner, but could we isolate this cleaning logic to only functions where we know we're capturing audit logs?

Maybe make a custom decorator and only decorate functions we know we want to audit log? Also only do the decorator logic if audit logs are enabled.

I also want to make sure that if someone doesn't want ANY logging, it shouldn't impact performance

paulineribeyre: good point, i'll look into that

Avantol13: Maybe something like [this conditional decorator](https://stackoverflow.com/questions/10724854/how-to-do-a-conditional-decorator-in-python) based on if audit logs are enabled. And then this function basically becomes a decorator only used when it's enabled and only on functions where we know we want to audit log.

fence/blueprints/data/indexd.py
```diff
@@ -1008,21 +1009,22 @@ def delete(self, bucket, file_id):
             return ("Failed to delete data file.", status_code)
 
 
-def _get_user_info(sub_to_string=True):
+def _get_user_info(sub_type=str):
     """
     Attempt to parse the request for token to authenticate the user. fallback to
     populated information about an anonymous user.
+    By default, cast `sub` to str. Use `sub_type` to override this behavior.
```
Avantol13, line 1016: why not just always cast to string?

fence/resources/audit_service_client.py
```diff
@@ -59,37 +78,58 @@ def check_response(self, resp, body):
                 err = resp.json()
             except Exception:
                 err = resp.text
-            self.logger.error(f"Unable to POST audit log `{body}`. Details:\n{err}")
+            self.logger.error(
+                f"Unable to POST audit log `{body}`. Status code: {resp.status_code} - Details:\n{err}"
+            )
             raise InternalError("Unable to create audit log")
 
+    def create_audit_log(self, category, data):
```
Avantol13, line 86: You have another function called `create_audit_log` in a different module. Probably fine but _might_ be confusing. Pending if/how we do the decorator/performance improvements

fence/blueprints/data/indexd.py
paulineribeyre: I didn't look too much into it, but `authutils.current_token["sub"]` is a string and `flask.g.user.id` is an int 🤔
I'm casting `authutils.current_token["sub"]` to int here because audit-service expects an int, to match the type of `sub` in the fence DB.

tests/test-fence-config.yaml
```diff
@@ -479,6 +479,8 @@ AUDIT_SERVICE: 'http://audit-service'
 ENABLE_AUDIT_LOGS:
   presigned_url: false
   login: false
+PUSH_AUDIT_LOGS_CONFIG:
+  type: api
```
Avantol13, line 483: we should also write some tests that monkey patch this and add others for `aws_sqs` and mock out the AWS call for better coverage of the new feature

Avantol13 (changes requested):
See comments. tl;dr I think we should isolate this more to reduce potential impact to performance

tests/test-fence-config.yaml
paulineribeyre: Yeah the current tests are testing "type: api", so the audit_service_client module without the SQS part. Tbh I wasn't sure how to test it since _everything_ would be mocked. But i'll spend some more time on that

fence/blueprints/data/indexd.py
Avantol13: sorry, I meant the previous code. like why are we switching from `indexed_file.indexed_file_locations` to `len(indexed_file.indexed_file_locations) > 0`. For situations where urls are empty the previous code should have been `False`.

paulineribeyre: woops, yeah i'm not sure what happened there 😆

fence/config-default.yaml
paulineribeyre: done

fence/resources/audit/client.py
```diff
@@ -0,0 +1,155 @@
...
+                aws_access_key_id=config["PUSH_AUDIT_LOGS_CONFIG"].get(
+                    "aws_access_key_id"
+                ),
+                aws_secret_access_key=config["PUSH_AUDIT_LOGS_CONFIG"].get(
+                    "aws_secret_access_key"
+                ),
+            )
+
+    def ping(self):
+        max_tries = 3
```
Avantol13, line 38: can you use backoff lib for this to remain consistent? we're using it elsewhere https://github.com/uc-cdis/fence/blob/b85118f4b0bbcba5de6a856ef275f7c45ef3b687/fence/resources/openid/ras_oauth2.py#L118

Avantol13: you can override the exception/handling of default if you need to

fence/resources/audit/client.py
```diff
@@ -0,0 +1,155 @@
...
+        self.logger = logger
+        self.push_type = config["PUSH_AUDIT_LOGS_CONFIG"].get("type", "api")
+
+        # audit logs should not be enabled if the audit-service is unavailable
+        if is_audit_enabled():
+            logger.info("Enabling audit logs")
+            self.ping()
+            self.validate_config()
+        else:
+            logger.warn("NOT enabling audit logs")
```
Avantol13, line 23: `.warn` is deprecated I believe. you should use `.warning`

fence/resources/audit/client.py
```diff
@@ -0,0 +1,155 @@
...
+                return  # all good!
+            if t + 1 < max_tries:
+                self.logger.debug(f"Retrying... (got status code {r.status_code})")
+                time.sleep(wait_time)
+                wait_time *= 2
+        raise Exception(
+            f"Audit logs are enabled but audit-service is unreachable at {status_url}: {r.text}"
+        )
+
+    def validate_config(self):
```
Avantol13, line 54: Don't hate me. I'm tempted to ask if we get some docstrings for all these public functions. Granted I think most are self explanatory. But for consistency, eh :shrug:

fence/resources/audit/utils.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def is_audit_enabled(category=None):
+    enable_audit_logs = config["ENABLE_AUDIT_LOGS"] or {}
+    if category:
+        return enable_audit_logs and enable_audit_logs.get(category, False)
+    return enable_audit_logs and any(v for v in enable_audit_logs.values())
+
+
+def clean_request_url(request_url):
```
Avantol13, line 21: minor thing, but if this is only handling Authorization requests and only used in here, we maybe want to make this something like `_clean_authorization_request_url` so that people only think this is cleaning any request of sensitive data

fence/resources/audit/utils.py
```diff
@@ -0,0 +1,98 @@
...
+            request_url += f"?{flask.request.query_string.decode('utf-8')}"
+        request_url = clean_request_url(request_url)
+
+        if method == "GET" and endpoint.startswith("/data/download/"):
+            flask.current_app.audit_service_client.create_presigned_url_log(
+                status_code=response.status_code,
+                request_url=request_url,
+                guid=endpoint[len("/data/download/") :],
+                action="download",
+                **flask.g.audit_data,
```
Avantol13, line 60: do you need `if hasattr(flask.g, "audit_data"):` before this? Alternatively, you could use `getattr(flask.g, "audit_data", some_default)` I think

fence/resources/audit/utils.py
```diff
@@ -0,0 +1,98 @@
...
+    this code.
+    """
+    try:
+        method = flask.request.method
+        endpoint = flask.request.path
+        request_url = endpoint
+        if flask.request.query_string:
+            # could use `flask.request.url` but we don't want the root URL
+            request_url += f"?{flask.request.query_string.decode('utf-8')}"
+        request_url = clean_request_url(request_url)
```
Avantol13, line 52: do you only need this for logins? and do we only ever want to try to log if there's info in flask.g.audit_data ?  If so, you could move this into the /login block and do the flask.g.audit_data check before the endpoint blocks below

tests/test-fence-config.yaml
Avantol13: well just the boto call out to sqs would need to be mocked, you could check that it actually gets called with expected data

tests/test_audit_service.py
```diff
@@ -403,3 +485,100 @@ def test_login_log_login_endpoint(
...
+
+    # the audit-service client has already been loaded during the app
+    # init, so reload it with the new config
+    fence._setup_audit_service_client(app)
+
+    return mocked_sqs
+
+
+@pytest.mark.parametrize("indexd_client_with_arborist", ["s3_and_gs"], indirect=True)
+def test_push_to_audit_sqs(
```
Avantol13, line 529: I might have just missed it, but if you don't already have one for the /login audit log can you add a test for that too :pray: 
?

tests/test-fence-config.yaml
Avantol13: oh, I see you added some :confetti_ball:

fence/resources/audit/utils.py
```diff
@@ -0,0 +1,98 @@
...
+                    **flask.g.audit_data,
+                )
+    except:
+        traceback.print_exc()
+        logger.error(f"!!! Unable to create audit log! Returning response anyway...")
+
+    return response
+
+
+def enable_audit_logging(f):
```
Avantol13, line 76: :sparkles: neat :sparkles:

Avantol13: if I'm reading this correctly and I add this to some random function for an endpoint that's NOT /login or /data/download then nothing will happen, right? Basically it'll still get into `create_audit_log_for_request` but it'll just miss all the conditions and return. e.g. you'd have to add more code into that function to catch the request

fence/resources/audit/utils.py
paulineribeyre: I was thinking `clean_request_url` might be needed for other endpoints in the future 🤔 but yeah for now it could go into the login block

fence/resources/audit/utils.py
paulineribeyre: i won't use `if hasattr(flask.g, "audit_data")` because we should record requests that failed before `flask.g.audit_data` was populated. `getattr` is a good idea 😄

fence/resources/audit/utils.py
```diff
@@ -0,0 +1,102 @@
...
+                **audit_data,
+            )
+        elif method == "GET" and endpoint.startswith("/login/"):
+            request_url = _clean_authorization_request_url(request_url)
+            flask.current_app.audit_service_client.create_login_log(
+                status_code=response.status_code,
+                request_url=request_url,
+                **audit_data,
+            )
+    except:
```
Avantol13, line 67:
```suggestion
    except Exception:
```

Avantol13 (approved):

PR #924, opened 2021-05-27
fix: "key_ops" is an array, per RFC 7517 #4.3

m0nhawk:
### Bug Fixes
* Fixes the type of `key_ops` in JWK, it should be an array per [RFC 7517, #4.3](https://datatracker.ietf.org/doc/html/rfc7517#section-4.3):
>Its value is an array of key operation values.  Values defined by this specification are:


Avantol13 (approved):

m0nhawk:
Part of #948 now.

PR #926, opened 2021-06-03
Jira Ticket: [PXP-8121] (https://ctds-planx.atlassian.net/browse/PXP-8121)

nss10:
- When an Assume role is requested for a duration less than 900 seconds(AWS min assume role limit), this code automatically sets it to 900 seconds. 

### Bug Fixes
-- Added logic to update the DurationSeconds field right before the STS api call. 
-- Manually tested the issue after the fix
-- Added unit tests to verify the fix
-- Made sure no other unit tests are breaking. 

### Improvements
-- Replaced flask.current_app.config with the modern FenceConfig singleton in a couple of places

### Dependency updates
-- Also removed some dead code (mock_assume_role function which was patching the actual assume_role function in boto_manager.py)

### DIFF
--- a/fence/config-default.yaml
+++ b/fence/config-default.yaml
@@ -821,7 +821,7 @@ SYNAPSE_AUTHZ_TTL: 86400
 # then we can increase the amount of time that a session is valid for
 MAX_ROLE_SESSION_INCREASE: false
 ASSUME_ROLE_CACHE_SECONDS: 1800
-
+AWS_ASSUME_ROLE_MIN_LIMIT: 900
 # RAS refresh_tokens expire in 15 days
 RAS_REFRESH_EXPIRATION: 1296000
 # Number of projects that can be registered to a Google Service Accont
--- a/tests/data/test_data.py
+++ b/tests/data/test_data.py
@@ -1074,6 +1074,108 @@ def test_indexd_download_with_uploader_authorized(
     assert response.status_code == 200
 
 
+@pytest.mark.parametrize("test_max_role_session_increase", [True, False])
+@pytest.mark.parametrize("preset_expires_in", [100, 1000])
+def test_assume_role_time_limit(
+    client,
+    user_client,
+    kid,
+    rsa_private_key,
+    test_max_role_session_increase,
+    preset_expires_in,
+):
+    """
+    Test ``GET /data/download/1`` with authorized user (user is the uploader).
+    """
+
+    fence.S3IndexedFileLocation._assume_role_cache.clear()
+
+    current_max_role_session_increase = config["MAX_ROLE_SESSION_INCREASE"]
+    config.update(MAX_ROLE_SESSION_INCREASE=test_max_role_session_increase)
+
+    # Mocking STS client assume role
+    duration_in_function = 0
+    mocked_sts_client = MagicMock()
+
+    def mock_sts_client_assume_role(RoleArn, DurationSeconds, RoleSessionName=None):
+        nonlocal duration_in_function
+        duration_in_function = DurationSeconds
+        return {
+            "Credentials": {
+                "AccessKeyId": "",
+                "SecretAccessKey": "",
+                "SessionToken": "",
+                "Expiration": datetime.now() + timedelta(seconds=DurationSeconds),
+            },
+            "AssumedRoleUser": {"AssumedRoleId": "", "Arn": RoleArn},
+        }
+
+    mocked_sts_client.assume_role = mock_sts_client_assume_role
+    mocked_sts_client.return_value = mocked_sts_client
+    assume_role_patcher = patch(
+        "fence.resources.aws.boto_manager.client", mocked_sts_client
+    )
+    assume_role_patcher.start()
+    # Mocking indexd document
+    did = str(uuid.uuid4())
+    index_document = {
+        "did": did,
+        "baseid": "",
+        "uploader": user_client.username,
+        "rev": "",
+        "size": 10,
+        "file_name": "file1",
+        "urls": ["s3://bucket5/key-{}".format(did[:8])],
+        "acl": ["phs000178"],
+        "hashes": {},
+        "metadata": {},
+        "form": "",
+        "created_date": "",
+        "updated_date": "",
+    }
+    mock_index_document = mock.patch(
+        "fence.blueprints.data.indexd.IndexedFile.index_document", index_document
+    )
+    mock_index_document.start()
+
+    indexed_file_location = "s3"
+    path = "/data/download/1"
+    query_string = {"protocol": indexed_file_location, "expires_in": preset_expires_in}
+    headers = {
+        "Authorization": "Bearer "
+        + jwt.encode(
+            utils.authorized_download_context_claims(
+                user_client.username, user_client.user_id
+            ),
+            key=rsa_private_key,
+            headers={"kid": kid},
+            algorithm="RS256",
+        ).decode("utf-8")
+    }
+    response = client.get(path, headers=headers, query_string=query_string)
+
+    if config["MAX_ROLE_SESSION_INCREASE"]:
+        buffered_expires_in = preset_expires_in + int(
+            config["ASSUME_ROLE_CACHE_SECONDS"]
+        )
+    else:
+        buffered_expires_in = max(
+            preset_expires_in, int(config["AWS_ASSUME_ROLE_MIN_LIMIT"])
+        )
+
+    assert response.status_code == 200
+    assert duration_in_function == buffered_expires_in  # assume role duration
+    assert (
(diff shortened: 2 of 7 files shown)


tests/data/test_data.py
```diff
@@ -1074,6 +1074,108 @@ def test_indexd_download_with_uploader_authorized(
...
+def test_assume_role_time_limit(
+    client,
+    user_client,
+    kid,
+    rsa_private_key,
+    test_max_role_session_increase,
+    preset_expires_in,
+):
+    """
+    Test ``GET /data/download/1`` with authorized user (user is the uploader).
```
paulineribeyre, line 1088: update the docstring to explain what you're testing specifically

tests/data/test_data.py
```diff
@@ -1074,6 +1074,108 @@ def test_indexd_download_with_uploader_authorized(
...
+                "AccessKeyId": "",
+                "SecretAccessKey": "",
+                "SessionToken": "",
+                "Expiration": datetime.now() + timedelta(seconds=DurationSeconds),
+            },
+            "AssumedRoleUser": {"AssumedRoleId": "", "Arn": RoleArn},
+        }
+
+    mocked_sts_client.assume_role = mock_sts_client_assume_role
+    mocked_sts_client.return_value = mocked_sts_client
```
paulineribeyre, line 1114: 🤔  this doesn't look right haha - maybe you don't need this statement?

tests/data/test_data.py
```diff
@@ -1074,6 +1074,108 @@ def test_indexd_download_with_uploader_authorized(
...
+    test_max_role_session_increase,
+    preset_expires_in,
+):
+    """
+    Test ``GET /data/download/1`` with authorized user (user is the uploader).
+    """
+
+    fence.S3IndexedFileLocation._assume_role_cache.clear()
+
+    current_max_role_session_increase = config["MAX_ROLE_SESSION_INCREASE"]
```
paulineribeyre, line 1093: i don't think you need to test `MAX_ROLE_SESSION_INCREASE` as part of this `AWS_ASSUME_ROLE_MIN_LIMIT` unit test; i think it's complicating your test, and testing more than one thing in the same test. I would remove that part - but i could be wrong

tests/data/test_data.py
```diff
@@ -1074,6 +1074,108 @@ def test_indexd_download_with_uploader_authorized(
...
+        "file_name": "file1",
+        "urls": ["s3://bucket5/key-{}".format(did[:8])],
+        "acl": ["phs000178"],
+        "hashes": {},
+        "metadata": {},
+        "form": "",
+        "created_date": "",
+        "updated_date": "",
+    }
+    mock_index_document = mock.patch(
```
paulineribeyre, line 1136: i think there are existing fixtures you can use instead of mocking the indexd document manually - `indexd_client` and such

tests/data/test_data.py
```diff
@@ -1074,6 +1074,108 @@ def test_indexd_download_with_uploader_authorized(
...
 
+@pytest.mark.parametrize("test_max_role_session_increase", [True, False])
+@pytest.mark.parametrize("preset_expires_in", [100, 1000])
+def test_assume_role_time_limit(
+    client,
+    user_client,
+    kid,
+    rsa_private_key,
+    test_max_role_session_increase,
+    preset_expires_in,
```
paulineribeyre, line 1085: i would rename `preset_expires_in` => `presigned_url_expires_in`, again for clarity

tests/data/test_data.py
nss10: I think I need this. Because if that config value is changed to true, then the expires_in param is increased by 1800 seconds automatically. If I don't add it in the test. then the tests fail when the config value changes. Don't you think?

tests/data/test_data.py
nss10: return_value or the assume_role?

paulineribeyre: the whole line - you are doing `X.return_value = X`

tests/data/test_data.py
paulineribeyre: since `MAX_ROLE_SESSION_INCREASE` is `true` in the test config, i think i would just change it to `false` in your test, so you can test `AWS_ASSUME_ROLE_MIN_LIMIT` independently. But if you want to keep it that's fine

tests/data/test_data.py
```diff
@@ -1074,6 +1074,108 @@ def test_indexd_download_with_uploader_authorized(
...
+):
+    """
+    Test ``GET /data/download/1`` with authorized user (user is the uploader).
+    """
+
+    fence.S3IndexedFileLocation._assume_role_cache.clear()
+
+    current_max_role_session_increase = config["MAX_ROLE_SESSION_INCREASE"]
+    config.update(MAX_ROLE_SESSION_INCREASE=test_max_role_session_increase)
```
paulineribeyre, line 1094: you should use `monkeypatch.setitem(config, ...)` so you don't have to revert the change at the end of the test. in the current state, if the test fails halfway, the end of the test will not run and the change will not be reverted

tests/data/test_data.py
nss10: I need something from `bucket5` to test the `assume_role` function . The only indexd record I found for bucket5 in the existing fixtures is in the `unauthorized_indexd_client` which didn't make sense to me. So I followed this ad hoc approach just as in `test_assume_role_cache` unit test. 

What I can do is write a new fixture called `indexd_client_with_assume_role` and call it here. Do you this it is better than this?

tests/data/test_data.py
paulineribeyre: i see

tests/data/test_data.py
paulineribeyre: i see 👍

tests/data/test_data.py
paulineribeyre: Looks like you need to mock 2 things: the `boto3.client()` function (aka `fence.resources.aws.boto_manager.client`), which returns a Client instance, and the Client's `assume_role()` function. You are only creating 1 MagicMock for both, which is working because the functions that are getting called don't overlap, but it makes your test hard to read and IMO you should have 2 mocks

nss10: I need the `assume_role` function to have a body and not just a `Return value`. I need to identify what is the value of `DurationSeconds` in the mocked function ([in line 1102](https://github.com/uc-cdis/fence/blob/df04f49b03325aad873aafdf39cac6e7cbe628eb/tests/data/test_data.py#L1102) ). So, If I create a `MagicMock` for assume role like you've suggested, I won't be able to handle it in the `return_value`. Do you think it will be a good idea to use `side_effect` for the second mock (Magic Mock for the `assume_role` function) and assign the value of side_effect to be the mock function that I've defined in [line 1100](https://github.com/uc-cdis/fence/blob/df04f49b03325aad873aafdf39cac6e7cbe628eb/tests/data/test_data.py#L1100)?

fence/config-default.yaml
```diff
@@ -821,7 +821,6 @@ SYNAPSE_AUTHZ_TTL: 86400
 # then we can increase the amount of time that a session is valid for
 MAX_ROLE_SESSION_INCREASE: false
 ASSUME_ROLE_CACHE_SECONDS: 1800
-
```
paulineribeyre, line 824: this is nitpicky, but add the whitespace back to separate the sections clearly 😄

tests/data/test_data.py
```diff
@@ -1074,6 +1074,83 @@ def test_indexd_download_with_uploader_authorized(
...
+
+    AWS_ASSUME_ROLE_MIN_EXPIRATION = 900
+    if config["MAX_ROLE_SESSION_INCREASE"]:
+        buffered_expires_in = presigned_url_expires_in + int(
+            config["ASSUME_ROLE_CACHE_SECONDS"]
+        )
+    else:
+        buffered_expires_in = max(
+            presigned_url_expires_in, AWS_ASSUME_ROLE_MIN_EXPIRATION
+        )
```
paulineribeyre, line 1145: If I am understanding this properly, if `MAX_ROLE_SESSION_INCREASE` is set, there should still be a "max" logic. The current "+" logic only works here because we assume the `ASSUME_ROLE_CACHE_SECONDS == 1800` value won't change.
I think it would be something like:
```
if config["MAX_ROLE_SESSION_INCREASE"]:
    presigned_url_expires_in += config["ASSUME_ROLE_CACHE_SECONDS"]
buffered_expires_in = max(presigned_url_expires_in, AWS_ASSUME_ROLE_MIN_EXPIRATION)
```

paulineribeyre (approved):

paulineribeyre (approved):

PR #928, opened 2021-06-04
(PXP-7717) Implement ad-hoc usersync for single user after logging in through RAS

BinamB:

### New Features
- Added functionality to allow clients to parse visas with `parse_visas` parameter in the `/authorize` endpoint

### Improvements
- Optimized visa parsing logic for faster ad-hoc visa parsing


fence/blueprints/login/ras.py
```diff
@@ -90,15 +93,25 @@ def post_login(self, user=None, token_result=None):
             user=user, refresh_token=refresh_token, expires=expires + issued_time
         )
 
+        parse_visas = (
```
Avantol13, line 96: can't you just do `parse_visas = query_params.get("parse_visas")` ? it's presence indicates to do it, doesn't matter what the value is

Avantol13: rather, any value provided would evaluate as true

fence/blueprints/login/ras.py
```diff
@@ -90,15 +93,25 @@ def post_login(self, user=None, token_result=None):
...
+            strtobool(query_params.get("parse_visas")[0])
+            if query_params.get("parse_visas")
+            else False
+        )
         usersync = config.get("USERSYNC", {})
         sync_from_visas = usersync.get("sync_from_visas", False)
         # Check if user has any project_access from a previous session or from usersync AND if fence is configured to use visas as authZ source
         # if not do an on-the-fly usersync for this user to give them instant access after logging in through RAS
-        if not user.project_access and sync_from_visas:
+        if not user.project_access and sync_from_visas and parse_visas:
```
Avantol13, line 105: we may want a cfg to override for clients? trying to think this through; I'm wondering if there are cases where we want to disallow this or only allow this

Avantol13: I think we do want a global override. `PARSE_VISAS_ON_LOGIN:` default is None which means, use whatever the client sends. But you can also supply True or False to globally override anything clients send

Avantol13: that way, if a deployment wants to say  "we will always do this no matter what" there's a way to change it easily

fence/sync/sync_users.py
```diff
@@ -1621,26 +1625,32 @@ def _update_authz_in_arborist(self, session, user_projects, user_yaml=None):
...
-            #      but sync should continue from this point so there are no partial
-            #      updates
-            self.logger.warning(
-                "Could not get list of users in Arborist, continuing anyway. "
-                "WARNING: this sync will NOT remove access for users no longer in "
-                f"authorization sources. Error: {error}"
-            )
+        if not single_user_sync:
+            try:
+                # TODO: if single visa sync then just get that one user. Pass user as a field here.
```
Avantol13, line 1630: is this still TODO in this PR?

fence/sync/sync_users.py
```diff
@@ -1584,7 +1586,9 @@ def _update_arborist(self, session, user_yaml):
 
         return True
 
-    def _update_authz_in_arborist(self, session, user_projects, user_yaml=None):
+    def _update_authz_in_arborist(
+        self, session, user_projects, user_yaml=None, single_user_sync=False
```
Avantol13, line 1590: I don't see where `single_user_sync` actually gets passed in as true to this

fence/sync/sync_users.py
```diff
@@ -2126,9 +2139,25 @@ def sync_single_user_visas(self, user, sess=None):
...
+
         # update fence db
         if user_projects:
             self.logger.info("Sync to db and storage backend")
             self.sync_to_db_and_storage_backend(user_projects, user_info, sess, True)
         else:
             self.logger.info("No users for syncing")
+
+        # update arborist db (user access)
+        if self.arborist_client:
```
Avantol13, line 2151: what's the else case here? we should probably log an error that arborist client wasn't setup or some other logic?

fence/sync/sync_users.py
```diff
@@ -2126,9 +2139,25 @@ def sync_single_user_visas(self, user, sess=None):
...
+            self.logger.info("Synchronizing arborist with authorization info...")
+            success = self._update_authz_in_arborist(
+                sess, user_projects, user_yaml, True
```
Avantol13, line 2154: this is mostly just a habit of mine, but if there are any optional params I try to have them explicit here to avoid shuffling later. e.g. `_update_authz_in_arborist(sess, user_projects, user_yaml=user_yaml, single_user_sync=True)` up to you though

pyproject.toml
```diff
@@ -27,7 +27,7 @@ flask-cors = "^3.0.3"
 flask-restful = "^0.3.6"
 flask_sqlalchemy_session = "^1.1"
 email_validator = "^1.1.1"
-gen3authz = "^1.0"
+gen3authz = { path = "./pkg/gen3authz-1.1.0.tar.gz" }
```
Avantol13, line 30: I assume this needs to change back or to `^1.1`?

Avantol13: wait, why are we adding pkg's?

fence/sync/sync_users.py
BinamB: Oh my bad. Removing this. This isn't required anymore and single-user project syncing is handled by `user_projects` containing just that one user and their projects

pyproject.toml
BinamB: Yup! Was just waiting for gen3authz stuff to be merged

fence/blueprints/login/ras.py
BinamB: oh okay yeah i think that makes sense. So,
None: Allow per client 
True: Parse for all clients
False: Parse for no clients
?

fence/blueprints/login/ras.py
BinamB: Okay yeah that makes sense. I guess people wont be going out of their way to write `parse_visas=False` :sweat_smile:

fence/blueprints/login/ras.py
Avantol13: Right, exactly

pyproject.toml
Avantol13: gotcha gotcha, we should merge that first and update this :+1:

fence/config-default.yaml
```diff
@@ -843,6 +843,13 @@ GA4GH_VISA_ISSUER_ALLOWLIST:
   - 'https://stsstg.nih.gov'
 # Number of projects that can be registered to a Google Service Accont
 SERVICE_ACCOUNT_LIMIT: 6
+# Global sync visas during login
+#----------------------------------------------------------------
+# None: Allow per client
+# True: Parse for all clients
+# False: Parse for no clients
+# GLOBAL_PARSE_VISAS_ON_LOGIN:
```
Avantol13, line 851: I think you need to still have it uncommented like `GLOBAL_PARSE_VISAS_ON_LOGIN: `

Avantol13: cfg should interpret that as `None` and then you check for it later to default

fence/blueprints/login/ras.py
```diff
@@ -133,29 +133,40 @@ def post_login(self, user=None, token_result=None):
...
-            if not isinstance(dbGaP, list):
-                dbGaP = [dbGaP]
-
-            sync = init_syncer(
-                dbGaP,
-                None,
-                DB,
-            )
-            sync.sync_single_user_visas(user, current_session)
+        GLOBAL_PARSE_VISAS_ON_LOGIN = config.get("GLOBAL_PARSE_VISAS_ON_LOGIN", None)
```
Avantol13, line 136: if you define this globally as None, you can read this like any other cfg w/o .get(), e.g. `config["GLOBAL_PARSE_VISAS_ON_LOGIN"]`

fence/config-default.yaml
BinamB: Okay yeah didnt know if it would interpret as `None` or `false`. I'll change that

fence/config-default.yaml
```diff
@@ -843,6 +843,13 @@ GA4GH_VISA_ISSUER_ALLOWLIST:
   - 'https://stsstg.nih.gov'
 # Number of projects that can be registered to a Google Service Accont
 SERVICE_ACCOUNT_LIMIT: 6
+# Global sync visas during login
+#----------------------------------------------------------------
+# None: Allow per client
+# True: Parse for all clients
+# False: Parse for no clients
+GLOBAL_PARSE_VISAS_ON_LOGIN:
+#----------------------------------------------------------------
```
paulineribeyre, line 852: maybe a longer explanation than `Global sync visas during login`, and replace `----` with whitespace

paulineribeyre: mention the `parse_visas` query arg in this longer comment, especially since there's n o swagger doc for the RAS login endpoint

fence/blueprints/login/ras.py
```diff
@@ -133,29 +133,40 @@ def post_login(self, user=None, token_result=None):
...
+            ):
+                # Close previous db sessions. Leaving it open causes a race condition where we're viewing user.project_access while trying to update it in usersync
+                # not closing leads to partially updated records
+                current_session.close()
+                DB = os.environ.get("FENCE_DB") or config.get("DB")
+                arborist = ArboristClient(
+                    arborist_base_url=config["ARBORIST"],
+                    logger=get_logger("user_syncer.arborist_client"),
+                    authz_provider="user-sync",
+                )
```
paulineribeyre, line 154: for clarity you might want to move this `arborist =` block to after the `DB` logic instead of in the middle of it

fence/blueprints/login/ras.py
```diff
@@ -133,29 +133,40 @@ def post_login(self, user=None, token_result=None):
             user=user, refresh_token=refresh_token, expires=expires + issued_time
         )
 
-        usersync = config.get("USERSYNC", {})
-        sync_from_visas = usersync.get("sync_from_visas", False)
-        # Check if user has any project_access from a previous session or from usersync AND if fence is configured to use visas as authZ source
-        # if not do an on-the-fly usersync for this user to give them instant access after logging in through RAS
-        if not user.project_access and sync_from_visas:
```
paulineribeyre, line 140: shouldn't we keep checking if `sync_from_visas` is enabled?

fence/blueprints/login/ras.py
```diff
@@ -133,29 +133,40 @@ def post_login(self, user=None, token_result=None):
...
-            )
-            sync.sync_single_user_visas(user, current_session)
+        global_parse_visas_on_login = config["GLOBAL_PARSE_VISAS_ON_LOGIN"]
+
+        if global_parse_visas_on_login or global_parse_visas_on_login == None:
+            parse_visas = query_params.get("parse_visas")
+            # Check if user has any project_access from a previous session or from usersync AND if fence is configured to use visas as authZ source
+            # if not do an on-the-fly usersync for this user to give them instant access after logging in through RAS
+            # If GLOBAL_PARSE_VISAS_ON_LOGIN is true then we want to run it regardless of whether or not the client sent parse_visas on request
+            if (global_parse_visas_on_login or parse_visas) and (
```
paulineribeyre, line 143: i think you could reduce the indentation by using a single `if` to make the logic look less complicated
```
sync_from_visas = ...
global_parse_visas_on_login = ...
parse_visas = global_parse_visas_on_login or
    (global_parse_visas_on_login == None and
    <however you choose to handle `parse_visas` query param>)
if sync_from_visas and parse_visas and not user.project_access:
```
\+ the existing comments of course

fence/sync/sync_users.py
paulineribeyre: if you want to be nice you can add this log to the `_sync` and `_sync_visas` functions that have the same logic 😄

fence/sync/sync_users.py
```diff
@@ -1692,7 +1699,19 @@ def _update_authz_in_arborist(self, session, user_projects, user_yaml=None):
...
+                            if policy_id not in self._created_policies:
+                                policy_json = {
+                                    "id": policy_id,
+                                    "description": "policy created by fence sync",
+                                    "role_ids": [permission],
+                                    "resource_paths": [path],
+                                }
+                                policies.append(policy_json)
+                        else:
                             try:
```
paulineribeyre, line 1715: you removed the logic of only updating policies if they are not in `self._created_policies` (for non-single-user sync). Is that on purpose? if not, i think your if statements should look like this:
```
if not single_user_sync and policy_id not in self._created_policies:
    try:
        self.arborist_client.update_policy(
        ...
    except ...
    self._created_policies.add(policy_id)

if single_user_sync:
    policy_id_list.append(policy_id)
    policy_json = ...
    policies.append(policy_json)

    try:
        self.arborist_client.update_bulk_policy(policies)
        ...
```

fence/sync/sync_users.py
```diff
@@ -1709,7 +1728,24 @@ def _update_authz_in_arborist(self, session, user_projects, user_yaml=None):
                                 )
                             self._created_policies.add(policy_id)
 
-                        self.arborist_client.grant_user_policy(username, policy_id)
+                        self.arborist_client.grant_user_policy(
+                            username, policy_id
+                        ) if not single_user_sync else None
```
paulineribeyre, line 1733: this can be in the `if` block above, instead of `if not single_user_sync else None`

fence/sync/sync_users.py
```diff
@@ -1709,7 +1728,24 @@ def _update_authz_in_arborist(self, session, user_projects, user_yaml=None):
...
+
+            if single_user_sync:
+                try:
+                    self.arborist_client.update_bulk_policy(policies)
+                    for policy_id in policy_id_list:
+                        self._created_policies.add(policy_id)
```
paulineribeyre, line 1739:
```suggestion
                    self._created_policies.update(policy_id_list)
```

paulineribeyre: if i am reading the code correctly, you actually don't need to maintain `self._created_policies` since you are creating in bulk.

Unless you add a bit of logic to update in bulk `policies - self._created_policies` 😄  And it should be faster if you don't add to the bulk creation the policies you already updated

fence/blueprints/login/ras.py
```diff
@@ -133,29 +133,40 @@ def post_login(self, user=None, token_result=None):
...
-            sync = init_syncer(
-                dbGaP,
-                None,
-                DB,
-            )
-            sync.sync_single_user_visas(user, current_session)
+        global_parse_visas_on_login = config["GLOBAL_PARSE_VISAS_ON_LOGIN"]
+
+        if global_parse_visas_on_login or global_parse_visas_on_login == None:
+            parse_visas = query_params.get("parse_visas")
```
paulineribeyre, line 139:
```suggestion
            parse_visas = "parse_visas" in query_params
```
i think right now just `?parse_visas` (no value) would be interpreted as False by `if parse_visas`. Since there is no documentation for this param yet, i'm not sure that's on purpose.

Also I kinda disagree with your and @Avantol13's discussion above - i think if we expect a value, we should handle `?parse_visas=false` properly.

tldr; choose whether we accept `?parse_visas`, or `?parse_visas=true/false`, or both, and update the doc and the code accordingly

fence/blueprints/login/ras.py
BinamB: I guess my argument was if we have `GLOBAL_PARSE_VISAS_ON_LOGIN` we dont really need the `sync_from_visas` since thats mostly just for the hourly usersync. I guess you can argue what if we dont want to sync from visas but someone turns on `GLOBAL_PARSE_VISAS_ON_LOGIN` or someone does `parse_visas=True` during login. I guess I would say thats a good argument. Fine you've convinced me. Adding it back

fence/sync/sync_users.py
BinamB: Will I be nice though? :smiling_imp:

fence/sync/sync_users.py
BinamB: That is correct. Can you see my inexperience shine. sparkles

fence/blueprints/login/ras.py
paulineribeyre: you made part of the changes but you still need to remove the `if` statement on line 153, and to move all the details comments to line 139

fence/sync/sync_users.py
```diff
@@ -1692,24 +1701,56 @@ def _update_authz_in_arborist(self, session, user_projects, user_yaml=None):
...
-                                self.logger.info(
-                                    "not creating policy in arborist; {}".format(str(e))
-                                )
-                            self._created_policies.add(policy_id)
 
-                        self.arborist_client.grant_user_policy(username, policy_id)
+                        # if (
+                        #     not single_user_sync
+                        #     and policy_id not in self._created_policies
+                        # ):
```
paulineribeyre, line 1708: remove these

fence/sync/sync_users.py
```diff
@@ -1692,24 +1701,56 @@ def _update_authz_in_arborist(self, session, user_projects, user_yaml=None):
...
+                                            str(e)
+                                        )
+                                    )
+                                self._created_policies.add(policy_id)
+                            self.arborist_client.grant_user_policy(username, policy_id)
+
+                        if single_user_sync:
+                            policy_id_list.append(policy_id)
+
+                            if policy_id not in self._created_policies:
```
paulineribeyre, line 1733: can remove this `if` since we don't maintain `_created_policies` for `single_user_sync`

paulineribeyre (approved):
🎉

PR #933, opened 2021-06-09
(PXP-6339): RAS visa validation

vpsx:
Jira Ticket: [PXP-6339](https://ctds-planx.atlassian.net/browse/PXP-6339)

minor version bump to 5.2.0

### New Features
- Validate RAS visas when they are fetched upon RAS login
- Validate RAS visas when they are updated during visa update cronjob
- Visas are validated according to the GA4GH AAI spec--see https://github.com/ga4gh/data-security/blob/master/AAI/AAIConnectProfile.md#embedded-access-token-format 

### Deployment changes
- By default Fence will allow only visas issued by RAS staging, RAS prod, and itself; if a different set of allowed visa issuers is needed, edit the GA4GH_VISA_ISSUER_ALLOWLIST config value. Otherwise, no action needed.


fence/resources/openid/ras_oauth2.py
```diff
@@ -141,23 +153,130 @@ def update_user_visas(self, user, db_session=current_session):
...
+                visa_kid = get_kid(encoded_visa)
+            except Exception as e:
+                self.logger.error(
+                    "Could not get issuer or kid from visa: {}. Discarding visa.".format(
+                        e
+                    )
+                )
+                continue  # Not raise: If visa malformed, does not make sense to retry
+
+            # See if pkey is in cronjob cache; if not, update cache.
```
vpsx, line 166: Note: Most of this code is very similar to the code from [this recent authutils commit](https://github.com/uc-cdis/authutils/pull/52/files#diff-c4ce378d338b1ee760432e2779c2608873a47a826c516e1ca9b06df551b4f469). 

The authutils utils rely on keeping a public key cache in the flask app context, and the cronjob runs without one, so this code is here.

The plan is to implement a more persistent cache than this, so here I have taken the simplest and dumbest approach of just repeating much of the authutils code instead of refactoring authutils.

fence/blueprints/login/ras.py
```diff
@@ -45,13 +51,50 @@ def post_login(self, user=None, token_result=None):
...
+                )
+                continue
+
+            try:
+                # Validate the visa per GA4GH AAI "Embedded access token" format rules.
+                # pyjwt also validates signature and expiration.
+                decoded_visa = validate_jwt(
+                    encoded_visa,
+                    public_key,
+                    # Embedded token must not contain aud claim
```
Avantol13, line 74: is this based on the AAI spec?

tests/ras/test_ras.py
```diff
@@ -134,7 +134,7 @@ def test_update_visa_token(
         },
     }
 
-    headers = {"kid": kid_2}
+    headers = {"kid": kid}
```
vpsx, line 137: Note: I couldn't tell if there was a particular reason to be using a different key for the second visa, so figured I'd just use the same one to reduce noise/readability and make it more evident what is being tested. But if I simply missed something, I can definitely switch this back.

Same goes for the rest of this file!

fence/blueprints/login/ras.py
vpsx: Yup! https://github.com/ga4gh/data-security/blob/master/AAI/AAIConnectProfile.md#embedded-access-token-format `The payload claims MUST NOT include aud.`

(dunno how to do github permalinks with markdown files but for posterity: commit  da56838 )

fence/config-default.yaml
```diff
@@ -824,6 +824,9 @@ ASSUME_ROLE_CACHE_SECONDS: 1800
 
 # RAS refresh_tokens expire in 15 days
 RAS_REFRESH_EXPIRATION: 1296000
+# List of JWT issuers from which Fence will accept GA4GH visas
+GA4GH_VISA_ISSUER_WHITELIST: []
```
Avantol13, line 828: can this populate itself with the "issuers" from well-known for configured logins on startup? might be overkill, but would be really convenient to not have to keep this up to date. Or, maybe even just adding the RAS issuer and ourselves into the default :man_shrugging: ?

Avantol13: https://github.com/uc-cdis/fence/blob/d75257ef0faadf8d5b0e02ae9063ea3506628c09/fence/config.py#L18 post process of cfg could maybe shoot out requests to get issuers and populate this, idk, maybe that is overkill. Not all IDPs are GA4GH brokers. But I think "hard coding" RAS and us in the default would make sense and be more convenient for deployments

Avantol13 (changes requested):
see comments

fence/config-default.yaml
vpsx: Oh I figured more explicit was better for a visa whitelist but I can look into this--do you mean like, on app startup, maybe during the client init, hit all of the configured idps' well-known endpoints and add(?) the "issuer" values to this config value? 

Otherwise if taking the second approach--that sounds fine; would you use `stsstg` or prod RAS or both?

Avantol13: Agreed, very good points. For the second approach, we should probably add both? whatever issuers RAS has for staging and prod

Avantol13: Oh! Minor suggestion, I've been trying to move towards "ALLOWLIST" vs "WHITELIST" but up to you

vpsx: Word! Thank you 🙏

PR #934, opened 2021-06-10
feat(email): populate email from IdPs, update deprecated .warn to .wa…

Avantol13:
…rning

### New Features
- Try to populate a user's email in the user db if available from IdP 

### Improvements
- change deprecated .warn to new .warning


fence/auth.py
```diff
@@ -106,6 +112,12 @@ def set_flask_session_values(user):
     if not idp:
         idp = IdentityProvider(name=provider)
 
+    if email and user.email != email:
+        logger.info(
+            f"Updating username {user.username}'s email from {user.email} to {email}"
+        )
+        user.email = email
+
```
vpsx, line 120: If a user already exists in the db and they already have a linked matching identity provider (so, I think, for most cases where a user had logged in before this fence commit), this code won't run, even if the email and the user.email don't match--right? 🧐

Avantol13: AhH! :sweat_smile: You're right, I missed the early return, will adjust. Good eye! :detective:

jawadqur:
I deployed this branch to https://fence.planx-pla.net for testing, will let you know if it's successful.

fence/blueprints/login/base.py
```diff
@@ -97,8 +100,9 @@ def get(self):
         code = flask.request.args.get("code")
         result = self.client.get_user_id(code)
         username = result.get(self.username_field)
+        email = result.get(self.email_field)
```
vpsx, line 103: So if I'm reading correctly, this happens to work in the google case because the `result` happens to contain `email` -- because when idp is google we use `email` as `username`. But in general if the get_user_id code for a given client is not changed so that the returned `result` has the email field, this part will not work? So for example RAS logins will still not populate the email? I might just be confused on the intended behavior.

fence/auth.py
```diff
@@ -249,3 +259,16 @@ def wrapper(*args, **kwargs):
 def admin_login_required(function):
     """Compose the login required and admin required decorators."""
     return login_required({"admin"})(admin_required(function))
+
+
+def _update_users_email(user, email):
+    """
+    Update email if provided and doesn't match db entry.
+
+    NOTE: This does NOT commit to the db, do so outside this function
```
vpsx, line 268: Oof I think in the case where a user already exists, the early return will now prevent the db commit from happening 😅

fence/auth.py
vpsx: Yay! One more thing, I think with the new call to _update_users_email up top, this second call is unnecessary

vpsx (commented):
Just flagging a couple possible kinks :pray:

vpsx (approved):
LGTM! 🎉 ✉️

PR #936, opened 2021-06-21
fix(fence-idp-jwt-val): Patch to expect old style tokens

vpsx:
Jira Ticket: n/a :fire:

### Bug Fixes
- patch to accept old style tokens from fence idp only, since fenceshib cannot be updated to issue new style tokens

### DIFF
--- a/fence/blueprints/login/fence_login.py
+++ b/fence/blueprints/login/fence_login.py
@@ -91,6 +91,14 @@ class FenceCallback(DefaultOAuth2Callback):
         tokens = flask.current_app.fence_client.fetch_access_token(
             redirect_uri, **flask.request.args.to_dict()
         )
+        # After Fence 5.0.0 "remove scopes from aud claim" changes,
+        # this validate_jwt is supposed to look like this:
+        #     id_token_claims = validate_jwt(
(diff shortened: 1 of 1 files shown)


Avantol13 (approved):

fence/blueprints/login/fence_login.py
```diff
@@ -91,8 +91,20 @@ def get(self):
         tokens = flask.current_app.fence_client.fetch_access_token(
             redirect_uri, **flask.request.args.to_dict()
         )
+        # After Fence 5.0.0 "remove scopes from aud claim" changes,
+        # this validate_jwt is supposed to look like this:
+        #     id_token_claims = validate_jwt(
+        #         tokens["id_token"], scope="openid", purpose="id", attempt_refresh=True
```
paulineribeyre, line 97:
```suggestion
        #         tokens["id_token"], scope={"openid"}, purpose="id", attempt_refresh=True
```
should it be that? or i am focusing too much on past issues? 😆

vpsx: DANGIT no you're right ahahahahahaaha thank you boss

PR #939, opened 2021-06-24
PXP-2240: Add monitoring SA email in error msg

MaribelleHGomez:
Jira Ticket: [PXP-2240](https://ctds-planx.atlassian.net/browse/PXP-2240)

### New Features

- email of monitor service account added to error msg
- moved get_monitoring_service_account_response func out of class to be used globally

### Improvements
Change warn to warning in logger due to dep warning


fence/blueprints/google.py
```diff
@@ -807,23 +810,33 @@ def _get_service_account_error_status(sa):
         response["errors"]["google_project_id"]["error"]
         == ValidationErrors.MONITOR_NOT_FOUND
     ):
+        monitor_account = (
+            get_monitoring_service_account_response()[0]["service_account_email"]
+            if (get_monitoring_service_account_response()[1] == 200)
+            else ""
```
BinamB, line 816: So I wouldn't do `else ""` in the case where the response is not 200. This would push an empty string to the error message below which is pretty vague.

MaribelleHGomez: that's how the message was before I made changes, what would you suggest there

fence/blueprints/google.py
```diff
@@ -807,23 +810,33 @@ def _get_service_account_error_status(sa):
         response["errors"]["google_project_id"]["error"]
         == ValidationErrors.MONITOR_NOT_FOUND
     ):
+        monitor_account = (
+            get_monitoring_service_account_response()[0]["service_account_email"]
```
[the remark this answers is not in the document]
MaribelleHGomez: lol yes

fence/blueprints/google.py
MaribelleHGomez: I would say the message as is when there is no sa found in the project seems fine so the "" is good? else it return the sa it has found on the project

BinamB (approved):
Really trusting you here. lgtm! :rofl:

PR #941, opened 2021-06-28
chore(pybase3): Update pybase img de novo

themarcelor:
Suppressing UWSGI & NGINX metrics in favour of DataDog’s telemetry / python APM / probing -- hoping we can achieve the same observability with it.

### Improvements
- Once we retire the k8s sidecar containers from fence's k8s YAML (here: https://github.com/uc-cdis/cloud-automation/pull/1643), old fence images will face a lot of log noise as the HTTP metrics endpoints will no longer exist.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "fence"
-version = "5.2.0"
+version = "5.3.1"
 description = "Gen3 AuthN/AuthZ OIDC Service"
 authors = ["CTDS UChicago <cdis@uchicago.edu>"]
 license = "Apache-2.0"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "fence"
-version = "5.2.0"
+version = "5.3.1"
```
paulineribeyre, line 3: might be 5.3.2, i see a 5.3.1 now

paulineribeyre (approved):

pyproject.toml
themarcelor: Yeah, as per the thread on Slack. I guess the semantic version tracking in pyproject.toml is necessarily critical 🤷🏼

PR #948, opened 2021-07-20
Feat/atlas hacks

rnerella92:
Jira Ticket: [PXP-7690](https://ctds-planx.atlassian.net/browse/PXP-7690)

Setup Atlas login through Fence with OIDC, it will provide more flexible authentication flow, in contrast of current all-or-nothing approach.

### Bug Fixes
PXP-7690


.secrets.baseline
```diff
@@ -56,167 +60,224 @@
...
-        "type": "Basic Auth Credentials",
-        "filename": "fence/config-default.yaml",
         "hashed_secret": "a94a8fe5ccb19ba61c4c0873d391e987982fbbd3",
         "is_verified": false,
-        "line_number": 31
+        "line_number": 31,
+        "type": "Basic Auth Credentials"
       },
       {
+<<<<<<< HEAD
```
paulineribeyre, line 85: lots of conflicts in this file - you need to re-build the lock file

fence/blueprints/user.py
```diff
@@ -24,6 +24,11 @@ def user_info():
         client_id = current_token["azp"]
     info = get_current_user_info()
     info["azp"] = client_id
+
+    # Atlas hack: Pretend this is actually a string.
+    # (sub is supposed to be string per the standard)
+    info["sub"] = str(info["sub"])
```
paulineribeyre, line 30: why is this an "atlas hack"?
and if we need to do this, let's do it in function `get_user_info` instead. You might need to update the API docs for this endpoint

fence/jwt/keys.py
```diff
@@ -188,7 +188,7 @@ def public_key_to_jwk(self):
             except AttributeError:
                 # there is no need to decode values that are already strings
                 pass
-        jwk_dict.update({"use": "sig", "key_ops": "verify", "kid": self.kid})
+        jwk_dict.update({"use": "sig", "key_ops": ["verify"], "kid": self.kid})
```
paulineribeyre, line 191: could you explain this change? and again, the API docs for this endpoint might need an update

fence/jwt/token.py
```diff
@@ -396,6 +396,13 @@ def generate_signed_access_token(
                 " running outside of flask application"
             )
 
+    audiences = []
+    if client_id:
+        audiences.append(client_id)
+    # append scopes for backwards compatibility
+    # eventual goal is to remove scopes from `aud`
+    audiences = audiences + scopes
```
paulineribeyre, line 404: `audiences` is never used. could you explain this change?

.secrets.baseline
rnerella92: Updated the secrets file.

fence/blueprints/user.py
m0nhawk: This is a non-complaint to the standard, per [OpenID](https://openid.net/specs/openid-connect-basic-1_0.html#rfc.section.2.2):
```
The sub value is a case sensitive string
```

The comment is redundant. :)

fence/jwt/keys.py
m0nhawk: Again, this is non-complaint with the standard — [RFC7517](https://datatracker.ietf.org/doc/html/rfc7517#section-4.3):

```
Its value is an array of key operation values.
```

fence/jwt/token.py
m0nhawk: I'm not sure on this one. It looks like it's safe to remove.

m0nhawk (changes requested):
@rnerella92 Some extra comments.

fence/blueprints/user.py
Avantol13: I believe the suggestion was to do it here instead: https://github.com/uc-cdis/fence/blob/e2779111dce9a5db4de850a0ecc785373597b70b/fence/resources/user/__init__.py#L80 as this is the most low level support for constructing userinfo

fence/jwt/token.py
Avantol13: Let's remove it then

fence/blueprints/user.py
m0nhawk: @rnerella92 and did you commit this change in the place Alex suggested?

m0nhawk:
@Avantol13 Can you take a look again if everything is addressed?

Avantol13 (approved):

m0nhawk (approved):

PR #949, opened 2021-07-20
PXP-8318 Split public and private audit configs

paulineribeyre:
Jira Ticket: [PXP-8318](https://ctds-planx.atlassian.net/browse/PXP-8318)

### New Features
- Split the private fields from the public fields in the audit logging configuration

### Improvements
- When deciding whether to sign a URL for download, do not rely on "acl" if the record has an "authz"

### DIFF
--- a/fence/resources/audit/client.py
+++ b/fence/resources/audit/client.py
@@ -38,11 +38,27 @@ class AuditServiceClient:
 
         if self.push_type == "aws_sqs":
             aws_sqs_config = config["PUSH_AUDIT_LOGS_CONFIG"]["aws_sqs_config"]
+            # we know the cred is in AWS_CREDENTIALS (see `_check_aws_creds_and_region`)
+            aws_creds = (
+                config.get("AWS_CREDENTIALS", {})[aws_sqs_config["aws_cred"]]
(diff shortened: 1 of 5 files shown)


fence/resources/audit/client.py
```diff
@@ -38,11 +38,27 @@ def __init__(self, service_url, logger):
...
+            aws_creds = (
+                config.get("AWS_CREDENTIALS", {})[aws_sqs_config["aws_cred"]]
+                if "aws_cred" in aws_sqs_config
+                else {}
+            )
+            if (
+                not aws_creds
+                and "aws_access_key_id" in aws_sqs_config
+                and "aws_secret_access_key" in aws_sqs_config
+            ):
```
BinamB, line 51: What if none of these are set up. Do we want to log something like `hey! no aws_creds found`

BinamB (approved):

fence/resources/audit/client.py
paulineribeyre: no, the AWS creds are optional, for example in our cloud-automation Fence already has a role which has access to the SQS, so we don't use it

BinamB: Got it

PR #952, opened 2021-08-09
Well known endpoint: issuer is always base URL

paulineribeyre:
### Improvements
- Fix the `/.well-known/openid-configuration` endpoint to always return the environment's base URL in the `issuer` field, instead of returning the upstream Fence IDP's base URL


BinamB (approved):
lgtm!!!

PR #956, opened 2021-08-19
(PXP-8510): Use RAS v1.1 passport instead of v1.0

BinamB:
Currently, not creating Gen3 Passports or making them available anywhere. Originally made the function to repackage RAS visas into a Gen3 Passport but due to RAS policies we aren't allowed to do that anymore. Still keeping it here so that we could repurpose this later when we need to package our own Gen3 Visas.

### New Features
- Use RAS V1.1 Passport instead of RAS v1.0 passport
- Added logic to create Gen3 Passports

### DIFF
--- a/fence/blueprints/login/ras.py
+++ b/fence/blueprints/login/ras.py
@@ -46,13 +46,21 @@ class RASCallback(DefaultOAuth2Callback):
         # TODO: This is not IdP-specific and will need a rethink when
         # we have multiple IdPs
         user.ga4gh_visas_v1 = []
+
         current_session.commit()
 
+        decoded_passport = None
+        encoded_visas = []
         encoded_passport = flask.g.userinfo.get("passport_jwt_v11")
-        decoded_passport = jwt.decode(encoded_passport, verify=False)
 
-        # encoded_visas = flask.g.userinfo.get("ga4gh_passport_v1", [])
-        encoded_visas = decoded_passport.get("ga4gh_passport_v1", [])
+        print("_-----------------------------")
+        print(encoded_passport)
+        print("_-----------------------------")
+
+
+        if encoded_passport:
+            decoded_passport = jwt.decode(encoded_passport, verify=False)
+            encoded_visas = decoded_passport.get("ga4gh_passport_v1", [])
 
         for encoded_visa in encoded_visas:
             try:
--- a/fence/blueprints/login/ras.py
+++ b/fence/blueprints/login/ras.py
@@ -53,11 +53,6 @@ class RASCallback(DefaultOAuth2Callback):
         encoded_visas = []
         encoded_passport = flask.g.userinfo.get("passport_jwt_v11")
 
-        print("_-----------------------------")
-        print(encoded_passport)
-        print("_-----------------------------")
-
-
         if encoded_passport:
             decoded_passport = jwt.decode(encoded_passport, verify=False)
             encoded_visas = decoded_passport.get("ga4gh_passport_v1", [])
--- a/fence/blueprints/login/ras.py
+++ b/fence/blueprints/login/ras.py
@@ -48,7 +48,11 @@ class RASCallback(DefaultOAuth2Callback):
         user.ga4gh_visas_v1 = []
         current_session.commit()
 
-        encoded_visas = flask.g.userinfo.get("ga4gh_passport_v1", [])
+        encoded_passport = flask.g.userinfo.get("passport_jwt_v11")
+        decoded_passport = jwt.decode(encoded_passport, verify=False)
+
+        # encoded_visas = flask.g.userinfo.get("ga4gh_passport_v1", [])
+        encoded_visas = decoded_passport.get("ga4gh_passport_v1", [])
 
         for encoded_visa in encoded_visas:
             try:
--- a/fence/config-default.yaml
+++ b/fence/config-default.yaml
@@ -856,3 +856,4 @@ USERSYNC:
   fallback_to_dbgap_sftp: false
   visa_types:
     ras: [https://ras.nih.gov/visas/v1, https://ras.nih.gov/visas/v1.1]
+# GEN3_PASSPORTS_EXPIRES_IN: 
--- a/fence/config-default.yaml
+++ b/fence/config-default.yaml
@@ -856,4 +856,4 @@ USERSYNC:
   fallback_to_dbgap_sftp: false
   visa_types:
     ras: [https://ras.nih.gov/visas/v1, https://ras.nih.gov/visas/v1.1]
-# GEN3_PASSPORTS_EXPIRES_IN: 
+GEN3_PASSPORTS_EXPIRES_IN: 43200
--- a/fence/resources/openid/ras_oauth2.py
+++ b/fence/resources/openid/ras_oauth2.py
@@ -84,11 +84,7 @@ class RASOauth2Client(Oauth2ClientBase):
             token = self.get_token(token_endpoint, code)
             keys = self.get_jwt_keys(jwks_endpoint)
             userinfo = self.get_userinfo(token, userinfo_endpoint)
-
-            print("-----------------v1.1--------------------")
-            print(userinfo)
-            print("--------------------------------")
-
+            
             claims = jose_jwt.decode(
                 token["id_token"],
                 keys,
--- a/fence/resources/openid/ras_oauth2.py
+++ b/fence/resources/openid/ras_oauth2.py
@@ -165,10 +165,17 @@ class RASOauth2Client(Oauth2ClientBase):
 
             token = self.get_access_token(user, token_endpoint, db_session)
             userinfo = self.get_userinfo(token, userinfo_endpoint)
-
+        
+            decoded_passport = None
+            encoded_visas = []
             encoded_passport = userinfo.get("passport_jwt_v11")
-            decoded_passport = jwt.decode(encoded_passport, verify=False)
-            encoded_visas = decoded_passport.get("ga4gh_passport_v1", [])
+
+            if encoded_passport:
+                decoded_passport = jwt.decode(encoded_passport, verify=False)
+                encoded_visas = decoded_passport.get("ga4gh_passport_v1", [])
+            else:
+                self.logger.error("Passport not available from RAS")
+                
         except Exception as e:
             err_msg = "Could not retrieve visa"
             self.logger.exception("{}: {}".format(err_msg, e))
--- a/fence/resources/openid/ras_oauth2.py
+++ b/fence/resources/openid/ras_oauth2.py
@@ -1,8 +1,9 @@
+import backoff
 import base64
 import flask
 import httpx
+import jwt
 import requests
-import backoff
 from flask_sqlalchemy_session import current_session
 from jose import jwt as jose_jwt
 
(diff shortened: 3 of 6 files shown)


fence/resources/openid/ras_oauth2.py
```diff
@@ -79,6 +80,11 @@ def get_user_id(self, code):
             userinfo_endpoint = self.get_value_from_discovery_doc(
                 "userinfo_endpoint", ""
             )
+            # as of now RAS does not provide their v1.1/userinfo in their .well-known/openid-configuration
+            # Need to manipuilate their v1 from the openid-config to say v1.1
+            # TODO: Remove this once RAS makes it availale in  (also in update_user_visas)
+            if "v1.1" not in userinfo_endpoint:
```
BinamB, line 86: Im not convinced this is the way to do this but cant think of another way.

fence/blueprints/login/ras.py
```diff
@@ -46,9 +46,16 @@ def post_login(self, user=None, token_result=None):
...
         user.ga4gh_visas_v1 = []
+
         current_session.commit()
 
-        encoded_visas = flask.g.userinfo.get("ga4gh_passport_v1", [])
+        decoded_passport = None
+        encoded_visas = []
+        encoded_passport = flask.g.userinfo.get("passport_jwt_v11")
+
+        if encoded_passport:
```
Avantol13, line 56: in what situation will this be false? we likely want to log some error/warning if that happens, right?

fence/blueprints/login/ras.py
```diff
@@ -46,9 +46,16 @@ def post_login(self, user=None, token_result=None):
...
+            decoded_passport = jwt.decode(encoded_passport, verify=False)
```
Avantol13, line 57: we want `verify=True`, right? We need to verify the outer passport's signature is RAS's before grabbing the visas

Avantol13: basically whatever verification we're doing for the RAS visas, we want to do the same type of verification here for the outer RAS passport

fence/config-default.yaml
```diff
@@ -870,3 +870,4 @@ USERSYNC:
   fallback_to_dbgap_sftp: false
   visa_types:
     ras: [https://ras.nih.gov/visas/v1, https://ras.nih.gov/visas/v1.1]
+GEN3_PASSPORTS_EXPIRES_IN: 43200
```
Avantol13, line 873: there's a specific section with other expiration times, I'd propose you move this there and add a quick description. also minor thing (I'm being really annoyingly picky here) but same naming pattern of `GEN3_PASSPORT_EXPIRES_IN` with non plural passport would match other "expires in" cfgs better :sweat_smile:

fence/resources/openid/ras_oauth2.py
Avantol13: ugh, yeah I really dislike this. Are we guaranteed that their openid-configuration userinfo will always behave in this way? It might be better (though seem worse) to just hard-code the whole dang thing. RAS_USERINFO_ENDPOINT in the cfg which (if provided) overrides the one from discovery

Avantol13: then if they surprise change the behavior, we don't break. and if they change the endpoint again, it's cfg and not code

Avantol13: but we also may want to bring this up as a request to RAS. like, how do we programatically determine this endpoint. if we haven't already

fence/resources/openid/ras_oauth2.py
```diff
@@ -151,9 +157,25 @@ def update_user_visas(self, user, pkey_cache, db_session=current_session):
...
+
             token = self.get_access_token(user, token_endpoint, db_session)
             userinfo = self.get_userinfo(token, userinfo_endpoint)
-            encoded_visas = userinfo.get("ga4gh_passport_v1", [])
+
+            decoded_passport = None
+            encoded_visas = []
+            encoded_passport = userinfo.get("passport_jwt_v11")
+
+            if encoded_passport:
```
Avantol13, line 173: same comments as above, also DRY. if this chunk of logic is used in multiple places, you may want to add some simple function `get_encoded_visas_from_v1.1_userinfo()` or something

Avantol13: so we don't have this parsing logic repeated across files

fence/resources/openid/ras_oauth2.py
BinamB: I think we should definitely ask them. Idk about adding a config though. Yeah, I guess we could use config to tell which version of userinfo endpoint to use. I guess I can pull the `issuer` from their openid-configuration and do `issuer + userinfo_version` since we connect to both their staging and prod.

fence/resources/openid/ras_oauth2.py
```diff
@@ -50,7 +51,15 @@ def get_auth_url(self):
 
         return uri
 
-    def get_userinfo(self, token, userinfo_endpoint):
+    def get_userinfo(self, token):
+        # As of now RAS does not provide their v1.1/userinfo in their .well-known/openid-configuration
+        # Need to manually change version at the moment with config
+        # TODO: Remove this once RAS makes it availale in their openid-config
```
Avantol13, line 57:
```suggestion
        # TODO: Remove this once RAS makes it available in their openid-config
```

fence/resources/openid/ras_oauth2.py
```diff
@@ -50,7 +51,15 @@ def get_auth_url(self):
...
+        issuer = self.get_value_from_discovery_doc("issuer", "")
+        userinfo_endpoint_version = config.get("RAS_USERINFO_ENDPOINT_VERSION", "v1.1")
+        userinfo_endpoint = (
+            issuer + "/openid/connect/" + userinfo_endpoint_version + "/userinfo"
```
Avantol13, line 61: even this is still maybe too hard coded, I was actually thinking of adding the full userinfo to the config somewhere.  so the `/openid/connect/v1.1/userinfo` string, just in case they change that later too

fence/resources/openid/ras_oauth2.py
```diff
@@ -69,20 +78,83 @@ def get_userinfo(self, token, userinfo_endpoint):
...
+
+        public_key = pkey_cache.get(passport_issuer, {}).get(passport_kid)
+        if not public_key:
+            try:
+                self.logger.info("Fetching public key from flask app...")
+                public_key = get_public_key_for_token(
+                    encoded_passport, attempt_refresh=True
+                )
+            except Exception as e:
+                self.logger.error(
```
Avantol13, line 110: is this really an error, won't this happen every time a new app needs to get the keys from the "cache"? If so I think this is more of an "info" level log

Avantol13: I think the final exceptions you catch below, truly are errors to be clear, just maybe not this one

fence/resources/openid/ras_oauth2.py
```diff
@@ -69,20 +78,83 @@ def get_userinfo(self, token, userinfo_endpoint):
...
+                aud=None,
+                scope={"openid"},
+                issuers=config.get("GA4GH_VISA_ISSUER_ALLOWLIST", []),
+                options={
+                    "require_iat": True,
+                    "require_exp": True,
+                },
+            )
+        except Exception as e:
+            self.logger.error(
```
Avantol13, line 142: In some cases we may want this to actually raise an error the user? but perhaps we'll catch stuff like that later on in a reasonable way. e.g. no visas show up, so they get 403. but we might need to revisit this later

fence/resources/openid/ras_oauth2.py
```diff
@@ -121,16 +193,81 @@ def get_user_id(self, code):
 
             self.logger.info("Using {} field as username.".format(field_name))
 
+            # encoded_visas = self.get_encoded_visas_v11_userinfo(userinfo, keys)
```
Avantol13, line 196: dead code, we should just remove this for now

fence/resources/openid/ras_oauth2.py
```diff
@@ -69,20 +78,83 @@ def get_userinfo(self, token, userinfo_endpoint):
             return {}
         return res.json()
 
+    def get_encoded_visas_v11_userinfo(self, userinfo, pkey_cache=None):
+        """
+        Validate Passport and get encoded visas
```
Avantol13, line 83: can you the docstring args with definition for the input type and short description?

fence/resources/openid/ras_oauth2.py
```diff
@@ -121,16 +193,81 @@ def get_user_id(self, code):
...
             flask.g.tokens = token
+            flask.g.keys = keys
 
         except Exception as e:
             self.logger.exception("{}: {}".format(err_msg, e))
             return {"error": err_msg}
 
         return {"username": username, "email": userinfo.get("email")}
 
+    def refresh_cronjob_pkey_cache(self, issuer, kid, pkey_cache):
```
Avantol13, line 209: docstring please :pray:

fence/resources/openid/ras_oauth2.py
```diff
@@ -121,16 +193,81 @@ def get_user_id(self, code):
...
+        try:
+            for key in jwt_public_keys:
+                if "kty" in key and key["kty"] == "RSA":
+                    self.logger.debug(
+                        "Serializing RSA public key (kid: {}) to PEM format.".format(
+                            key["kid"]
+                        )
+                    )
+                    # Decode public numbers https://tools.ietf.org/html/rfc7518#section-6.3.1
+                    n_padded_bytes = base64.urlsafe_b64decode(
```
Avantol13, line 230: is this done elsewhere in the code? could we make a function to not duplicate this?

Avantol13: or maybe you just moved it?

BinamB: yeah this isn't really done anywhere else. Just this specific place.

fence/resources/openid/ras_oauth2.py
```diff
@@ -69,20 +76,90 @@ def get_userinfo(self, token, userinfo_endpoint):
...
+        public_key = pkey_cache.get(passport_issuer, {}).get(passport_kid)
+        if not public_key:
+            try:
+                self.logger.info("Fetching public key from flask app...")
+                public_key = get_public_key_for_token(
+                    encoded_passport, attempt_refresh=True
+                )
+            except Exception as e:
+                self.logger.info(
+                    "Could not fetch public key from flask app to validate passport: {}. Trying .".format(
```
Avantol13, line 116:
```suggestion
                    "Could not fetch public key from flask app to validate passport: {}. Trying  to fetch from source.".format(
```

fence/resources/openid/ras_oauth2.py
```diff
@@ -69,20 +76,90 @@ def get_userinfo(self, token, userinfo_endpoint):
...
+
+        Args:
+            userinfo (dict): userinfo response
+            pkey_cache (dict): app cache of public keys_dir
+
+        Return:
+            list: list of encoded GA4GH visas
+        """
+        decoded_passport = {}
+        encoded_passport = userinfo.get("passport_jwt_v11")
```
Avantol13, line 91: this is minor (we could do this in future PRs) but I'm thinking now that this might be better to split into 2 functions (or do the passport from userinfo outside) and make a single `get_encoded_visas_from_encoded_passport` (since we'll need that exact functionality when we're getting a passport in our DRS endpoint. Up to you if you want to do this now or we can do as part of the Passport -> DRS support. I'd maybe at least add a note/comment about it though so we remember

BinamB: Made a ticket for myself for this. https://ctds-planx.atlassian.net/browse/PXP-8673

Avantol13 (approved):

PR #957, opened 2021-09-01
feat(ga4gh): move DRS access API to fence

Avantol13:
Move the /access portion of the DRS API to fence (from indexd). This more closely consolidates the functionality of creating signed URLs into a single service (fence). This should improve performance (as the existing setup requires roundtrip requests from indexd -> fence -> indexd (creating some strange circular dependencies). 

### New Features
- Expose `/ga4gh/drs/v1/objects/{object_id}/access/{access_id}` GA4GH DRS Access API via Fence

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -67,10 +67,6 @@
       "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
     },
     {
-      "path": "detect_secrets.filters.common.is_baseline_file",
-      "filename": ".secrets.baseline"
-    },
-    {
       "path": "detect_secrets.filters.common.is_ignored_due_to_verification_policies",
       "min_level": 2
     },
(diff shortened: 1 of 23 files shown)


BinamB:
I guess it doesnt matter if we remove this access endpoint from indexd now or later since it would route to fence but we should definitely remove it at some point so as to not bloat indexd. 

Also should probably add these to the swagger doc

Avantol13:
> I guess it doesnt matter if we remove this access endpoint from indexd now or later since it would route to fence but we should definitely remove it at some point so as to not bloat indexd.
> 
> Also should probably add these to the swagger doc

Yeah, I didn't want to remove it from indexd the same time I added to fence; I want to leave a transition period. It won't necessarily hurt to have the indexd one (people will just get better performance from the fence one). But agreed we should remove it eventually

BinamB (approved):
LGTM!

PR #963, opened 2021-09-21
(PXP-8679): Add new db table and update login to add entries

BinamB:

### New Features
- Add idp_to_user table that maps multiple IdP's to a single user (Currently only using for RAS)

### Deployment changes

- Requires database migration. Adding a new table to Fence.


fence/blueprints/data/indexd.py
```diff
@@ -67,7 +67,6 @@ def get_signed_url_for_file(
 ):
     requested_protocol = requested_protocol or flask.request.args.get("protocol", None)
     r_pays_project = flask.request.args.get("userProject", None)
-
```
Avantol13, line 70: could leave this line

fence/blueprints/ga4gh.py
```diff
@@ -20,7 +20,6 @@
 def get_ga4gh_signed_url(object_id, access_id):
     if not access_id:
         raise UserError("Access ID/Protocol is required.")
-
```
Avantol13, line 23: could leave this line

fence/resources/openid/idp_oauth2.py
```diff
@@ -182,3 +182,30 @@ def store_refresh_token(
...
+            idp_sub (str): sub provided by the IdP
+            provider (str): name of the Identity Provider as seen on db
+            extra_info (dict): any info sent by the IdP that could be useful
+        """
+        idp = (
+            current_session.query(IdentityProvider)
+            .filter(IdentityProvider.name == provider)
+            .first()
+        )
+        idp_id = idp.id
```
Avantol13, line 200: what if idp is not found, e.g. None. we need to handle that case

fence/blueprints/login/ras.py
```diff
@@ -143,6 +145,15 @@ def post_login(self, user=None, token_result=None):
             user=user, refresh_token=refresh_token, expires=expires + issued_time
         )
 
+        if not user.idp_to_users:
+            try:
+                # map user to idp
+                flask.current_app.ras_client.map_user_idp_info(
+                    user, userinfo.get("sub"), "ras"
```
Avantol13, line 152: should maybe use `IdentityProvider.ras` instead of "ras" string

Avantol13 (changes requested):
left a few comments

fence/resources/openid/idp_oauth2.py
BinamB: Okay so what I was thinking was that this is what happens post login which means that the user has already logged in through the idp. Even the input for `provider` is  `IdentityProvieder.ras` which means we know that it already exists. In term of us using this function for other idps it would still be post login so we know that this idp is already registered? Ohhh but i do see it as Im typing this. Its not just about logging in i guess. What if we get an idp from a different passport that we haven't seen before. We might be leveraging the same code. So I guess we should add the idp to our db in the case where we havent seen this passport issuer before :thinking:

fence/models.py
```diff
@@ -592,6 +592,40 @@ class UpstreamRefreshToken(Base):
     expires = Column(BigInteger, nullable=False)
 
 
+class IdPUser(Base):
+    # IdP & IdP sub mapping to Gen3 User sub
+
+    __tablename__ = "idp_to_user"
+
+    sub = Column(String(), primary_key=True)
```
Avantol13, line 600: actually we can't have this be the primary key b/c diff idps could have diff subs. we need to make a combined primary key of sub+idp.id

fence/models.py
```diff
@@ -592,6 +592,40 @@ class UpstreamRefreshToken(Base):
     expires = Column(BigInteger, nullable=False)
 
 
+class IdPUser(Base):
```
Avantol13, line 595: Maybe this should match Camel Case to the table name for clarity? `IdpToUser` ?

fence/blueprints/login/ras.py
```diff
@@ -143,6 +145,16 @@ def post_login(self, user=None, token_result=None):
             user=user, refresh_token=refresh_token, expires=expires + issued_time
         )
 
+        if not user.idp_to_users:
```
Avantol13, line 148: not sure how we need to handle this, but this is gonna break if the db is not migrated yet. I think we can figure that stuff out later when we're ready to merge this whole branch

fence/blueprints/login/base.py
```diff
@@ -110,6 +111,44 @@ def get(self):
...
+        idp = (
+            current_session.query(IdentityProvider)
+            .filter(IdentityProvider.name == provider)
+            .first()
+        )
+        if not idp:
+            # this is for cases where we might receive a passport that we haven't seen before.
+            # add to db if we haven't seen this before.
+            idp = IdentityProvider(
+                name=provider, description="IdP from foreign Passport"
```
Avantol13, line 134: we should make sure we only do this after the passport has been validated/verified. we don't want to add things to our db from unknown sources if we can't at least verify the signature

fence/blueprints/login/ras.py
```diff
@@ -143,6 +145,16 @@ def post_login(self, user=None, token_result=None):
...
+            try:
+                # map user to idp
+                self.map_user_idp_info(
+                    user, userinfo.get("sub"), IdentityProvider.ras, current_session
+                )
+            except Exception as e:
```
Avantol13, line 154: in what situation are we expecting this to fail?

Avantol13 (changes requested):
need some unit tests for this new behavior as well

fence/blueprints/login/base.py
BinamB: Okay moved in ras.py to do it only if its been validated. I'll leave a note in the function saying that we should only do this iff passport has been validated

fence/blueprints/login/ras.py
BinamB: Dont we usually put this  stuff in the deployment changes so devops knows we need to do a db migration. Thats what i've  seen in the past but yeah with all the changes we're going to have we'll need to figure that out

fence/models.py
```diff
@@ -592,6 +592,40 @@ class UpstreamRefreshToken(Base):
...
 
+class IdPToUser(Base):
+    # IdP & IdP sub mapping to Gen3 User sub
+
+    __tablename__ = "idp_to_user"
+
+    sub = Column(String(), primary_key=True)
+
+    fk_to_idp = Column(
+        Integer, ForeignKey(IdentityProvider.id, ondelete="CASCADE"), nullable=False
```
Avantol13, line 603: in order to make a multi-column primary key (idp+sub) I think you need this:
```suggestion
        Integer, ForeignKey(IdentityProvider.id, ondelete="CASCADE"), nullable=False, primary_key=True
```

Avantol13: you should recreate the table and ensure that's what happens, but I think that's how it should work

Avantol13 (approved):

PR #965, opened 2021-09-28
Added data use agreement to the download registration form

rnerella92:
This PR will add a data use agreement to the download registration form when "REGISTER_USERS_ON" is on. 

How to verify:

1. Set up a useryaml that has a GROUP with a policy (for example: "registered_users" with "data_upload" policy)
2. Your test user should not be in this group
3. Set up and run your Arborist
4. Run usersync
5. Put the name of the group in fence config's REGISTERED_USERS_GROUP; also set fence REGISTER_USERS_ON to true.
6. Look in the Fence db and confirm that your test user is not registered (he has no registration_info block in his additional_info column)
7. Look in Arborist db and confirm that your test user is not part of the group and does not have the group's policies
8. Set up and run your Fence, which is talking to your Arborist
9. Log in. You should be redirected to the registration form and registration form will have "Terms and Conditions" section.
10. When you enable checkbox and click on "Register" button the details will be stored in the fence db.

### New Features
Added data use agreement to the download registration form

### DIFF
--- a/fence/blueprints/register.py
+++ b/fence/blueprints/register.py
@@ -5,9 +5,10 @@ in order to gain some predefined permissions.
 """
 
 import flask
+from flask.templating import render_template
 from flask_sqlalchemy_session import current_session
 from flask_wtf import FlaskForm
-from wtforms import StringField
+from wtforms import StringField, BooleanField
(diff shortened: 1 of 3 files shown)


Avantol13 (changes requested):
We can't have project specific things in the fence repo

PR #973, opened 2021-10-20
(PXP-8682): Add visa validation and sync for data download

johnfrancismccann:
Jira Ticket: [PXP-8682](https://ctds-planx.atlassian.net/browse/PXP-8682)

### New Features
- Visa validation and authz sync for POST /data/download/{guid} endpoint

### Deployment changes

- Requires a database migration to create `iss_sub_pair_to_user` table

### DIFF
--- a/fence/__init__.py
+++ b/fence/__init__.py
@@ -425,9 +425,15 @@ def _setup_oidc_clients(app):
             HTTP_PROXY=config.get("HTTP_PROXY"),
             logger=logger,
         )
-        # TODO maybe get this from discovery doc or from the config file
-        split_url = urlparse(app.ras_client.discovery_url)
-        issuer = f"{split_url.scheme}://{split_url.netloc}"
+        issuer = app.ras_client.get_value_from_discovery_doc("issuer", "")
+        if not issuer:
+            logger.warn(
+                "Unable to determine RAS issuer from discovery doc. Instead, "
+                "RAS issuer will be set to the base url of the RAS "
+                "client's discovery url."
+            )
+            split_url = urlparse(app.ras_client.discovery_url)
+            issuer = f"{split_url.scheme}://{split_url.netloc}"
         app.issuer_to_idp = {issuer: IdentityProvider.ras}
 
     # Add OIDC client for Synapse if configured.
--- a/fence/__init__.py
+++ b/fence/__init__.py
@@ -14,13 +14,14 @@ from userdatamodel.driver import SQLAlchemyDriver
 from werkzeug.middleware.dispatcher import DispatcherMiddleware
 from azure.storage.blob import BlobServiceClient
 from azure.core.exceptions import ResourceNotFoundError
+from urllib.parse import urlparse
 
 from fence.auth import logout, build_redirect_url
 from fence.blueprints.data.indexd import S3IndexedFileLocation
 from fence.blueprints.login.utils import allowed_login_redirects, domain
 from fence.errors import UserError
 from fence.jwt import keys
-from fence.models import migrate
+from fence.models import migrate, IdentityProvider
 from fence.oidc.client import query_client
 from fence.oidc.server import server
 from fence.resources.audit.client import AuditServiceClient
@@ -424,6 +425,10 @@ def _setup_oidc_clients(app):
             HTTP_PROXY=config.get("HTTP_PROXY"),
             logger=logger,
         )
+        # TODO maybe get this from discovery doc or from the config file
+        split_url = urlparse(app.ras_client.discovery_url)
+        issuer = f"{split_url.scheme}://{split_url.netloc}"
+        app.issuer_to_idp = {issuer: IdentityProvider.ras}
 
     # Add OIDC client for Synapse if configured.
     if "synapse" in oidc:
--- a/fence/__init__.py
+++ b/fence/__init__.py
@@ -401,6 +401,7 @@ def _set_authlib_cfgs(app):
 
 def _setup_oidc_clients(app):
     oidc = config.get("OPENID_CONNECT", {})
+    app.issuer_to_idp = {}
 
     # Add OIDC client for Google if configured.
     if "google" in oidc:
@@ -425,16 +426,12 @@ def _setup_oidc_clients(app):
             HTTP_PROXY=config.get("HTTP_PROXY"),
             logger=logger,
         )
-        issuer = app.ras_client.get_value_from_discovery_doc("issuer", "")
-        if not issuer:
-            logger.warn(
-                "Unable to determine RAS issuer from discovery doc. Instead, "
-                "RAS issuer will be set to the base url of the RAS "
-                "client's discovery url."
-            )
-            split_url = urlparse(app.ras_client.discovery_url)
-            issuer = f"{split_url.scheme}://{split_url.netloc}"
-        app.issuer_to_idp = {issuer: IdentityProvider.ras}
+        for allowed_issuer in config["GA4GH_VISA_ISSUER_ALLOWLIST"]:
+            if app.ras_client.discovery_url.startswith(allowed_issuer):
+                app.issuer_to_idp[allowed_issuer] = IdentityProvider.ras
+                break
+        else:
+            logger.warn("Could not determine issuer for the RAS OIDC client")
 
     # Add OIDC client for Synapse if configured.
     if "synapse" in oidc:
--- a/fence/blueprints/login/ras.py
+++ b/fence/blueprints/login/ras.py
@@ -14,7 +14,9 @@ from gen3authz.client.arborist.client import ArboristClient
 
 from fence.blueprints.login.base import DefaultOAuth2Login, DefaultOAuth2Callback
 from fence.config import config
-from fence.scripting.fence_create import init_syncer
+
+# TODO comment for this, maybe move to top
+import fence.scripting.fence_create
 from fence.utils import get_valid_expiration
 
 logger = get_logger(__name__)
@@ -186,12 +188,12 @@ class RASCallback(DefaultOAuth2Callback):
             if not isinstance(dbGaP, list):
                 dbGaP = [dbGaP]
 
-            sync = init_syncer(
+            sync = fence.scripting.fence_create.init_syncer(
                 dbGaP,
                 None,
                 DB,
                 arborist=arborist,
             )
-            sync.sync_single_user_visas(user, current_session)
+            sync.sync_single_user_visas(user, user.ga4gh_visas_v1, current_session)
 
         super(RASCallback, self).post_login()
--- a/fence/blueprints/login/ras.py
+++ b/fence/blueprints/login/ras.py
@@ -106,11 +106,6 @@ class RASCallback(DefaultOAuth2Callback):
                 # Also require 'sub' claim (see note above about pyjwt and the options arg).
                 if "sub" not in decoded_visa:
                     raise JWTError("Visa is missing the 'sub' claim.")
-                if not user.idp_to_users:
-                    # map user to idp
-                    self.map_user_idp_info(
-                        user, userinfo.get("sub"), IdentityProvider.ras, current_session
-                    )
             except Exception as e:
                 logger.error("Visa failed validation: {}. Discarding visa.".format(e))
                 continue
--- a/fence/blueprints/login/ras.py
+++ b/fence/blueprints/login/ras.py
@@ -1,6 +1,9 @@
 import flask
 import jwt
 import os
+
+# the whole fence_create module is imported to avoid issue with circular imports
+import fence.scripting.fence_create
 from distutils.util import strtobool
 from authutils.errors import JWTError
 from authutils.token.core import validate_jwt
@@ -15,8 +18,6 @@ from gen3authz.client.arborist.client import ArboristClient
 from fence.blueprints.login.base import DefaultOAuth2Login, DefaultOAuth2Callback
 from fence.config import config
 
-# TODO comment for this, maybe move to top
-import fence.scripting.fence_create
 from fence.utils import get_valid_expiration
 
 logger = get_logger(__name__)
--- a/fence/config-default.yaml
+++ b/fence/config-default.yaml
@@ -864,6 +864,14 @@ ASSUME_ROLE_CACHE_SECONDS: 1800
 # will have access to download data.
 REGISTER_USERS_ON: false
 REGISTERED_USERS_GROUP: ''
+
+# Number of projects that can be registered to a Google Service Account
+SERVICE_ACCOUNT_LIMIT: 6
+
+# //////////////////////////////////////////////////////////////////////////////////////
+# GA4GH SUPPORT: DATA ACCESS AND AUTHORIZATION SYNCING
+# //////////////////////////////////////////////////////////////////////////////////////
+
 # RAS refresh_tokens expire in 15 days
 RAS_REFRESH_EXPIRATION: 1296000
 # List of JWT issuers from which Fence will accept GA4GH visas
@@ -876,11 +884,7 @@ GA4GH_VISA_V1_CLAIM_REQUIRED_FIELDS:
   asserted: 0
   value: "https://stsstg.nih.gov/passport/dbgap/v1.1"
   source: "https://ncbi.nlm.nih.gov/gap"
-DELETE_EXPIRED_GOOGLE_ACCESS_JOB_FREQUENCY_IN_SECONDS: 300
-
-# Number of projects that can be registered to a Google Service Account
-SERVICE_ACCOUNT_LIMIT: 6
-
+EXPIRED_AUTHZ_REMOVAL_JOB_FREQ_IN_SECONDS: 300
 # Global sync visas during login
 # None(Default): Allow per client i.e. a fence client can pick whether or not to sync their visas during login with parse_visas param in /authorization endpoint
 # True: Parse for all clients i.e. a fence client will always sync their visas during login
--- a/fence/config-default.yaml
+++ b/fence/config-default.yaml
@@ -871,7 +871,13 @@ GA4GH_VISA_ISSUER_ALLOWLIST:
   - '{{BASE_URL}}'
   - 'https://sts.nih.gov'
   - 'https://stsstg.nih.gov'
-# Number of projects that can be registered to a Google Service Accont
+GA4GH_VISA_V1_CLAIM_REQUIRED_FIELDS:
+  type: "https://ras.nih.gov/visas/v1.1"
+  asserted: 0
+  value: "https://stsstg.nih.gov/passport/dbgap/v1.1"
+  source: "https://ncbi.nlm.nih.gov/gap"
+
+# Number of projects that can be registered to a Google Service Account
 SERVICE_ACCOUNT_LIMIT: 6
 
 # Global sync visas during login
--- a/fence/config-default.yaml
+++ b/fence/config-default.yaml
@@ -881,7 +881,6 @@ GA4GH_VISA_ISSUER_ALLOWLIST:
   - 'https://stsstg.nih.gov'
 GA4GH_VISA_V1_CLAIM_REQUIRED_FIELDS:
   type:
-    - "https://ras.nih.gov/visas/v1"
     - "https://ras.nih.gov/visas/v1.1"
   value:
     - "https://stsstg.nih.gov/passport/dbgap/v1.1"
--- a/fence/config-default.yaml
+++ b/fence/config-default.yaml
@@ -876,6 +876,7 @@ GA4GH_VISA_V1_CLAIM_REQUIRED_FIELDS:
   asserted: 0
   value: "https://stsstg.nih.gov/passport/dbgap/v1.1"
   source: "https://ncbi.nlm.nih.gov/gap"
+DELETE_EXPIRED_GOOGLE_ACCESS_JOB_FREQUENCY_IN_SECONDS: 300
 
 # Number of projects that can be registered to a Google Service Account
 SERVICE_ACCOUNT_LIMIT: 6
--- a/fence/config-default.yaml
+++ b/fence/config-default.yaml
@@ -880,10 +880,13 @@ GA4GH_VISA_ISSUER_ALLOWLIST:
   - 'https://sts.nih.gov'
   - 'https://stsstg.nih.gov'
 GA4GH_VISA_V1_CLAIM_REQUIRED_FIELDS:
-  type: "https://ras.nih.gov/visas/v1.1"
-  asserted: 0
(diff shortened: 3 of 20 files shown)


fence/resources/ga4gh/passports.py
```diff
@@ -118,40 +167,106 @@ def get_unvalidated_visas_from_valid_passport(passport, pkey_cache=None):
...
-    pass
-
+def validate_visa(raw_visa):
+    # TODO check that there is no JKU field in header?
+    decoded_visa = validate_jwt(
+        raw_visa,
+        attempt_refresh=True,
+        scope={"openid", "ga4gh_passport_v1"},
+        require_purpose=False,
+        issuers=config.get("GA4GH_VISA_ISSUER_ALLOWLIST", []),
```
Avantol13, line 177: if you're using config you don't have to .get(), you can access directly and it will default to whatever is in the default config file. so just 
```suggestion
        issuers=config["GA4GH_VISA_ISSUER_ALLOWLIST"],
```

fence/resources/ga4gh/passports.py
```diff
@@ -1,62 +1,128 @@
...
-            subject_id, issuer = get_sub_iss_from_visa(raw_visa)
+        if not raw_visas:
+            continue
 
-            # query idp user table
-            gen3_user = get_or_create_gen3_user_from_sub_iss(subject_id, issuer)
-            user_ids_from_passports.append(gen3_user.id)
+        identity_to_visas = collections.defaultdict(list)
+        min_visa_expiration = int(time.time()) + datetime.timedelta(hours=1).seconds
+        for raw_visa in raw_visas:
```
johnfrancismccann, line 69: Note that this for loop is indented within `for passport in passports`. The reason for this is the assumption that `min_visa_expiration` should be calculated for all visas within a single passport, as opposed to all visas across all passports.

fence/config-default.yaml
```diff
@@ -871,7 +871,14 @@ GA4GH_VISA_ISSUER_ALLOWLIST:
   - '{{BASE_URL}}'
   - 'https://sts.nih.gov'
   - 'https://stsstg.nih.gov'
-# Number of projects that can be registered to a Google Service Accont
+GA4GH_VISA_V1_CLAIM_REQUIRED_FIELDS:
```
Avantol13, line 874: we may want to section off all the ga4gh/ras specific config with something like: 
```
# DATA ACCESS: GA4GH SUPPORT AND AUTHORIZATION SYNCING
```

fence/blueprints/login/ras.py
```diff
@@ -91,11 +93,6 @@ def post_login(self, user=None, token_result=None):
                 # Embedded token must not contain aud claim
                 if "aud" in decoded_visa:
                     raise JWTError("Visa contains the 'aud' claim.")
-                if not user.idp_to_users:
-                    # map user to idp
-                    self.map_user_idp_info(
```
Avantol13, line 96: does this happen later now? it seems like we still need to add entries to this table on login if possible

fence/resources/ga4gh/passports.py
```diff
@@ -1,62 +1,128 @@
...
+                min_visa_expiration = min(
+                    min_visa_expiration, validated_decoded_visa.get("exp")
+                )
+            except Exception as exc:
+                logger.warning(f"invalid visa provided, ignoring. Error: {exc}")
+                continue
 
-        except Exception as exc:
-            logger.warning(f"invalid visa provided, ignoring. Error: {exc}")
+        delete_expired_google_access_job_frequency = config.get(
```
Avantol13, line 85: we should name this something else since it's not just the google job but also the arborist job. Maybe it should be `EXPIRED_AUTHZ_REMOVAL_JOB_FREQ_IN_SECONDS` ?

fence/resources/ga4gh/passports.py
```diff
@@ -1,62 +1,128 @@
...
+                    type=validated_decoded_visa["ga4gh_visa_v1"]["type"],
+                    asserted=int(validated_decoded_visa["ga4gh_visa_v1"]["asserted"]),
+                    expires=int(validated_decoded_visa["exp"]),
+                    ga4gh_visa=raw_visa,
+                )
+                for raw_visa, validated_decoded_visa in visas
+            ]
+            # NOTE: does not validate, assumes validation occurs above.
+            sync_visa_authorization(gen3_user, ga4gh_visas, min_visa_expiration)
+            usernames_from_current_passport.append(gen3_user.username)
```
Avantol13, line 115: I believe this function should return a list of the `gen3_user`'s. Later when we only need their usernames we can do that. But it's going to be more extensible to pass around the references to the users in case we ever need anything other than the usernames. Or even just change the caller of this function to immediately make a list of just of the usernames, but the function returns all the users

fence/resources/ga4gh/passports.py
```diff
@@ -1,62 +1,128 @@
...
 
-        except Exception as exc:
-            logger.warning(f"invalid visa provided, ignoring. Error: {exc}")
+        delete_expired_google_access_job_frequency = config.get(
+            "DELETE_EXPIRED_GOOGLE_ACCESS_JOB_FREQUENCY_IN_SECONDS", 300
+        )
+        min_visa_expiration -= delete_expired_google_access_job_frequency
+        if min_visa_expiration <= int(time.time()):
+            logger.warning(
+                "the passport's minimum visa expiration time fell within "
```
Avantol13, line 91: we should alter this text, b/c this could also mean that the expiration is before now(). basically it's already expired or it will expire before the next authorization expiration job can run

fence/config-default.yaml
```diff
@@ -871,7 +871,14 @@ GA4GH_VISA_ISSUER_ALLOWLIST:
...
+  type: "https://ras.nih.gov/visas/v1.1"
```
Avantol13, line 875: we are going to want all these to be lists of allowed/possible values

fence/resources/ga4gh/passports.py
```diff
@@ -118,40 +184,146 @@ def get_unvalidated_visas_from_valid_passport(passport, pkey_cache=None):
...
 
-    return subject_id, issuer
+    Args:
+        raw_visa (str): a raw, encoded visa including header, payload, and signature
 
+    Return:
+        dict: the decoded payload if validation was successful. an exception
+              is raised if validation was unsuccessful
+    """
+    # TODO check that there is no JKU field in header?
```
Avantol13, line 200: do we need to do this?

fence/blueprints/login/ras.py
johnfrancismccann: It actually happens earlier now, in `get_user_id` in `ras_oauth2.py`, just before a user is logged in. This was so that we can handle the cases when we change the username from a prior DRS access request (instead of creating a duplicate user) or when we want to prefer the prior DRS user over the stale user from an earlier login.

fence/__init__.py
```diff
@@ -424,6 +425,16 @@ def _setup_oidc_clients(app):
             HTTP_PROXY=config.get("HTTP_PROXY"),
             logger=logger,
         )
+        issuer = app.ras_client.get_value_from_discovery_doc("issuer", "")
```
Avantol13, line 428: not sure how I feel about this block. I think this should be something we try to configure ahead of time to reduce this startup call. we can add an optional `token_issuer` to the configured clients under the `OPENID_CONNECT` configuration block

Avantol13: we actually already have `GA4GH_VISA_ISSUER_ALLOWLIST`, we could maybe merge that with the `OPENID_CONNECT` block? list of iss per client. We already use this block to setup the ras_client https://github.com/uc-cdis/fence/blob/d38c08ecd2bc7abef8835715c4a784cbea46b081/fence/job/visa_update_cronjob.py#L47-L57 . On startup we could construct `issuer_to_idp` using only configuration and not reaching out externally.


when we add ourselves we'll have to have some concept of `_self` for IdP or something but we can figure that out later

fence/resources/ga4gh/passports.py
```diff
@@ -1,62 +1,128 @@
...
+                    source=validated_decoded_visa["ga4gh_visa_v1"]["source"],
+                    type=validated_decoded_visa["ga4gh_visa_v1"]["type"],
+                    asserted=int(validated_decoded_visa["ga4gh_visa_v1"]["asserted"]),
+                    expires=int(validated_decoded_visa["exp"]),
+                    ga4gh_visa=raw_visa,
+                )
+                for raw_visa, validated_decoded_visa in visas
+            ]
+            # NOTE: does not validate, assumes validation occurs above.
+            sync_visa_authorization(gen3_user, ga4gh_visas, min_visa_expiration)
```
Avantol13, line 114: you pass in `min_visa_expiration` but I think later it pulls the `expires` from the visa. We need to make sure that what you pass in overrides that or that later where it's used it still takes into account the authz job freq

Avantol13: also I think we need to rework/rethink `sync_single_user_visas` in general... it currently removes all visas before adding the ones provided. So this loop you have here will overwrite anything for subsequent iss/sub pairs if it's for the same `gen3_user` . For now... we don't expect multiple Gen3 users in a single RAS Passport, but this will be a problem in the future.

Avantol13 (changes requested):

fence/resources/ga4gh/passports.py
```diff
@@ -1,62 +1,128 @@
...
         except Exception as exc:
             logger.warning(f"invalid passport provided, ignoring. Error: {exc}")
             continue
 
-    for raw_visa in raw_visas:
-        try:
-            # below function also validates visa (or raises exception) and
-            # extracts the subject id
-            subject_id, issuer = get_sub_iss_from_visa(raw_visa)
+        if not raw_visas:
```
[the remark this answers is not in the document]
johnfrancismccann: Yeah no problem, and an empty `raw_visas` list could also mean that the passport wasn't valid, since `get_unvalidated_visas_from_valid_passport` returns an empty list for an invalid passport.

fence/resources/ga4gh/passports.py
johnfrancismccann: Would there still be overwriting of old visa policies even with [this change](https://github.com/uc-cdis/fence/pull/973/files#r746023680)? Not seeing where else that would happen.

fence/blueprints/login/ras.py
Avantol13: ohh, okay that makes sense. I'll take another look with that in mind

fence/config-default.yaml
Avantol13: or exclude it from the configuration entirely and just embed that logic in the code itself (if you don't expect we'll ever need to change it)

fence/resources/ga4gh/passports.py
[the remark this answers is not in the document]
Avantol13: ooh very good point, yeah definitely add a comment clarifying all that

fence/resources/ga4gh/passports.py
Avantol13: yeah maybe something generic like: `The passport's earliest valid visa expiration time is set to occur within {EXPIRED_AUTHZ_REMOVAL_JOB_FREQ_IN_SECONDS} seconds from now, which is too soon an expiration to handle.`

fence/resources/ga4gh/passports.py
Avantol13: hmm we may want to check for this and raise a notimplemented exception or something? `We do not currently support Visa Document Tokens by passing `jku` in the header. We only support Visa Access Tokens`

fence/sync/sync_users.py
```diff
@@ -1670,9 +1671,9 @@ def _update_authz_in_arborist(
                 username = user.username
 
             self.arborist_client.create_user_if_not_exist(username)
-
-            # TODO make this smarter - it should do a diff, not revoke all and add
-            self.arborist_client.revoke_all_policies_for_user(username)
+            if not single_user_sync:
+                # TODO make this smarter - it should do a diff, not revoke all and add
+                self.arborist_client.revoke_all_policies_for_user(username)
```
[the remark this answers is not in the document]
Avantol13: trying to think this through.. so if `single_user_sync` is True, we always expect that an expiration is set on the arborist policy such that it's removed at the right time. But won't this create new policies for every request with a new passport (I guess this might be expected)? What about with the same passport? Will this mean that if I send 500 requests with the same passport I get 500 policies in Arborist? Or does this code already handle those conflicts?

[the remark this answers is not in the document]
johnfrancismccann: New policies will be created for every request with a new passport assuming each POSTed passport has a different set of dbgap permissions. New policies won't be created for overlapping dbgap permissions across different passports/requests. 

So for 500 requests with [this sample passport](https://sts.nih.gov/passport/dbgap/v1.1), only two policies would be created in Arborist: `programs.phs000021.c1-read` and `programs.phs000021.c1-read-storage`. The policy name is only [created](https://github.com/uc-cdis/fence/blob/defadaab9916317828a64559effdc44f014d64c5/fence/sync/sync_users.py#L1703) based off of study id, consent group, and permission.

Below are Arborist logs for POSTing the sample passport on my laptop for the second time (username redacted).
```
::1 - - [11/Nov/2021:06:57:28 -0800] "GET /health HTTP/1.1" 200 0 "" "python-httpx/0.20.0"
::1 - - [11/Nov/2021:06:57:28 -0800] "GET /resource/programs/phs000021.c1 HTTP/1.1" 200 124 "" "python-httpx/0.20.0"
2021/11/11 06:57:28 response.go:85: INFO: failed to insert user: user with this ID already exists: <redacted>
::1 - - [11/Nov/2021:06:57:28 -0800] "POST /user HTTP/1.1" 409 146 "" "python-httpx/0.20.0"
::1 - - [11/Nov/2021:06:57:28 -0800] "GET /health HTTP/1.1" 200 0 "" "python-httpx/0.20.0"
2021/11/11 06:57:28 response.go:85: INFO: failed to insert user: user with this ID already exists: <redacted>
::1 - - [11/Nov/2021:06:57:28 -0800] "POST /user HTTP/1.1" 409 146 "" "python-httpx/0.20.0"
2021/11/11 06:57:28 response.go:85: INFO: failed to insert role: role with this ID already exists: read-storage
::1 - - [11/Nov/2021:06:57:28 -0800] "POST /role/ HTTP/1.1" 409 104 "" "python-httpx/0.20.0"
2021/11/11 06:57:28 server.go:667: INFO: overwrote policy programs.phs000021.c1-read-storage
::1 - - [11/Nov/2021:06:57:28 -0800] "PUT /policy/programs.phs000021.c1-read-storage HTTP/1.1" 201 138 "" "python-httpx/0.20.0"
2021/11/11 06:57:28 server.go:1154: INFO: granted policy programs.phs000021.c1-read-storage to user <redacted>
::1 - - [11/Nov/2021:06:57:28 -0800] "POST /user/<redacted>/policy HTTP/1.1" 204 0 "" "python-httpx/0.20.0"
2021/11/11 06:57:28 response.go:85: INFO: failed to insert role: role with this ID already exists: read
::1 - - [11/Nov/2021:06:57:28 -0800] "POST /role/ HTTP/1.1" 409 96 "" "python-httpx/0.20.0"
2021/11/11 06:57:28 server.go:667: INFO: overwrote policy programs.phs000021.c1-read
::1 - - [11/Nov/2021:06:57:28 -0800] "PUT /policy/programs.phs000021.c1-read HTTP/1.1" 201 130 "" "python-httpx/0.20.0"
2021/11/11 06:57:28 server.go:1154: INFO: granted policy programs.phs000021.c1-read to user <redacted>
::1 - - [11/Nov/2021:06:57:28 -0800] "POST /user/<redacted>/policy HTTP/1.1" 204 0 "" "python-httpx/0.20.0"
```
The only thing I would be concerned about is the PUT /policy/{policy_name} request, in which `overwrote policy` gets logged. For example, what if that were overwriting permissions from a separate usersync?

fence/resources/openid/ras_oauth2.py
```diff
@@ -145,7 +169,79 @@ def get_user_id(self, code):
...
+
+            self.logger.info("Mapping issuer and subject id to Fence user")
+            iss_sub_pair_to_user = IssSubPairToUser(iss=issuer, sub=subject_id)
+            iss_sub_pair_to_user.user = user
+            db_session.add(iss_sub_pair_to_user)
+            db_session.commit()
+            return iss_sub_pair_to_user.user.username
 
     @backoff.on_exception(backoff.expo, Exception, **DEFAULT_BACKOFF_SETTINGS)
     def update_user_visas(self, user, pkey_cache, db_session=current_session):
```
BinamB, line 247: I think we should use the `validate_visa` you implemented here as well. We call `update_user_visas` during access token polling that validates both passports and visas. It would be nice to have the same visa validation logic in this job.

BinamB: Oh also reminder that when we're adding the `validate_visa` here we might want to pass `pkey_cache` to the `validate_jwt` function since flask cache wont work while running this job.

fence/resources/ga4gh/passports.py
```diff
@@ -1,59 +1,135 @@
...
+            users_from_current_passport.append(gen3_user)
+
+        put_gen3_usernames_for_passport_into_cache(
+            passport, users_from_current_passport
+        )
+        users_from_all_passports.extend(users_from_current_passport)
 
-        if not was_cached:
-            put_gen3_user_ids_for_passport_into_cache(passport, user_ids_from_passports)
+    # TODO use user_ids_from_all_passports that were returned from cache to
```
Avantol13, line 125: I didn't really think through this case, honestly I think it's a bit too much to include this db call without a reason. I hate to reverse course but just handling the IDs here is probably fine for now

fence/__init__.py
johnfrancismccann: Updated so that `issuer_to_idp` is constructed only using configuration. I didn’t nest `GA4GH_VISA_ISSUER_ALLOWLIST` in the `ras` block of `OPENID_CONNECT` in `config-default.yaml` because it was getting overwritten by the local config file when testing.

fence/resources/ga4gh/passports.py
```diff
@@ -118,40 +194,176 @@ def get_unvalidated_visas_from_valid_passport(passport, pkey_cache=None):
...
-    # TODO
+    if "conditions" in decoded_visa["ga4gh_visa_v1"]:
+        logger.warning(
+            'Condition checking is not yet supported, but a visa was received that contained the "conditions" field'
+        )
+        if decoded_visa["ga4gh_visa_v1"]["conditions"]:
+            raise Exception('"conditions" field in "ga4gh_visa_v1" is not empty')
 
-    return subject_id, issuer
+    logger.info("Visa was successfully validated")
```
Avantol13, line 266: This should probably have some identifier for the visa in the log

fence/resources/openid/ras_oauth2.py
```diff
@@ -145,7 +169,80 @@ def get_user_id(self, code):
...
+        """
+        Map <issuer, subject_id> combination to a Fence user whose username
+        equals the username argument passed into this function.
+
+        One exception to this is when two Fence users exist who both
+        correspond to the user who is trying to log in. Please see logged
+        warning for more details.
+
+        Args:
+            issuer (str): RAS issuer
```
Avantol13, line 184: Remove "RAS", will eventually map other idps

fence/resources/openid/ras_oauth2.py
```diff
@@ -145,7 +169,80 @@ def get_user_id(self, code):
...
+                (issuer, subject_id)
+            )
+            user = query_for_user(db_session, username)
+            if iss_sub_pair_to_user:
+                if not user:
+                    # TODO just say DRS, not DRS/data
+                    self.logger.info(
+                        "Issuer and subject id have already been mapped to a "
+                        "Fence user created from the DRS endpoint. "
+                        "Changing said user's username to the username "
```
Avantol13, line 207: You should include the actual sub and iss and username previous and future in this log to aide I'm debugging

fence/resources/openid/ras_oauth2.py
```diff
@@ -145,7 +169,80 @@ def get_user_id(self, code):
...
+                        "returned from the RAS userinfo endpoint."
+                    )
+                    # TODO also change username in Arborist
+                    iss_sub_pair_to_user.user.username = username
+                    iss_sub_pair_to_user.user.email = email
+                    db_session.commit()
+                elif iss_sub_pair_to_user.user.username != username:
+                    self.logger.warning(
+                        "Two users exist in the Fence database corresponding "
+                        "to the RAS user who is currently trying to log in: one "
```
Avantol13, line 217: Same as above comment

fence/resources/ga4gh/passports.py
johnfrancismccann: So would that mean that `get_gen3_users_from_ga4gh_passports` should return a list of user ids or usernames? If a list of user ids is returned, then a later db query would still be necessary to determine usernames for the later Arborist authz check.

I think the minimum number of db queries would be made by having the cache store usernames and `get_gen3_users_from_ga4gh_passports` also return usernames.

Avantol13: yeah, I agree. Let's do usernames for now. If in the future we need more/something else, we'll handle it then

fence/resources/ga4gh/passports.py
johnfrancismccann: Updated. Logged `jti` and `txn` above just after call to `validate_jwt`

fence/resources/openid/ras_oauth2.py
```diff
@@ -145,7 +169,80 @@ def get_user_id(self, code):
...
+                        "created from an earlier login and one created from "
+                        "the DRS endpoint. The one created from the "
+                        "DRS endpoint will be logged in, rendering the "
+                        "other one inaccessible."
+                    )
+                return iss_sub_pair_to_user.user.username
+
+            if not user:
+                self.logger.info(
```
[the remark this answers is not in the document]
johnfrancismccann: Added a `create_user` function in `fence/models.py` that is now used for [DRS access](https://github.com/uc-cdis/fence/pull/973/files#diff-6eb903144ac0e54f46571c6b7a5825bae74dc9122c01bd99c2defa609a11bf5cR290) and during [RAS login](https://github.com/uc-cdis/fence/pull/973/files#diff-55adb0b16b1bdae657a354a20608d0a97bafc22eee42b2ff55d9a0e51baf9434R228). I think it would be tricky to also use `create_user` in [`login_user`](https://github.com/uc-cdis/fence/blob/7880778fef77655d27ff245eef2aeee115903c43/fence/auth.py#L98) because of a possible code path in which the user already exists but idp isn't linked yet.

Avantol13 (approved):

PR #975, opened 2021-11-10
feat(users): add id from idp to db table if possible (also update exi…

Avantol13:
…sting entries)

### Improvements
- populate the ID the IdP uses in the database if provided (for future reference) NOTE: This may be the same as the unique username determined, but if not, this allows storing both

### DIFF
--- a/fence/blueprints/login/base.py
+++ b/fence/blueprints/login/base.py
@@ -56,7 +56,14 @@ class DefaultOAuth2Login(Resource):
 
 
 class DefaultOAuth2Callback(Resource):
-    def __init__(self, idp_name, client, username_field="email", email_field="email"):
+    def __init__(
+        self,
+        idp_name,
+        client,
+        username_field="email",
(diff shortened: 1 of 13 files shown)


fence/blueprints/login/base.py
```diff
@@ -101,33 +112,36 @@ def get(self):
         result = self.client.get_user_id(code)
```
johnfrancismccann, line 112: I'm pretty sure that [`RASOauth2Client.get_user_id`](https://github.com/uc-cdis/fence/blob/defadaab9916317828a64559effdc44f014d64c5/fence/resources/openid/ras_oauth2.py#L204) (and `get_user_id` of any other clients for which we want to store sub) will need to be updated to also return sub.

Avantol13: oh, really good catch. Yup, just pushed an update

johnfrancismccann (approved):

PR #977, opened 2021-11-15
PXP-9099 Fix redirection to default IDP in oauth2 flow

paulineribeyre:
Jira Ticket: [PXP-9099](https://ctds-planx.atlassian.net/browse/PXP-9099)

Fix bug causing a redirection to the default NIH entity ID instead of the custom entity ID when users login through the oauth2 flow WITHOUT SPECIFYING AN IDP.

The config below should redirect to the "X" provider when users log into Gen3 via a client portal which does not specify the IDP, but it redirects to "urn:mace:incommon:nih.gov" instead (default "shib_idp") because the "shib_idp" param is not passed to centralized fence.

```
DEFAULT_LOGIN_IDP: fence
LOGIN_OPTIONS:
  - name: 'NIH Login'
    idp: fence
    fence_idp: shibboleth
    shib_idps:
      - X
```

### Bug Fixes
- Fix the redirection to the default IDP in the `/oauth2/authorize` endpoint: when the default IDP is "fence", include `idp` and `shib_idp` query parameters

### Improvements
- Update the unit tests so we don't make real requests to get the list of InCommon IDPs

### DIFF
--- a/fence/blueprints/login/__init__.py
+++ b/fence/blueprints/login/__init__.py
@@ -44,6 +44,188 @@ IDP_URL_MAP = {
 }
 
 
+def absolute_login_url(provider_id, fence_idp=None, shib_idp=None):
+    """
+    Args:
+        provider_id (str): provider to log in with; an IDP_URL_MAP key.
+        fence_idp (str, optional): if provider_id is "fence"
+            (multi-tenant Fence setup), fence_idp can be any of the
+            providers supported by the other Fence. If not specified,
+            will default to NIH login.
+        shib_idp (str, optional): if provider_id is "fence" and
+            fence_idp is "shibboleth", shib_idp can be any Shibboleth/
+            InCommon provider. If not specified, will default to NIH
+            login.
+
+    Returns:
+        str: login URL for this provider, including extra query
+            parameters if fence_idp and/or shib_idp are specified.
+    """
+    try:
+        base_url = config["BASE_URL"].rstrip("/")
+        login_url = base_url + "/login/{}".format(IDP_URL_MAP[provider_id])
+    except KeyError as e:
+        raise InternalError("identity provider misconfigured: {}".format(str(e)))
+
+    params = {}
+    if fence_idp:
+        params["idp"] = fence_idp
+    if shib_idp:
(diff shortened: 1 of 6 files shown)


fence/blueprints/oauth2.py
```diff
@@ -77,15 +77,17 @@ def authorize(*args, **kwargs):
...
+        if not config.get("DEFAULT_LOGIN_IDP") and "default" not in config.get(
+            "ENABLED_IDENTITY_PROVIDERS", {}
+        ):
+            # fall back on deprecated DEFAULT_LOGIN_URL
+            login_url = config.get("DEFAULT_LOGIN_URL")
+        else:
+            default_provider_info, _ = get_login_providers_info()
+            idp = default_provider_info["idp"]
+            # if more than 1 URL is configured, default to the 1st one
+            login_url = default_provider_info["urls"][0]["url"]
```
paulineribeyre, line 90: this url already includes the parameters, see https://github.com/uc-cdis/fence/blob/ffa4319218a0df5a9adac0421ef228a5af0a0977/fence/blueprints/login/__init__.py#L70-L75

fence/blueprints/login/__init__.py
```diff
@@ -44,6 +44,188 @@
 }
 
 
+def absolute_login_url(provider_id, fence_idp=None, shib_idp=None):
```
paulineribeyre, line 47: just moving these functions out of `make_login_blueprint()` and replacing `app` with `flask.current_app`

BinamB (approved):
lgtm!

fence/blueprints/login/__init__.py
```diff
@@ -44,6 +44,188 @@
...
+        raise InternalError(
+            "default provider misconfigured: DEFAULT_LOGIN_IDP is set to {}, which is not configured in LOGIN_OPTIONS".format(
+                default_idp
+            )
+        )
+
+    return default_provider_info, all_provider_info
+
+
 def make_login_blueprint(app):
```
johnfrancismccann, line 229: Can we remove `app` as an argument, both here and in `fence/__init__.py`?

PR #982, opened 2021-11-22
fix(audit): fix audit log to send Gen3 sub instead of idp sub

Avantol13:
bug introduced in 5.5.2

### Bug Fixes
- Fix issue with IdP subject being sent to audit log on login instead of Gen3 subject

### DIFF
--- a/fence/blueprints/login/base.py
+++ b/fence/blueprints/login/base.py
@@ -128,7 +128,7 @@ class DefaultOAuth2Callback(Resource):
 def prepare_login_log(idp_name, id_from_idp=None):
     flask.g.audit_data = {
         "username": flask.g.user.username,
-        "sub": id_from_idp,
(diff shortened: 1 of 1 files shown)


fence/blueprints/login/base.py
```diff
@@ -128,7 +128,7 @@ def post_login(self, user=None, token_result=None, id_from_idp=None):
 def prepare_login_log(idp_name, id_from_idp=None):
     flask.g.audit_data = {
         "username": flask.g.user.username,
-        "sub": id_from_idp,
+        "sub": flask.g.user.id,
```
paulineribeyre: we can remove the `id_from_idp` param from the functions too

it’s been added to `post_login` but it doesn’t look like there’s any reason to, now that we don’t need it in `prepare_login_log`

paulineribeyre (approved):

PR #984, opened 2021-11-23
Feat/create presigned url passports

BinamB:
### New Features
- Return aws signed urls from `POST /ga4gh/drs/v1/objects/<guid>/access/<access method>` endpoint with passports
- Return google signed urls from `POST /ga4gh/drs/v1/objects/<guid>/access/<access method>` endpoint with passports

### Improvements
- install vim in dockerfile for live debugging

### Dependency updates
- gen3authz ^1.1.1 -> ^1.3.0 to support sending auth/request to arborist with user_id instead of jwt token


fence/blueprints/data/indexd.py
```diff
@@ -1434,21 +1460,28 @@ def delete(self, container, blob):  # pylint: disable=R0201
...
 
 
-def _get_user_info(sub_type=str):
+def _get_user_info(sub_type=str, user=None):
     """
     Attempt to parse the request for token to authenticate the user. fallback to
     populated information about an anonymous user.
     By default, cast `sub` to str. Use `sub_type` to override this behavior.
     """
     # TODO Update to support POSTed passport
```
johnfrancismccann, line 1469: We can take this TODO comment out, right?

johnfrancismccann:
`gen3authz-1.3.0.tar.gz` can be removed.

fence/blueprints/data/indexd.py
```diff
@@ -411,20 +413,25 @@ def get_signed_url(
...
-            ):
+            )
+            if not authorized_user_id:
                 raise Unauthorized(
                     f"Either you weren't logged in or you don't have "
                     f"{action_to_permission[action]} permission "
                     f"on authz resource: {self.index_document['authz']}"
                 )
+            authorized_user_id = (
+                authorized_user_id if isinstance(authorized_user_id, str) else None
```
Avantol13, line 433: why is this necessary? I assume it's b/c the function below expects `None` and not something False-y? Is it even possible for the above function to return `""`?

tests/test_drs.py
```diff
@@ -236,66 +242,595 @@ def test_get_presigned_url_with_query_params(
...
     rsa_private_key,
+    rsa_public_key,
+    indexd_client_accepting_record,
+    mock_arborist_requests,
     google_proxy_group,
     primary_google_service_account,
     cloud_manager,
     google_signed_url,
-    monkeypatch,
 ):
```
Avantol13, line 263: In a new PR can we get some quick, small docstrings about what these tests are actually testing for in more detail? Don't worry about doing it in this PR since the tests are already passing

PR #986, opened 2021-11-29
(PXP-8992): Expire user policies and update Arborist username

johnfrancismccann:
Jira Ticket: [PXP-8992](https://ctds-planx.atlassian.net/browse/PXP-8992)

### New Features
- Set expiration for created user policies when a passport is POSTed to DRS endpoint
- Change username in Arborist when a user POSTs to DRS endpoint and then logs in

### Dependency updates
- Update gen3authz to 1.4.0


fence/sync/sync_users.py
```diff
@@ -1664,6 +1666,9 @@ def _update_authz_in_arborist(
         policy_id_list = []
         policies = []
 
+        if expires is not None:
+            expires = datetime.datetime.utcfromtimestamp(expires)
```
Avantol13, line 1670: see https://github.com/uc-cdis/gen3authz/pull/41#pullrequestreview-818133983

fence/resources/openid/ras_oauth2.py
```diff
@@ -206,7 +206,11 @@ def map_iss_sub_pair_to_user(self, issuer, subject_id, username, email):
                         "from the DRS endpoint. Changing said user's username"
                         f' to "{username}".'
                     )
-                    # TODO also change username in Arborist
+                    flask.current_app.arborist.update_user(
```
Avantol13, line 209: I think we need to handle the error here? or, I see you raise exception in this function. maybe there's not much we can actually do if this fails... but we should at least catch the exception and log useful information (and not implicitly rely on the library to do that). If it's relatively simple, we could try some simple backoff to retry one more time maybe?

Avantol13 (changes requested):

Avantol13 (approved):

PR #988, opened 2021-11-29
feat(passport-exp): handle expiration edge cases [PXP-9235]

Avantol13:
PXP-9235

In order to better handle updating/expiration of authorization from Passports, this PR has number of pieces.

Adjusts overall architecture such that:

1. usersync no longer updates visas
2. Access Token Polling gets updated passport(s) for a user and updates authorization and visas. It also sets expiration for authorization.
3. A frequently-running cronjob handles the removal of stale, expired visas in the database

This PR also adds a variety of refactoring and updating around the above 3 to consolidate code, clarify logic, and fix unit tests.

Relies on https://github.com/uc-cdis/cloud-automation/pull/1806

### New Features
- usersync no longer updates visas and Access Token Polling now both gets updated passport(s) for a user and updates authorization
- A frequently-running cronjob handles the removal of expired visas

### Breaking Changes
- URL Signing when `no_force_sign` query param is provided: Previously Fence would make a decision based off if the data was public **and** `no_force_sign` provided. Fence will now NEVER sign if `no_force_sign` is provided (since the concept of "public" data has been abstracted out of Fence. Data access - public or not - is the sole responsibility of the policy engine). In other words, if `no_force_sign` is provided at the API level, Fence will respect that regardless of whether the resulting URL will actually work. The default Fence behavior should remain the same.

### Bug Fixes
- fix issue with anonymous Google signed URL generation (would fail with unauthenticated requests even if the data was supposed to be accessible)

### Improvements
- various refactoring around Passports


fence/blueprints/data/indexd.py
```diff
@@ -46,10 +45,10 @@
     get_google_app_creds,
     give_service_account_billing_access_if_necessary,
 )
-from fence.resources.ga4gh.passports import get_gen3_users_from_ga4gh_passports
+from fence.resources.ga4gh.passports import sync_gen3_users_authz_from_ga4gh_passports
```
Avantol13, line 48: more descriptive to show that it's changing database and not just "get"-ing information

fence/blueprints/data/indexd.py
```diff
@@ -62,15 +61,21 @@
 
 SUPPORTED_PROTOCOLS = ["s3", "http", "ftp", "https", "gs", "az"]
 SUPPORTED_ACTIONS = ["upload", "download"]
-ANONYMOUS_USER_ID = "anonymous"
+ANONYMOUS_USER_ID = "-1"
```
Avantol13, line 64: the fact that the user_id was not an integer was causing downstream validation projects since User.id is int in the database

fence/blueprints/data/indexd.py
```diff
@@ -517,12 +526,28 @@ def set_acls(self):
         else:
             raise Unauthorized("This file is not accessible")
 
-    def check_authz(self, action, user_ids_from_passports=None):
+    def get_authorized_and_user_id(self, action, user_ids_from_passports=None):
```
Avantol13, line 529: more descriptive / consistent return for this functionality

fence/blueprints/login/ras.py
```diff
@@ -37,76 +39,52 @@ def __init__(self):
         )
 
     def post_login(self, user=None, token_result=None, id_from_idp=None):
-        # TODO: I'm not convinced this code should be in post_login.
```
Avantol13, line 40: for this file, the goal was to reduce duplication and replace this additional custom logic with the already built `sync_gen3_users_authz_from_ga4gh_passports` functionality as much as possible

fence/resources/openid/ras_oauth2.py
```diff
@@ -270,91 +285,31 @@ def map_iss_sub_pair_to_user(self, issuer, subject_id, username, email):
             return iss_sub_pair_to_user.user.username
 
     @backoff.on_exception(backoff.expo, Exception, **DEFAULT_BACKOFF_SETTINGS)
-    def update_user_visas(self, user, pkey_cache, db_session=current_session):
+    def update_user_authorization(self, user, pkey_cache, db_session=current_session):
```
Avantol13, line 288: for this file, the goal was to reduce duplication and replace this additional custom logic with the already built `sync_gen3_users_authz_from_ga4gh_passports` functionality as much as possible

fence/scripting/fence_create.py
```diff
@@ -210,7 +239,6 @@ def init_syncer(
     sync_from_local_yaml_file=None,
     arborist=None,
     folder=None,
-    sync_from_visas=False,
```
Avantol13, line 213: usersync no longer provides the option to sync from visas. This is handled entirely by the access token polling job

fence/sync/passport_sync/ras_sync.py
```diff
@@ -9,44 +9,43 @@ class RASVisa(DefaultVisa):
...
-                if parse_consent_code and consent_group:
-                    full_phsid += "." + consent_group
-                privileges = {"read-storage", "read"}
-                permission_expiration = permission.get("expiration")
-                if permission_expiration and expires <= permission_expiration:
-                    project[full_phsid] = privileges
-                    info["tags"] = {"dbgap_role": permission.get("role", "")}
-        else:
-            # Remove visas if its invalid or expired
-            user.ga4gh_visas_v1 = []
```
Avantol13, line 48: don't mess with db here, handle upstream

fence/sync/sync_users.py
```diff
@@ -1954,196 +1952,24 @@ def parse_user_visas(self, db_session):
 
         return (user_projects, user_info)
 
-    def _sync_visas(self, sess):
```
Avantol13, line 1957: this is handled entirely by access token polling and `sync_gen3_users_authz_from_ga4gh_passports` now

tests/dbgap_sync/conftest.py
```diff
@@ -221,8 +221,8 @@ def add_visa_manually(db_session, user, rsa_private_key, kid):
         "ga4gh_visa_v1": {
             "type": "https://ras.nih.gov/visas/v1",
             "asserted": int(time.time()),
-            "value": "https://nig/passport/dbgap",
-            "source": "https://ncbi/gap",
+            "value": "https://stsstg.nih.gov/passport/dbgap/v1.1",
```
Avantol13, line 224: A lot of tests had to change so that the fake visas were actually "valid", since `sync_gen3_users_authz_from_ga4gh_passports` validates all the Visas

bin/fence_create.py
```diff
@@ -463,6 +461,8 @@ def main():
         verify_bucket_access_group(DB)
     elif args.action == "delete-expired-google-access":
         delete_expired_google_access(DB)
+    elif args.action == "cleanup-expired-ga4gh-information":
+        cleanup_expired_ga4gh_information(DB)
```
Avantol13, line 465: this is the new job that will remove expired visas from the database (since the new logic is that the access token polling will only ADD new visas)

bin/fence_create.py
```diff
@@ -474,8 +474,6 @@ def main():
             sync_from_local_yaml_file=args.yaml,
             folder=args.folder,
             arborist=arborist,
-            sync_from_visas=sync_from_visas,
```
Avantol13, line 477: usersync will no longer handle visas at all

fence/resources/ga4gh/passports.py
```diff
@@ -264,7 +277,7 @@ def validate_visa(raw_visa):
     return decoded_visa
 
 
-def get_or_create_gen3_user_from_iss_sub(issuer, subject_id):
+def get_or_create_gen3_user_from_iss_sub(issuer, subject_id, db_session=None):
```
Avantol13, line 280: there are a handful of places where I'm passing the db_session now instead of recreating new ones

fence/resources/ga4gh/passports.py
```diff
@@ -278,80 +291,90 @@ def get_or_create_gen3_user_from_iss_sub(issuer, subject_id):
...
-            db_session,
-            expires=expiration,
-            policy_prefix="GA4GH.DRS",
-        )
+    # after syncing authorization, perist the visas that were parsed successfully.
+    for visa in ga4gh_visas:
+        if visa not in synced_visas:
+            db_session.remove(visa)
+        else:
+            db_session.add(visa)
```
Avantol13, line 377: this is where visas are added to the database after successful parsing

fence/resources/openid/ras_oauth2.py
```diff
@@ -195,165 +212,108 @@ def map_iss_sub_pair_to_user(self, issuer, subject_id, username, email):
...
+        db_session.commit()
+        return iss_sub_pair_to_user.user.username
+
     @backoff.on_exception(backoff.expo, Exception, **DEFAULT_BACKOFF_SETTINGS)
-    def update_user_visas(self, user, pkey_cache, db_session=current_session):
+    def update_user_authorization(self, user, pkey_cache, db_session=current_session):
         """
         Updates user's RAS refresh token and uses the new access token to retrieve new visas from
-        RAS's /userinfo endpoint and update the db with the new visa.
-        - delete user's visas from db if we're not able to get a new access_token
```
Avantol13, line 276: NOTE: this function **no longer deletes** anything. The new behavior is that the expiration cronjobs and Arborist itself will handle expiration, this only adds

fence/scripting/fence_create.py
```diff
@@ -674,6 +693,48 @@ def delete_users(DB, usernames):
         session.commit()
 
 
+def cleanup_expired_ga4gh_information(DB):
```
Avantol13, line 696: will be a new cronjob

fence/sync/sync_users.py
```diff
@@ -813,6 +813,74 @@ def sync_to_db_and_storage_backend(
 
         sess.commit()
 
+    def sync_to_storage_backend(self, user_project, user_info, sess, expires):
```
Avantol13, line 816: new function that removes the step of syncing to fence's deprecated authorization database and only handles syncing to the storage backends like Google

Avantol13: this is a copy paste of the existing `sync_to_db_and_storage_backend` with as many steps removed as possible that were related to the fence database

fence/resources/ga4gh/passports.py
```diff
@@ -278,80 +291,90 @@ def get_or_create_gen3_user_from_iss_sub(issuer, subject_id):
...
-
-    with flask.current_app.db.session as db_session:
-        syncer.sync_single_user_visas(
-            gen3_user,
-            ga4gh_visas,
-            db_session,
-            expires=expiration,
-            policy_prefix="GA4GH.DRS",
-        )
+    # after syncing authorization, perist the visas that were parsed successfully.
```
johnfrancismccann, line 372:
```suggestion
    # after syncing authorization, persist the visas that were parsed successfully.
```

fence/resources/ga4gh/passports.py
```diff
@@ -44,20 +47,23 @@ def get_gen3_users_from_ga4gh_passports(passports):
...
     logger.info("Getting gen3 users from passports")
-    usernames_from_all_passports = []
+    users_from_all_passports = []
     for passport in passports:
         try:
             # TODO check cache
             cached_usernames = get_gen3_usernames_for_passport_from_cache(passport)
             if cached_usernames:
-                usernames_from_all_passports.extend(cached_usernames)
+                users_from_all_passports.extend(cached_usernames)
```
johnfrancismccann, line 58: Wouldn't we need to additionally query the Fence db here to obtain the user objects corresponding to `cached_usernames`? Otherwise, `users_from_all_passports` will contain a mix of usernames (from the cache) and user objects (from `get_or_create_gen3_user_from_iss_sub`).

https://github.com/uc-cdis/fence/pull/973#discussion_r749467025

fence/resources/ga4gh/passports.py
```diff
@@ -114,15 +122,18 @@ def get_gen3_users_from_ga4gh_passports(passports):
...
         put_gen3_usernames_for_passport_into_cache(
-            passport, usernames_from_current_passport
+            passport, users_from_current_passport
         )
-        usernames_from_all_passports.extend(usernames_from_current_passport)
+        users_from_all_passports.extend(users_from_current_passport)
 
-    return list(set(usernames_from_all_passports))
+    db_session.commit()
+    return list(set(users_from_all_passports))
```
johnfrancismccann, line 136: In the case that there is more than one user object in `users_from_all_passports` corresponding to the same user, will `set(users_from_all_passports)` still regard them as separate objects? In [this example](https://stackoverflow.com/questions/3942303/how-does-a-python-set-check-if-two-objects-are-equal-what-methods-does-an-o):
```
>>> set([Item(1,2), Item(1,2)])
set([Item(1, 2), Item(1, 2)])
```

Maybe a dictionary comprehension could be used:
```
return list({u.id: u for u in users_from_all_passports}.values())
```

fence/blueprints/data/indexd.py
```diff
@@ -85,10 +90,13 @@ def get_signed_url_for_file(
...
 
-    user_ids_from_passports = None
+    user_ids_from_passports = []
     if ga4gh_passports:
-        # TODO change this to usernames
-        user_ids_from_passports = get_gen3_users_from_ga4gh_passports(ga4gh_passports)
+        users_from_passports = sync_gen3_users_authz_from_ga4gh_passports(
+            ga4gh_passports, db_session=db_session
+        )
+        user_ids_from_passports = [user.id for user in users_from_passports]
```
johnfrancismccann, line 98: Since the items in `user_ids_from_passports` get passed to Arborist's `POST /auth/request` [endpoint](https://petstore.swagger.io/?url=https://raw.githubusercontent.com/uc-cdis/arborist/master/docs/openapi.yaml#/auth/post_auth_request), shouldn't this be `user_ids_from_passports = [user.username for user in users_from_passports]` or maybe `usernames_from_passports = [user.username for user in users_from_passports]`?

fence/resources/ga4gh/passports.py
```diff
@@ -12,6 +12,7 @@
 from authutils.token.core import get_iss, get_kid
 from cdislogging import get_logger
 from gen3authz.client.arborist.client import ArboristClient
```
johnfrancismccann, line 14: I don't think `ArboristClient` is needed anymore (same with `from fence.models import User` just below).

fence/blueprints/data/indexd.py
Avantol13: okay, check the latest changes I think I fixed this.. but it uncovered a possibly bigger problem, which is that the use of user_id and username is pretty inconsistent in the naming throughout this PR now. I was trying to use `user_id` more as a generic term for User Identifier (which happens to be their username), but I think it might get confusing as to whether it means User.id (the column from the database). How do you feel about this? I could see some benefit to going back and refactoring most places to `username` instead of `user_id`?

fence/resources/ga4gh/passports.py
Avantol13: there's one single commit after all this stuff here: https://github.com/uc-cdis/fence/blob/cf059a20f5114e7a92f2daa0e5deb2f0362d417a/fence/resources/ga4gh/passports.py#L141

I'll add a comment to clarify that

Avantol13: I'll also make that function 'internal' by prefixing with `_`

fence/blueprints/data/indexd.py
johnfrancismccann: Yeah I feel it's somewhat confusing and that it would be beneficial to use `username`. Maybe with the upcoming bulk endpoints work, we could also aim to consistently use `username` in Arborist and [gen3authz](https://github.com/uc-cdis/gen3authz/blob/3ef8d0cdaf9f6f929a1d4ef3880ea29b81141985/python/src/gen3authz/client/arborist/base.py#L274)? Overall, that would be my vote.

fence/blueprints/login/ras.py
```diff
@@ -37,76 +39,52 @@ def __init__(self):
...
 
         userinfo = flask.g.userinfo
 
-        encoded_visas = []
-
-        try:
-            encoded_visas = flask.current_app.ras_client.get_encoded_visas_v11_userinfo(
-                userinfo
+        global_parse_visas_on_login = config["GLOBAL_PARSE_VISAS_ON_LOGIN"]
+        usersync = config.get("USERSYNC", {})
```
BinamB, line 48: This can be removed since we're not using it anymore

fence/blueprints/login/ras.py
```diff
@@ -37,76 +39,52 @@ def __init__(self):
...
-                ga4gh_visa=encoded_visa,
+                err_msg = "Could not retrieve passport or visas"
+                logger.error("{}: {}".format(e, err_msg))
+                raise
+
+            # now sync authz updates
+            users_from_passports = fence.resources.ga4gh.passports.sync_gen3_users_authz_from_ga4gh_passports(
+                [passport], pkey_cache=pkey_cache, db_session=current_session
             )
-            current_session.add(visa)
```
BinamB, line 108: I might've forgotten details about this but are we not storing visas anymore?

fence/blueprints/login/ras.py
```diff
@@ -116,12 +94,11 @@ def post_login(self, user=None, token_result=None, id_from_idp=None):
         decoded_id = jwt.decode(id_token, verify=False)
 
         # Add 15 days to iat to calculate refresh token expiration time
+        # TODO do they really not provide exp?
```
BinamB, line 97: Nope :upside_down_face:

fence/models.py
```diff
@@ -61,11 +61,15 @@
 def query_for_user(session, username):
     return (
         session.query(User)
-        .filter(func.lower(User.username) == username.lower())
+        .filter(func.lower(User.username) == str(username).lower())
```
johnfrancismccann, line 64: When would `username` not be a string?

Avantol13: mm I think this was leftover from debugging some issue that ended up being user.id's being passed into here. I'll change it back

fence/blueprints/data/indexd.py
Avantol13: Yeah, I came to the same conclusion. I just pushed a commit that refactors the indexd side of things to make that consistent (and I'll fix the cache related stuff in the new cache PR that's in progress). The latest commit(s) also generally try to simplify the signed url generation logic and reduce db calls as well

fence/blueprints/login/ras.py
Avantol13: it happens downstream in `sync_gen3_users_authz_from_ga4gh_passports` now

Avantol13: so we are, just not here

BinamB: oh okay so we're not storing visas if `parse_visas=False`? Well i guess later when we run the job we're getting a new visas anyway?

Avantol13: right, if parse_visas=false we don't sync authorization or parse visas

fence/resources/ga4gh/passports.py
```diff
@@ -278,84 +311,90 @@ def get_or_create_gen3_user_from_iss_sub(issuer, subject_id):
...
-        try:
-            from fence.settings import DB
-        except ImportError:
-            pass
-    storage_creds = config["STORAGE_CREDENTIALS"]
-    syncer = fence.scripting.fence_create.init_syncer(
-        dbgap_config, storage_creds, DB, arborist=arborist_client
+    db_session = db_session or current_session
+    default_args = fence.scripting.fence_create.get_default_init_syncer_inputs()
+    syncer = fence.scripting.fence_create.init_syncer(**default_args)
```
johnfrancismccann, line 382: Although the policies created by `sync_gen3_users_authz_from_ga4gh_passports` are only deleted by an Arborist job when expired, I feel that it might be useful mainly for testing/debugging purposes if `policy_prefix` is also used as `authz_provider`.

Avantol13: ohhh good call, yeah, I sorta forgot about that being used on the client setup. Okay I'll have to take another look

fence/auth.py
```diff
@@ -114,6 +114,7 @@ def set_flask_session_values(user):
 
         if id_from_idp:
             user.id_from_idp = id_from_idp
+            # TODO: update iss_sub mapping table?
```
johnfrancismccann, line 117: For RAS users, the mapping table gets updated [here](https://github.com/uc-cdis/fence/blob/0efaf3b8a6e177b5df8b0b412dd120bd248c1693/fence/resources/openid/ras_oauth2.py#L155) during login.

Since we're not supporting passport issuers other than RAS at the moment, IMO, it might become more clear how to generalize this when we support passports issued by Gen3.

fence/blueprints/data/indexd.py
```diff
@@ -626,21 +623,10 @@ def get_authorized_with_username(self, action, usernames_from_passports=None):
...
 
-    @cached_property
-    def public(self):
-        if self.index_document.get("authz", []):
-            return self.public_authz
-        else:
-            return self.public_acl
-
     @cached_property
     def public_acl(self):
```
Avantol13, line 627: need this for some backwards compatibility

tests/data/test_azure_blob_storage_indexed_file_location.py
```diff
@@ -16,64 +16,39 @@
...
-        ("upload", 5, False, None, "*", ANONYMOUS_USER_ID, False, False),
-        ("upload", 5, None, None, "fake conn str", "some user", True, True),
-        ("upload", 5, None, None, "fake conn str", "some user", False, False),
-        ("upload", 5, None, True, "fake conn str", "some user", True, True),
-        ("upload", 5, None, True, "fake conn str", "some user", False, False),
-        ("upload", 5, None, False, "fake conn str", "some user", True, True),
-        ("upload", 5, None, False, "fake conn str", "some user", False, False),
-        ("upload", 5, True, None, "fake conn str", ANONYMOUS_USER_ID, True, False),
-        ("upload", 5, True, None, "fake conn str", ANONYMOUS_USER_ID, False, False),
+        ("download", 5, None, "fake conn str", "some user", False, False),
```
Avantol13, line 21: remove `public_data` as an option, remove duplicate tests, and adjust expected behavior to meet new expectations. See "Breaking Changes" in the PR description

johnfrancismccann: Maybe the [swagger doc](https://github.com/uc-cdis/fence/blob/8d6e69b773c8e6303b69e11f1c3b3a389870d2f3/openapis/swagger.yaml#L464) could be updated too?

fence/blueprints/login/ras.py
```diff
@@ -37,76 +39,54 @@ def __init__(self):
...
-                ga4gh_visa=encoded_visa,
+                err_msg = "Could not retrieve passport or visas"
+                logger.error("{}: {}".format(e, err_msg))
+                raise
+
+            # now sync authz updates
+            users_from_passports = fence.resources.ga4gh.passports.sync_gen3_users_authz_from_ga4gh_passports(
+                [passport],
+                authz_policy_prefix="POST.LOGIN",
+                pkey_cache=pkey_cache,
```
johnfrancismccann, line 80: I tried logging in and am getting: `NameError: name 'pkey_cache' is not defined` for this line.

johnfrancismccann (approved):

PR #995, opened 2022-01-03
(PXP-8994): Populate mapping table for existing users

johnfrancismccann:
Jira Ticket: [PXP-8994](https://ctds-planx.atlassian.net/browse/PXP-8994)

Populate `iss_sub_pair_to_user` table using `User` table's `id_from_idp` column. Population of `User` table's `id_from_idp` column began with the `2021.12` [release](https://github.com/uc-cdis/fence/compare/2021.11...2021.12), which included [this related PR](https://github.com/uc-cdis/fence/pull/975). 

### New Features
- Populate `iss_sub_pair_to_user` table using `User` table's `id_from_idp` column

### DIFF
--- a/fence/__init__.py
+++ b/fence/__init__.py
@@ -401,7 +401,6 @@ def _set_authlib_cfgs(app):
 
 def _setup_oidc_clients(app):
     oidc = config.get("OPENID_CONNECT", {})
-    app.issuer_to_idp = {}
 
     # Add OIDC client for Google if configured.
     if "google" in oidc:
@@ -426,12 +425,6 @@ def _setup_oidc_clients(app):
             HTTP_PROXY=config.get("HTTP_PROXY"),
             logger=logger,
         )
-        for allowed_issuer in config["GA4GH_VISA_ISSUER_ALLOWLIST"]:
-            if app.ras_client.discovery_url.startswith(allowed_issuer):
-                app.issuer_to_idp[allowed_issuer] = IdentityProvider.ras
-                break
-        else:
-            logger.warn("Could not determine issuer for the RAS OIDC client")
 
     # Add OIDC client for Synapse if configured.
     if "synapse" in oidc:
--- a/fence/__init__.py
+++ b/fence/__init__.py
@@ -21,17 +21,13 @@ from urllib.parse import urlparse
 logger = get_logger(__name__, log_level="debug")
 
 # Load the configuration *before* importing modules that rely on it
-from fence.config import config, DEFAULT_CFG_PATH
+from fence.config import config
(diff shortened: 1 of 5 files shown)


fence/models.py
```diff
@@ -651,6 +653,62 @@ class IssSubPairToUser(Base):
...
+                for allowed_issuer in config["GA4GH_VISA_ISSUER_ALLOWLIST"]:
+                    if discovery_url.startswith(allowed_issuer):
+                        issuer_to_idp[allowed_issuer] = idp
+                        break
+
+        return issuer_to_idp
+
+    ISSUER_TO_IDP = _get_issuer_to_idp()
+
+    del _get_issuer_to_idp
```
Avantol13, line 673:
```suggestion
    # no longer need function since results stored in global
    del _get_issuer_to_idp
```

fence/__init__.py
```diff
@@ -16,6 +16,23 @@
...
 
+# Can't read config yet. Just set to debug for now, else no handlers.
+# Later, in app_config(), will actually set level based on config
+logger = get_logger(__name__, log_level="debug")
+
+# Load the configuration *before* importing modules that rely on it
+from fence.config import config, DEFAULT_CFG_PATH
+from fence.settings import CONFIG_SEARCH_FOLDERS
+
+try:
```
Avantol13, line 27: what prompted this, did you run into a specific issue?

Avantol13: the gen3config library should handle all this complexity with this call https://github.com/uc-cdis/fence/blob/57672f27cabcbe551c49716ef6c0d1adfc7da308/fence/__init__.py#L328-L332

johnfrancismccann: Yeah so the issue with not loading the config here was that `oidc = config.get("OPENID_CONNECT", {})` in `_get_issuer_to_idp` evaluated to an empty dictionary since `_get_issuer_to_idp` is called when `fence/models.py` is imported (which happens before [app_config](https://github.com/uc-cdis/fence/blob/57672f27cabcbe551c49716ef6c0d1adfc7da308/fence/__init__.py#L305) runs).

The `if os.environ.get("FENCE_CONFIG_PATH"):` check is to handle loading `test-fence-config.yaml` when the tests are being run.

Like you're saying, though, I think this can be simplified by putting:
```
config.load(
    config_path=os.environ.get("FENCE_CONFIG_PATH"),
    search_folders=CONFIG_SEARCH_FOLDERS
)
```
in place of the whole try/except.

Avantol13: Gotcha, yeah I'd rather contain this logic within the gen3config library, so changing to that .load call would make me feel better

Avantol13 (approved):

PR #997, opened 2022-01-18
Fix/visa expiration

Avantol13:
### Bug Fixes
- ensure we log an error but continue when the expiration cannot be correctly parsed for a RAS Visa


fence/sync/passport_sync/ras_sync.py
```diff
@@ -39,7 +39,16 @@ def _parse_single_visa(
                 if parse_consent_code and consent_group:
                     full_phsid += "." + consent_group
```
johnfrancismccann, line 40: If say an integer was supplied for `"phs_id"` or `"consent_group"`, this concatenation would throw an error. Maybe it could also be wrapped within a `try` block?

Avantol13: or just cast to string should work too, I think I'll do that

Avantol13: take another look and see if this makes sense as an alternative

johnfrancismccann: I would have thought that if `"phs_id"` or `"consent_group"` is an integer, then technically the study cannot be determined (i.e. we always expect them to be strings beginning with `"phs"` or `"c"`).

Casting also means that a JSON `null` value in the passport could result in an Arborist resource being created such as `/programs/None` or `/programs/DEV/projects/None`.

Avantol13: hmm... good point. We may not always expect that they are formatted that way though. It seems weird to try to cast to integer and raise an exception if it succeeds... The concatenation assumes strings, I guess what I can do is first check if they even exist

johnfrancismccann (approved):

PR #999, opened 2022-01-24
feat(passport): add cache [PXP-8838]

Avantol13:
https://ctds-planx.atlassian.net/browse/PXP-8838
  
### Improvements
- Passport cache for substantially increasing speed of subsequent parsing of the same, valid passport

### DIFF
--- a/tests/test-fence-config.yaml
+++ b/tests/test-fence-config.yaml
@@ -635,7 +635,7 @@ GA4GH_VISA_V1_CLAIM_REQUIRED_FIELDS:
     - "https://stsstg.nih.gov/passport/dbgap/v1.1"
   source:
     - "https://ncbi.nlm.nih.gov/gap"
-EXPIRED_AUTHZ_REMOVAL_JOB_FREQ_IN_SECONDS: 300
+EXPIRED_AUTHZ_REMOVAL_JOB_FREQ_IN_SECONDS: 1
 # Global sync visas during login
 # None(Default): Allow per client i.e. a fence client can pick whether or not to sync their visas during login with parse_visas param in /authorization endpoint
 # True: Parse for all clients i.e. a fence client will always sync their visas during login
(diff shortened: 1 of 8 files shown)


tests/test-fence-config.yaml
```diff
@@ -635,7 +635,7 @@ GA4GH_VISA_V1_CLAIM_REQUIRED_FIELDS:
     - "https://stsstg.nih.gov/passport/dbgap/v1.1"
   source:
     - "https://ncbi.nlm.nih.gov/gap"
-EXPIRED_AUTHZ_REMOVAL_JOB_FREQ_IN_SECONDS: 300
+EXPIRED_AUTHZ_REMOVAL_JOB_FREQ_IN_SECONDS: 1
```
johnfrancismccann, line 638: Is the [frequency of the cronjob](https://github.com/uc-cdis/cloud-automation/blob/master/kube/services/jobs/fence-cleanup-expired-ga4gh-info-cronjob.yaml#L7) going to be changed as well?

johnfrancismccann:
With passport caching, the expiration of policies in Arborist now doesn't get extended on a cache hit. This is probably not a very realistic flow and not necessarily a negative side effect, but the behavior could change for a user that:

1. POSTs a valid passport at 1pm
2. POSTs the same passport again at 1:30pm
3. Generates a Gen3 access token from an API key and GETs the `/data/download` endpoint at 2:15pm

Avantol13:
> With passport caching, the expiration of policies in Arborist now doesn't get extended on a cache hit. 

I don't think it should get extended though. If the passport expires in an hour and there's not a new passport, we shouldn't extend it. If we DO get a new passport with new expiration, then it would miss the cache and update Arborist like normal. 

The cache is the whole passport, so any extension in `exp` would be a different cache entry / update Arborist.

PR #1005, opened 2022-02-11
(PXP-9480): Improve sync_single_user_visas performance

johnfrancismccann:
Jira Ticket: [PXP-9480](https://ctds-planx.atlassian.net/browse/PXP-9480)

Improvements to `sync_single_user_visas` will result in better performance for DRS endpoint, RAS login, and Access Token Polling.

Corresponding PRs
- [gen3authz](https://github.com/uc-cdis/gen3authz/pull/44)
- [Arborist](https://github.com/uc-cdis/arborist/pull/143)

### Improvements
- Significantly improve `sync_single_user_visas` performance by use of a single Arborist policy to represent access

### Dependency updates
- Bump `gen3authz` to `1.5.0`


fence/sync/sync_users.py
```diff
@@ -1813,100 +1820,69 @@ def _update_authz_in_arborist(
...
+                    policy_hash = self._hash_policy_contents(
+                        ordered_roles, ordered_resources
+                    )
+                    if not self._create_arborist_policy(
+                        policy_hash,
+                        ordered_roles,
+                        ordered_resources,
+                        skip_if_exists=True,
+                    ):
+                        return False
```
Avantol13, line 1848: let's add some logs here so if this happens we have some stuff to look back at - we may also want to update the error message logged from the calling function when this returns False - I'm not sure if it says "couldn't update Arborist resources" we should make it generic like "something went wrong when updating Arborist". 

We also may want to consider if it's a problem if this breaks part way through. I think it's fine.. but we need to make sure that a partial update doesn't unintentionally have a side-effect.

And does it make sense to completely die immediately upon a single error or should we continue and attempt the rest of the updates and return False at the end if any failed?

I think usersync has had that philosphy, update everything we can and don't die part way through, then return false exit(1), whatever, at the end if anything went wrong. That way, if there's a critical authorization update that for whatever reason would happen _after_ some error, it would still go through

fence/sync/sync_users.py
```diff
@@ -1813,100 +1820,69 @@ def _update_authz_in_arborist(
...
+                    if not self._grant_arborist_policy(
+                        username, policy_hash, expires=expires
+                    ):
+                        return False
+            else:
+                for roles, resources in unique_policies.items():
```
Avantol13, line 1854: so all this is the same as before, just using unique_policies now?

fence/sync/sync_users.py
```diff
@@ -1940,22 +1916,242 @@ def _update_authz_in_arborist(
...
+
+        canonical_roles = ",".join(tuple(escape(r) for r in roles))
+        canonical_resources = ",".join(tuple(escape(r) for r in resources))
+        canonical_policy = f"{canonical_roles},,f{canonical_resources}"
+        policy_hash = hashlib.sha256(canonical_policy.encode("utf-8")).hexdigest()
+        return policy_hash
+
+    def _grant_arborist_policy(self, username, policy_id, expires=None):
+        """
+        Wrapper around gen3authz's grant_user_policy with additional logging
```
Avantol13, line 2115: all these wrapper functions, while internal to sync, might be better organized into a utils file, maybe in fence/sync/utils.py?

fence/sync/sync_users.py
```diff
@@ -1940,22 +1916,242 @@ def _update_authz_in_arborist(
...
+        """
+        roles_to_resources = collections.defaultdict(list)
+        for study, roles in user_project_info.items():
+            ordered_roles = tuple(sorted(roles))
+            study_authz_paths = self._dbgap_study_to_resources.get(study, [study])
+            if study in project_to_authz_mapping:
+                study_authz_paths = [project_to_authz_mapping[study]]
+            roles_to_resources[ordered_roles].extend(study_authz_paths)
+
+        policies = {}
```
Avantol13, line 1954: should we use an OrderedDict here?

Avantol13: or maybe this doesn't really matter so long as the roles and resources within this are ordered. B/c we'll hash each entry separately

fence/sync/sync_users.py
```diff
@@ -1940,22 +1916,242 @@ def _update_authz_in_arborist(
...
+            )
             return False
 
+        if response_json is None:
+            self.logger.info("policy `{}` already exists in Arborist".format(policy_id))
+        else:
+            self.logger.info("created policy `{}` in Arborist".format(policy_id))
+        return True
+
+    def _hash_policy_contents(self, roles, resources):
```
Avantol13, line 2092: for clarity and to avoid any future issues with using this in a reliable way, can we name the inputs `ordered_roles` and `ordered_resources` like you have elsewhere? Alternatively, we could do the `sorted()` in here

Avantol13 (changes requested):

fence/sync/sync_users.py
johnfrancismccann: Yep, the block nested under `for resource in resources:` is almost the same as the old block nested under ` for path in paths:`, except that there's now no code for policy prefixes (which weren't used by usersync anyway) and the granting of a policy is now done with a gen3authz wrapper function.

fence/sync/sync_users.py
johnfrancismccann: I think the worst thing that could happen with a partial update is that there are resources and/or policies lying around in Arborist that aren't actually granted to any users.

I think it would be better to continue and attempt the rest of the updates. If Arborist was temporarily down only during the request for policy creation, the granting of a policy could succeed if the policy already existed from a prior DRS request.

fence/sync/sync_users.py
johnfrancismccann: The only challenge with this is that these wrapper functions use instance variables `self.arborist_client` and `self.logger`.

fence/sync/sync_users.py
johnfrancismccann: How do you feel about having the logs be in those wrapper functions? Might be a personal preference, but I was trying to keep logging and try/except logic in the wrapper functions in the hope that it would make the higher-level code more concise.

Avantol13 (approved):

PR #1006, opened 2022-02-15
PXP-4235 Add all the user's scopes to API keys

paulineribeyre:
Jira Ticket: [PXP-4235](https://ctds-planx.atlassian.net/browse/PXP-4235) (not exactly, but i think this is better?)

### Improvements
- Add all the user's scopes to API keys

### DIFF
--- a/fence/blueprints/storage_creds/api.py
+++ b/fence/blueprints/storage_creds/api.py
@@ -93,6 +93,13 @@ class ApiKeyList(Resource):
         for s in scope:
             if s not in config["USER_ALLOWED_SCOPES"]:
                 flask.abort(400, "Scope {} is not supported".format(s))
+
+        # add all scopes from the user's access token
+        scope.extend(current_token["scope"])
+
+        # a token created using an API key cannot be used to create a new API key
+        scope = [e for e in set(scope) if e != "credentials"]
+
         max_ttl = config.get("MAX_API_KEY_TTL", 2592000)
         expires_in = min(int(flask.request.args.get("expires_in", max_ttl)), max_ttl)
         api_key, claims = create_api_key(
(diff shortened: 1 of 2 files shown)


fence/blueprints/storage_creds/api.py
```diff
@@ -93,6 +93,13 @@ def post(self):
         for s in scope:
             if s not in config["USER_ALLOWED_SCOPES"]:
                 flask.abort(400, "Scope {} is not supported".format(s))
+
+        # add all scopes from the user's access token
+        scope.extend(current_token["scope"])
+
+        # a token created using an API key cannot be used to create a new API key
+        scope = [e for e in set(scope) if e != "credentials"]
```
Avantol13, line 101: I think we also need to ensure the scope is in our ALLOWED_SCOPES for users

paulineribeyre: do you mean in addition to this https://github.com/uc-cdis/fence/blob/d316a9fbb8c5b7b4ae09fbfb70e129db3934c773/fence/blueprints/storage_creds/api.py#L93-L95
we should make sure not to add any scopes from `current_token["scope"]`  that are not in `USER_ALLOWED_SCOPES`? i figured if the scopes are in the token, then we know they are allowed - but i could add that

Avantol13: Yeah it wouldn't hurt and would avoid edge cases where we update that configuration and there are still tokens out. Like if we want to remove some scope, without checking here if the user already had an access token with it, they could get another API key with it :man_shrugging: It's maybe a minor risk but might as well mitigate it since

paulineribeyre: updated that

Avantol13 (approved):

PR #1009, opened 2022-02-23
Create image_build_push.yaml

Avantol13:
### Improvements
- use  central workflow  github actions for image build and push

### DIFF
--- a/.github/workflows/image_build_push.yaml
+++ b/.github/workflows/image_build_push.yaml
@@ -5,7 +5,7 @@ on: push
 jobs:
   ci:
     name: Build Image and Push to Quay 
-    uses: uc-cdis/.github/.github/workflows/image_build_push.yaml@feat/build-push-image
+    uses: uc-cdis/.github/.github/workflows/image_build_push.yaml@master
     secrets:
       ECR_AWS_ACCESS_KEY_ID: ${{ secrets.ECR_AWS_ACCESS_KEY_ID }}
       ECR_AWS_SECRET_ACCESS_KEY: ${{ secrets.ECR_AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/image_build_push.yaml
```diff
@@ -0,0 +1,13 @@
...
+
+jobs:
+  ci:
+    name: Build Image and Push to Quay 
+    uses: uc-cdis/.github/.github/workflows/image_build_push.yaml@feat/build-push-image
+    secrets:
+      ECR_AWS_ACCESS_KEY_ID: ${{ secrets.ECR_AWS_ACCESS_KEY_ID }}
+      ECR_AWS_SECRET_ACCESS_KEY: ${{ secrets.ECR_AWS_SECRET_ACCESS_KEY }}
+      QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+      QUAY_ROBOT_TOKEN: ${{ secrets.QUAY_ROBOT_TOKEN }}
```
paulineribeyre, line 13: if these are set at the org level, is there no way to read them directly from the reusable workflow instead of passing them as parameters?

Avantol13: It's seems like not, I did try. I removed the "required" secrets from the central workflow and here (and hoped that it would automatically use the org level ones) but somehow (I assume because of the way the reusable workflows are separate), it didn't work. I also tried setting the secrets in the central repo and not passing them from here (also didn't work).

It's kinda annoying, but at least it's explicit that secrets from this repo are being used (even if they're at the org level). I think we're still going to limit the secrets at the org level to only approved repos.

paulineribeyre (dismissed):

paulineribeyre (approved):

emalinowski (approved):

PR #1017, opened 2022-03-26
PXP-9500 Generic OIDC support

paulineribeyre:
Jira Ticket: [PXP-9500](https://ctds-planx.atlassian.net/browse/PXP-9500)

**This has not been tested in QA yet**

### New Features
- Generic OIDC support

### Improvements
- Refactor so the list of supported OIDC IDPs is not hardcoded anymore
- Update unit tests so all supported OIDC IDPs are tested


tests/login/test_conftest_idps.py
```diff
@@ -0,0 +1,36 @@
...
+    If you added an IDP and this test fails, do not edit this test. Either the
+    new IDP is not in LOGIN_IDPS, or the new file names do not match the
+    convention and cannot be parsed by this test.
+    """
+    current_dir = os.path.dirname(os.path.realpath(__file__))
+    err_msg = "For IDP file '{}': IDP '{}' should be added to the tests' 'LOGIN_IDPS' to be tested"
+
+    subdir = "resources/openid"
+    path = os.path.join(current_dir, "../../fence", subdir)
+    for _, _, files in os.walk(path):
```
paulineribeyre, line 20: I guess `LOGIN_IDPS` could just be automatically generated this way, which might be even more overkill than this test :laughing:

tests/login/test_conftest_idps.py
```diff
@@ -0,0 +1,36 @@
...
+
+def test_contfest_idps():
+    """
+    This test checks that any newly added custom OIDC IDP has been added to
+    the list of IDPs to test (LOGIN_IDPS).
+
+    If you added an IDP and this test fails, do not edit this test. Either the
+    new IDP is not in LOGIN_IDPS, or the new file names do not match the
+    convention and cannot be parsed by this test.
+    """
```
paulineribeyre, line 14: Todo: Or the new IdP should be added to the test config file

fence/config-default.yaml
```diff
@@ -88,11 +88,25 @@ ENABLE_DB_MIGRATION: true
...
+    client_id: ''
+    client_secret: ''
+    redirect_url: '{{BASE_URL}}/login/some_idp/login' # replace IDP name
+    # use `discovery` to configure IDPs that do not expose a discovery
+    # endpoint. One of `discovery_url` or `discovery` should be configured
+    discovery_url: ''
+    discovery:
+      authorization_endpoint: ''
+      token_endpoint: ''
+      jwks_uri: ''
```
paulineribeyre, line 106: ImmPort does not expose a discovery endpoint, that's why I added `discovery`. They also do not expose a JWKs endpoint, so I might have to allow configuring the keys directly in the config as well. If you think it would be better to implement ImmPort as another custom OIDC IDP instead of bloating the generic config, let me know

fence/__init__.py
```diff
@@ -399,71 +398,65 @@ def _set_authlib_cfgs(app):
...
+                logger=logger,
+            )
+        elif idp == "fence":
+            app.fence_client = OAuthClient(**settings)
+        else:  # generic OIDC implementation
+            client = Oauth2ClientBase(
+                settings=settings,
+                logger=logger,
+                HTTP_PROXY=config.get("HTTP_PROXY"),
+                idp=settings.get("name") or idp.title(),
```
Avantol13, line 457: if people are writing this in the user.yaml, it's possible there might be spaces, we should remove them before this

fence/config-default.yaml
Avantol13: I'm generally okay with this. maybe it would be clearer if the default discovery_url had a value though. It's a standard URL https://server.com/.well-known/openid-configuration

fence/resources/openid/idp_oauth2.py
```diff
@@ -89,62 +90,90 @@ def get_value_from_discovery_doc(self, key, default_value):
...
-                "{} ERROR from {} API, could not retrieve `{}` from response {}. Defaulting to {}".format(
-                    self.discovery_doc.status_code,
-                    self.idp,
-                    key,
-                    self.discovery_doc.json(),
-                    default_value,
-                )
-            )
-            return_value = default_value
+        else:  # no `discovery_url`, try to use `discovery` config instead
```
Avantol13, line 125:
```suggestion
        # no `discovery_url`, try to use `discovery` config instead
        else:
```

fence/blueprints/login/__init__.py
```diff
@@ -226,6 +221,45 @@ def get_login_providers_info():
...
+            client=getattr(flask.current_app, f"{idp_name}_client"),
+            username_field=settings.get("user_id_field", "sub"),
+            email_field=settings.get("email_field", "email"),
+            id_from_idp_field=settings.get("id_from_idp_field", "sub"),
+        )
+
+    return type(
+        f"GenericCallback_{idp_name}",
+        (DefaultOAuth2Callback,),
+        {"__init__": initiate},
```
Avantol13, line 259: 1. this is cool and I didn't know about this `type` dynamic class thing
2. We should add a detailed comment about what this is doing
3. I'm nervous about the 3rd param where we're passing the function. I have a feeling that it's using the same function object across multiple instances of any generic classes.. which, might be okay but really makes me anxious about any accidental shared state. Can we confirm that when there are 2+ generic OIDC clients that they get initialized correctly without any shared things. It actually might be fine b/c you `def` the function within this overall `create` function :thinking:

paulineribeyre: 3. Yes, that's a good call, that's the use case that delayed this PR for 2+ weeks - I configured 2 generic IDPs in the unit tests to make sure this worked, and it didn't until I moved the class creation to its own function (`createLoginClass` and `createCallbackClass`). That's the use case I was crying to you about the other day 😄

Avantol13: re: 3, this is all making sense now. okay so this is the version that solves that problem

paulineribeyre: 2. i added some details but i'm not sure what you are looking for. lmk if something is missing

fence/__init__.py
paulineribeyre: updated in the line below, but not in this line because `idp` is only used as a display name in logs and error messages

Avantol13 (approved):

PR #1020, opened 2022-04-14
Feat/custom data access

BinamB:
### New Features
- Add non-dbgap whitelist support in usersync.


fence/sync/sync_users.py
```diff
@@ -369,11 +373,15 @@ def _match_pattern(filepath, encrypted=True):
...
-        pattern += "$"
-        return re.match(pattern, os.path.basename(filepath))
+        id_patterns.append("authentication_file_phs(\d{6}).(csv|txt)")
+        for pattern in id_patterns:
+            pattern = r"{}".format(pattern)
+            if encrypted:
+                pattern += ".enc"
+            pattern += "$"
+            if re.match(pattern, os.path.basename(filepath)):
+                return True
```
Avantol13, line 383: is this intended to return early on the first matched pattern? Or should this only return True if ALL patterns match that are passed in?

fence/sync/sync_users.py
```diff
@@ -527,7 +543,11 @@ def _parse_csv(self, file_dict, sess, dbgap_config={}, encrypted=True):
                         continue
 
                     phsid_privileges = {}
-                    phsid = row.get("phsid", "").split(".")
+                    if self.non_dbGaP_whitelist:
```
Avantol13, line 546: should you check `allow_non_dbgap_whitelist`?

tests/dbgap_sync/test_user_sync.py
```diff
@@ -633,7 +711,7 @@ def mock_merge(dbgap_servers, sess):
 
     # this function will be called once for each sftp server
     # the test config file has 2 dbgap sftp servers
```
Avantol13, line 713:
```suggestion
    # the test config file has 3 dbgap sftp servers
```

tests/test-fence-config.yaml
```diff
@@ -662,6 +676,10 @@ EXPIRED_AUTHZ_REMOVAL_JOB_FREQ_IN_SECONDS: 1
 GLOBAL_PARSE_VISAS_ON_LOGIN:
 # Settings for usersync with visas
 USERSYNC:
+  non_dbgap_whitelist_enabled: false
```
Avantol13, line 679: why have both this and the per-sftp config `allow_non_dbGaP_whitelist`?

tests/test-fence-config.yaml
```diff
@@ -662,6 +676,9 @@ EXPIRED_AUTHZ_REMOVAL_JOB_FREQ_IN_SECONDS: 1
 GLOBAL_PARSE_VISAS_ON_LOGIN:
 # Settings for usersync with visas
 USERSYNC:
+  sync_from_visas: false
```
Avantol13, line 679: I think these were removed since our merge of feat/passport, can you double check we actually need this and the one below?

Avantol13 (changes requested):

tests/dbgap_sync/test_user_sync.py
```diff
@@ -74,16 +74,25 @@ def test_sync_incorrect_user_yaml_file(syncer, monkeypatch, db_session):
     assert syncer.arborist_client.create_policy.not_called()
 
 
+@pytest.mark.parametrize("allow_non_dbgap_whitelist", [True, False])
 @pytest.mark.parametrize("syncer", ["google", "cleversafe"], indirect=True)
-@pytest.mark.parametrize("parse_consent_code_config", [False, True])
+@pytest.mark.parametrize("parse_consent_code_config", [True, False])
```
Avantol13, line 79: Is this just to be consistently passing true then false or was there some real issue when it was ordered the other way?

BinamB: oh woops. I was messing around with it and reversed the order. I'll change it back.

fence/sync/sync_users.py
```diff
@@ -369,11 +370,18 @@ def _match_pattern(filepath, encrypted=True):
...
-        return re.match(pattern, os.path.basename(filepath))
+        id_patterns.append("authentication_file_phs(\d{6}).(csv|txt)")
+        for pattern in id_patterns:
+            pattern = r"{}".format(pattern)
+            if encrypted:
+                pattern += ".enc"
+            pattern += "$"
+            pattern = pattern.encode().decode(
+                "unicode_escape"
+            )  # when converting from yaml, python reads it as Python string literal. So "\" turns into "\\" which messes with the regex match
```
Avantol13, line 381: can you move this above the line, I generally try to avoid inline comments unless absolutely necessary

Avantol13: also, why is there YAML at all now?

fence/sync/sync_users.py
```diff
@@ -527,9 +542,13 @@ def _parse_csv(self, file_dict, sess, dbgap_config={}, encrypted=True):
                         continue
 
                     phsid_privileges = {}
-                    phsid = row.get("phsid", "").split(".")
+                    if dbgap_config.get("allow_non_dbGaP_whitelist", False):
+                        phsid = row.get("phsid", row.get("project_id", "")).split(".")
```
Avantol13, line 546: should this be configurable? Or I guess we can just require that people either use `phsid` or `project_id`, but this feels like something that could be a cfg

fence/sync/sync_users.py
```diff
@@ -1236,6 +1260,24 @@ def _get_user_permissions_from_csv_list(
         )
         return user_projects, user_info
 
+    def _merge_multiple_local_csv_files(
+        self, dbgap_file_list, encrypted, dbgap_configs, session
+    ):
+
```
Avantol13, line 1266: maybe a quick docstring here?

fence/sync/sync_users.py
BinamB: Oh should I keep the comment and instead of "yaml", call it fence-config or should I just completely get rid of the comment?

Avantol13: I think the comment is helpful, let's clarify that it's YAML from the fence-config

fence/sync/sync_users.py
Avantol13: yeah that's fine

Avantol13 (dismissed):

Avantol13 (approved):

PR #1025, opened 2022-05-11
fix(gs): ensure inmem cache exp getschecked

Avantol13:
### Bug Fixes
- ensure inmem gs cache of sa keys has exp checked


fence/blueprints/data/indexd.py
```diff
@@ -1075,11 +1075,17 @@ def _generate_google_storage_signed_url(
...
         is_cached = False
-
         if proxy_group_id in self._assume_role_cache_gs:
-            private_key, key_db_entry = self._assume_role_cache_gs.get(proxy_group_id)
-            is_cached = True
-        elif hasattr(flask.current_app, "db"):
+            raw_private_key, raw_key_db_entry = self._assume_role_cache_gs.get(
+                proxy_group_id
+            )
+            if raw_key_db_entry and raw_key_db_entry.expires > expiration_time:
```
Avantol13, line 1082: I'm confused, are you sure this works? isn't `raw_key_db_entry` a string? How are you accessing `.expires`?

Avantol13: I think you have to get the expiration we added to the cache and use that to compare

tests/data/test_indexed_file.py
```diff
@@ -381,6 +387,67 @@ def test_internal_get_signed_url_no_location_match(
...
+    ):
+        with mock.patch.object(
+            cirrus.google_cloud.utils,
+            "get_signed_url",
+            return_value="https://cloud.google.com/compute/url",
+        ):
+            indexed_file = IndexedFile(file_id="some id")
+            google_object = GoogleStorageIndexedFileLocation("gs://some/location")
+            keydbentry = UserGoogleAccountToProxyGroup()
+            keydbentry.expires = 10
```
Avantol13, line 432: this is a problem, that doesn't actually exist on the real object. you're artificially creating that property in the test

BinamB: Oh wait, isnt this it? https://github.com/uc-cdis/fence/blob/b8822d79137a1d2ac83806308c31497c2552b40c/fence/models.py#L367

fence/blueprints/data/indexd.py
```diff
@@ -1090,13 +1086,25 @@ def _generate_google_storage_signed_url(
...
                         json.loads(cache.gcp_key_db_entry),
+                        cache.expires_at,
                     )
                     self._assume_role_cache_gs[proxy_group_id] = rv
                     private_key, key_db_entry = self._assume_role_cache_gs.get(
                         proxy_group_id
                     )
                     is_cached = True
 
+        if proxy_group_id in self._assume_role_cache_gs:
```
Avantol13, line 1097: we need this before the database, we want to check the in-memory cache first - then fallback to database

fence/blueprints/data/indexd.py
```diff
@@ -1090,13 +1086,25 @@ def _generate_google_storage_signed_url(
...
+            raw_private_key, raw_key_db_entry = self._assume_role_cache_gs.get(
+                proxy_group_id
+            )
+            if raw_key_db_entry and raw_key_db_entry.expires > expiration_time:
```
Avantol13, line 1101: this isn't right, there is no `.expires` on that entry b/c it should be a string. Did you try this outside the unit test? I expect it to fail here. What we need is to pull the expiration out of the cache and use that to compare here.

```
           raw_private_key, raw_key_db_entry, expires_at = self._assume_role_cache_gs.get(
                proxy_group_id
            )
            if raw_key_db_entry and expires_at > expiration_time:
```

And then in this line: https://github.com/uc-cdis/fence/pull/1025/files#diff-d3860215a35de5dbc7fc8602855eeeda3d34aba7c721b7f3003ee49cb218453eR1128
we need to make sure to add the expiration into the cache like we did here: https://github.com/uc-cdis/fence/pull/1025/files#diff-d3860215a35de5dbc7fc8602855eeeda3d34aba7c721b7f3003ee49cb218453eR1089

fence/blueprints/data/indexd.py
```diff
@@ -1188,13 +1188,22 @@ def _generate_google_storage_signed_url(
...
 
         if proxy_group_id in self._assume_role_cache_gs:
-            private_key, key_db_entry = self._assume_role_cache_gs.get(proxy_group_id)
-            is_cached = True
-        elif hasattr(flask.current_app, "db"):
+            (
+                raw_private_key,
+                raw_key_db_entry,
+                expires_at,
+            ) = self._assume_role_cache_gs.get(proxy_group_id)
```
Avantol13, line 1198: I think this might actually error if .get returns None. I think we need to default to `(None, None, None)`

```
>>> test = {}
>>> foo, bar, fizz = test.get("buzz")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: cannot unpack non-iterable NoneType object
```

fence/blueprints/data/indexd.py
```diff
@@ -1205,11 +1214,14 @@ def _generate_google_storage_signed_url(
...
                     )
                     self._assume_role_cache_gs[proxy_group_id] = rv
-                    private_key, key_db_entry = self._assume_role_cache_gs.get(
-                        proxy_group_id
-                    )
+                    (
+                        private_key,
+                        key_db_entry,
+                        expires_at,
+                    ) = self._assume_role_cache_gs.get(proxy_group_id)
```
Avantol13, line 1224: see above comment

fence/blueprints/data/indexd.py
Avantol13: it is a string? See line https://github.com/uc-cdis/fence/blob/00f6fda595617c24bc932a1a3028481d1cb37c47/fence/blueprints/data/indexd.py#L1216 

I'm still confused how this is working

Avantol13: okay, I think there's some inconsistency in the types that can end up in the cache

https://github.com/uc-cdis/fence/blob/00f6fda595617c24bc932a1a3028481d1cb37c47/fence/blueprints/data/indexd.py#L1216
is a dictionary.

https://github.com/uc-cdis/fence/blob/00f6fda595617c24bc932a1a3028481d1cb37c47/fence/blueprints/data/indexd.py#L1249 
is a `GoogleServiceAccountKey` type. 

So the use of .expires is problematic if it's a dictionary. I think we need the types to be consistent so this always works

fence/blueprints/data/indexd.py
```diff
@@ -1188,13 +1188,22 @@ def _generate_google_storage_signed_url(
...
+                expires_at,
+            ) = self._assume_role_cache_gs.get(proxy_group_id, (None, None, None))
+            if raw_key_db_entry and raw_key_db_entry.expires > expiration_time:
+                is_cached = True
+                private_key = raw_private_key
+                key_db_entry = raw_key_db_entry
+            else:
+                del self._assume_role_cache_gs[proxy_group_id]
+
+        if is_cached == False and hasattr(flask.current_app, "db"):
```
Avantol13, line 1206:
```suggestion
        if not is_cached and hasattr(flask.current_app, "db"):
```

BinamB (approved):

PR #1029, opened 2022-06-23
PXP-8028 Use Alembic for DB migrations

paulineribeyre:
Jira Ticket: [PXP-8028](https://ctds-planx.atlassian.net/browse/PXP-8028)

Requires https://github.com/uc-cdis/cloud-automation/pull/1991

For now, migrations must still be run manually - we don't deprecate the `fence-create migrate` command and `ENABLE_DB_MIGRATION` setting yet. This will be done at a later time and will require these cloud-automation changes: https://github.com/uc-cdis/cloud-automation/pull/1988

### New Features
- Use Alembic for DB migrations


migrations/env.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+
+    with connectable.connect() as connection:
+        context.configure(connection=connection, target_metadata=target_metadata)
+
+        with context.begin_transaction():
+            print("Locking database to ensure only 1 migration runs at a time")
+            connection.execute("SELECT pg_advisory_xact_lock(100);")
+            context.run_migrations()
+        print("Releasing database lock")
```
paulineribeyre, line 69: This prevents 2 fence instances from trying to migrate the same db at the same time, but does not prevent a usersync (or other) job from updating the DB while a migration is running

.travis.yml
```diff
@@ -25,7 +25,7 @@ before_script:
   - cd -
 
 script:
-  - poetry run pytest -vv --cov=fence --cov-report xml tests
+  - poetry run pytest -vv --cov=fence --cov=migrations/versions --cov-report xml tests
```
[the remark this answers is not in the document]
paulineribeyre: part of `migrations/env.py` does not run during the tests and that's normal, adding it to the coverage reports brings the coverage down for no reason, i think it's enough to check coverage in the actual migrations like we do in other repos

migrations/env.py
```diff
@@ -0,0 +1,75 @@
+from alembic import context
```
Avantol13, line 1: is this file based on an official example? If so can we link that in a docstring at the top of the file with a quick blurb about what we added?

migrations/env.py
Avantol13: can you add this GH comment as a comment in the code?

migrations/env.py
paulineribeyre: This file is auto-generated by Alembic, you will also see it in the other repos that use Alembic. I can still document what was added if you think it's necessary

migrations/env.py
```diff
@@ -0,0 +1,75 @@
...
+        config.get_section(config.config_ini_section),
+        prefix="sqlalchemy.",
+        poolclass=pool.NullPool,
+    )
+
+    with connectable.connect() as connection:
+        context.configure(connection=connection, target_metadata=target_metadata)
+
+        with context.begin_transaction():
+            print("Locking database to ensure only 1 migration runs at a time")
```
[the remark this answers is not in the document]
paulineribeyre: alembic has its own logger that is interfering with cdislogging. I can use `logging.getLogger` directly though, so I changed the print statements to that

migrations/env.py
```diff
@@ -0,0 +1,92 @@
...
+    with connectable.connect() as connection:
+        context.configure(connection=connection, target_metadata=target_metadata)
+
+        with context.begin_transaction():
+            logger.info("Locking database to ensure only 1 migration runs at a time")
+            # This prevents 2 fence instances from trying to migrate the same
+            # DB at the same time, but does not prevent a job (such as
+            # usersync) from updating the DB while a migration is running.
+            # TODO lock the DB for all processes during migrations
+            connection.execute("SELECT pg_advisory_xact_lock(100);")
```
Avantol13, line 84: Maybe the lock number should be part of the Fence configuration? 

Also we're coupling ourselves more with postgres directly here (I know we use some postgres specific stuff in sqlalchemy so idk how big of a deal this is).

paulineribeyre: > Maybe the lock number should be part of the Fence configuration?

i could make it a global var but i am not sure of the benefit of making it configurable, unless we're considering having the Fence jobs read from the config and lock as well?

>Also we're coupling ourselves more with postgres directly here (I know we use some postgres specific stuff in sqlalchemy so idk how big of a deal this is).

that's true, i could add a check so it doesn't attempt to lock if not using postgres

Avantol13 (dismissed):

Avantol13 (approved):

johnfrancismccann (approved):

PR #1033, opened 2022-08-03
PXP-10113 Client credentials grant

paulineribeyre:
Jira Ticket: [PXP-10113](https://ctds-planx.atlassian.net/browse/PXP-10113)

⚠️ "Client credentials" tokens are not supported by all endpoints, only by the ones needed for the driving use cases

### New Features
- Support the OAuth2 "client credentials" grant
- Support "client credentials" tokens that are not linked to a user in the presigned URL endpoint for files in S3

### Improvements
- `fence-create client-create` now outputs the whole stacktrace in case of error

### Deployment changes
- Requires a Fence DB migration (`Client.redirect_uri` is now optional)


README.md
```diff
@@ -515,6 +515,14 @@ To specify allowed scopes, use the `allowed-scopes` argument:
 fence-create client-create ...  --allowed-scopes openid user data
 ```
 
+#### Register an Oauth Client for a Client Credentials flow
+
+As a Gen3 commons administrator, if you want to create an oauth client for a client credentials flow:
```
Avantol13, line 520: we may want more docs than this, or an explanation of when you might want this (and how authorization is applied the same as users and this only works against certain endpoints)

fence/blueprints/data/indexd.py
```diff
@@ -1537,31 +1550,40 @@ def _get_user_info_for_id_or_from_request(
...
     except Exception as exc:
         logger.info(
-            "could not determine user info from request. setting anonymous user information."
+            f"could not determine user auth info from request. setting anonymous user information. Details:\n{exc}"
+        )
+
+    client_id = ""
+    try:
+        if not token:
+            token = validate_request(scope=[], audience=config.get("BASE_URL"))
```
Avantol13, line 1566: we may want to consider a scope for client-level information similar to `user`, maybe `client`. I'm not sure when we wouldn't provide this though.. and if the spec doesn't dictate then maybe it's not necessary

fence/jwt/token.py
```diff
@@ -358,9 +360,9 @@ def generate_api_key(kid, private_key, user_id, expires_in, scopes, client_id):
 def generate_signed_access_token(
```
Avantol13, line 360: we're not actually allowing client credentials tokens to generate signed URLs, right? All these changes are just to ensure that a sane 401 is returned (instead of 500-ing off of all the user specific access stuff) and pave the road for potentially adding that support in the future?

fence/blueprints/data/indexd.py
```diff
@@ -643,7 +643,9 @@ def check_legacy_authorization(self, action):
             )
             return self.index_document.get("uploader") == username
 
-        given_acls = set(filter_auth_ids(action, flask.g.user.project_access))
+        given_acls = set()
```
Avantol13, line 646: Okay so my read of this is that we're not actually allowing client credentials tokens to generate signed URLs, right? All these changes are just to ensure that a sane 401 is returned (instead of 500-ing off of all the user specific access stuff) and pave the road for potentially adding that support in the future?

fence/models.py
```diff
@@ -260,6 +279,7 @@ def check_client_secret(self, client_secret):
 
     def check_requested_scopes(self, scopes):
         if "openid" not in scopes:
+            logger.error(f"Invalid scopes: 'openid' not in requested scopes ({scopes})")
```
Avantol13, line 282: This is one of those error lines that I imagine you added after hours of frustration trying to figure out why something wasn't working, lol. Or at least that's what it'd be for me

paulineribeyre: yes 😄

fence/blueprints/data/indexd.py
paulineribeyre: it does work (s3 only) if the client has read-storage access to the resource. we can block it if you would rather we do that though

fence/oidc/grants/client_credentials_grant.py
```diff
@@ -0,0 +1,7 @@
+from authlib.oauth2.rfc6749.grants import (
+    ClientCredentialsGrant as AuthlibClientCredentialsGrant,
+)
+
+
+class ClientCredentialsGrant(AuthlibClientCredentialsGrant):
+    pass
```
Avantol13, line 7: why the declaration and pass here? Why not just import from authlib?

fence/oidc/jwt_generator.py
```diff
@@ -220,3 +222,25 @@ def generate_token_response(
         "refresh_token": refresh_token,
         "expires_in": expires_in,
     }
+
+
+def generate_client_response(client, expires_in=None, scope=None, **kwargs):
+    keypair = flask.current_app.keypairs[0]
```
Avantol13, line 228: docstring please

fence/oidc/jwt_generator.py
```diff
@@ -220,3 +222,25 @@ def generate_token_response(
         "refresh_token": refresh_token,
         "expires_in": expires_in,
     }
+
+
+def generate_client_response(client, expires_in=None, scope=None, **kwargs):
```
Avantol13, line 227: the validation checks happen before this, right? This is never called from anywhere before validation of the token? I assume so (I can't see where it wouldn't be) but I just want to double check

fence/sync/sync_users.py
```diff
@@ -1481,6 +1481,7 @@ def _sync(self, sess):
                 self.sync_from_local_yaml_file, encrypted=False, logger=self.logger
             )
         except (EnvironmentError, AssertionError) as e:
+            # TODO return an error code so usersync doesn't fail silently
```
Avantol13, line 1484: does changing the `return` to `raise` break a bunch of automated tests? why not just do this now

fence/utils.py
```diff
@@ -1,3 +1,4 @@
+from cgitb import grey
```
Avantol13, line 1: what is this for? it doesn't look like it's used

fence/utils.py
```diff
@@ -51,7 +52,6 @@ def create_client(
     client_id = random_str(40)
     if arborist is not None:
         arborist.create_client(client_id, policies)
-    grant_types = grant_types
```
Avantol13, line 54: :D I want to `git blame` but I won't. mostly because I'm afraid it might have been me

fence/utils.py
```diff
@@ -61,22 +61,27 @@ def create_client(
             client_secret.encode("utf-8"), bcrypt.gensalt()
         ).decode("utf-8")
     auth_method = "client_secret_basic" if confidential else "none"
+
```
Avantol13, line 64: thank you for more whitespace for readability :tada:

migrations/versions/e4c7b0ab68d3_create_tables.py
```diff
@@ -381,7 +381,7 @@ def upgrade():
         "client",
         sa.Column("issued_at", sa.Integer(), nullable=False),
         sa.Column("expires_at", sa.Integer(), nullable=False),
-        sa.Column("redirect_uri", sa.Text(), nullable=True),
+        sa.Column("redirect_uri", sa.Text(), nullable=False),
```
Avantol13, line 384: wait, why are we changing existing revisions - isn't the idea that we always create a new revision to update/downgrade what we need to change?

tests/conftest.py
```diff
@@ -1376,11 +1390,48 @@ def oauth_client_public(app, db_session, oauth_user):
...
     )
     db_session.commit()
-    return Dict(client_id=client_id, url=url)
+    return Dict(client_id=client_id, url=url, grant_types=grant_types)
+
+
+@pytest.fixture(scope="function")
+def oauth_client_with_client_credentials(
+    db_session, get_all_shib_idps_patcher
+):  # app, db_session, oauth_user, get_all_shib_idps_patcher):
```
Avantol13, line 1403: dead comment, we can remove?

tests/data/test_data.py
```diff
@@ -1822,3 +1823,46 @@ def test_delete_files(app, client, auth_client, encoded_creds_jwt, user_client):
         assert status == 400
 
     fence.auth.config["MOCK_AUTH"] = False
+
+
+def test_download_s3_file_with_client_token(
```
Avantol13, line 1828: wait we are allowing data download from clients? Is this a driving requirement? I guess I'm maybe misremembering, but I had it in my head that we were actually NOT going to allow this to start with

Avantol13 (changes requested):
see comments

tests/data/test_data.py
paulineribeyre: it's not a driving requirement. without thinking that's the endpoint i used to test that things were working. since it was working i left it in

fence/blueprints/data/indexd.py
Avantol13: ehhh, I feel like we want to restrict client direct access to data until we absolutely need to support it. Least privilege and all that.

fence/blueprints/data/indexd.py
paulineribeyre: good point - since we're not allowing client tokens for these endpoints anyway, i'm leaving this as-is so we don't complicate things. let me know if you think we should add the scope.

fence/oidc/grants/client_credentials_grant.py
paulineribeyre: i assumed i'd need to overwrite something, but i didn't 😆 i'll update this

fence/sync/sync_users.py
paulineribeyre: i would rather not do that in this PR because i haven't tested it

migrations/versions/e4c7b0ab68d3_create_tables.py
paulineribeyre: you're right but i wasn't sure what to do

`redirect_uri` is supposed to be required at the moment. I think what happened is that i was working on the client-credentials feature before starting work on the alembic feature, so i had made this change to my local db and alembic picked it up when generating the initial migration 🤦‍♀️ 

the migration to make it nullable doesn't fail if it is already nullable, so i think i can revert this change 🤦‍♀️

fence/oidc/jwt_generator.py
paulineribeyre: i only see it called by authlib, which does validate the client ID and secret before generating the token:
```
Fence `POST /token`
=> authlib `AuthorizationServer.create_token_response`
   => first authlib `ClientCredentialsGrant.validate_token_request`
   => then authlib `ClientCredentialsGrant.create_token_response`
      => authlib `ClientCredentialsGrant.generate_token`
         => Fence `jwt_generator.generate_token`
            => Fence `jwt_generator.generate_client_response`
```

fence/jwt/token.py
```diff
@@ -421,13 +415,27 @@ def generate_signed_access_token(
     if scopes:
         claims["aud"] += scopes
 
-    # only add google linkage information if provided
-    if linked_google_email:
-        claims["context"]["user"]["google"][
-            "linked_google_account"
-        ] = linked_google_email
+    sub = None
```
UchicagoZchen138, line 418: why do we initalize sub here? Should we remove it?

fence/models.py
```diff
@@ -214,6 +214,25 @@ def __init__(self, client_id, **kwargs):
             # assume it's already in correct format
             kwargs["grant_type"] = grant_types
 
+        supported_grant_types = [
```
UchicagoZchen138, line 217: can we turn this into an enum?

Avantol13 (approved):

Avantol13 (approved):

fence/models.py
paulineribeyre: how would it be more flexible? unless it's really needed i prefer to go for the simpler solutions

PR #1037, opened 2022-08-26
fix(gs-cache): cleanup cache usage, don't try to save key_db_entry no…

Avantol13:
### Bug Fixes
- Fix intermittent issue with db cache containing info that can't be loaded into JSON

### Dependency updates
- Update poetry install in travis.yml due to https://python-poetry.org/blog/announcing-poetry-1.2.0a1/#deprecation-of-the-get-poetrypy-script

### DIFF
--- a/fence/blueprints/data/indexd.py
+++ b/fence/blueprints/data/indexd.py
@@ -1190,6 +1190,7 @@ class GoogleStorageIndexedFileLocation(IndexedFileLocation):
         proxy_group_id = get_or_create_proxy_group_id(
             user_id=user_id, username=username
         )
+        print("------------------0--------------------")
         expiration_time = int(time.time()) + expires_in
         is_cached = False
 
@@ -1198,6 +1199,9 @@ class GoogleStorageIndexedFileLocation(IndexedFileLocation):
                 raw_private_key,
                 expires_at,
             ) = self._assume_role_cache_gs.get(proxy_group_id, (None, None))
+            print("------------------1----------------")
+            print(self._assume_role_cache_gs)
+            print(raw_private_key)
 
             if expires_at and expires_at > expiration_time:
                 is_cached = True
@@ -1214,7 +1218,10 @@ class GoogleStorageIndexedFileLocation(IndexedFileLocation):
                     .first()
                 )
                 if cache and cache.expires_at > expiration_time:
+                    print("------------------2----------------")
+
                     private_key = json.loads(cache.gcp_private_key)
+                    print(private_key)
                     expires_at = cache.expires_at
                     self._assume_role_cache_gs[proxy_group_id] = (
                         private_key,
@@ -1228,6 +1235,8 @@ class GoogleStorageIndexedFileLocation(IndexedFileLocation):
             private_key, key_db_entry = get_or_create_primary_service_account_key(
                 user_id=user_id, username=username, proxy_group_id=proxy_group_id
             )
+            print("------------------3---------------")
+            print(private_key)
 
             # Make sure the service account key expiration is later
             # than the expiration for the signed url. If it's not, we need to
@@ -1243,6 +1252,8 @@ class GoogleStorageIndexedFileLocation(IndexedFileLocation):
                 private_key = create_primary_service_account_key(
                     user_id=user_id, username=username, proxy_group_id=proxy_group_id
                 )
+                print("------------------4----------------")
+                print(private_key)
             self._assume_role_cache_gs[proxy_group_id] = (
                 private_key,
                 key_db_entry.expires,
@@ -1250,6 +1261,9 @@ class GoogleStorageIndexedFileLocation(IndexedFileLocation):
 
             db_entry = {}
             db_entry["gcp_proxy_group_id"] = proxy_group_id
+            print("------------------5----------------")
+            print(private_key)
+            print(type(private_key))
             db_entry["gcp_private_key"] = json.dumps(private_key)
             db_entry["expires_at"] = key_db_entry.expires
 
@@ -1287,6 +1301,10 @@ class GoogleStorageIndexedFileLocation(IndexedFileLocation):
         # use configured project if it exists and no user project was given
         if config["BILLING_PROJECT_FOR_SIGNED_URLS"] and not r_pays_project:
             r_pays_project = config["BILLING_PROJECT_FOR_SIGNED_URLS"]
+
+        print("------------------6----------------")
+        print(private_key)
+        
         final_url = cirrus.google_cloud.utils.get_signed_url(
             resource_path,
             http_verb,
--- a/fence/blueprints/data/indexd.py
+++ b/fence/blueprints/data/indexd.py
@@ -1201,7 +1201,7 @@ class GoogleStorageIndexedFileLocation(IndexedFileLocation):
 
             if expires_at and expires_at > expiration_time:
                 is_cached = True
-                private_key = json.loads(raw_private_key.replace("'", '"'))
+                private_key = json.loads(str(raw_private_key).replace("'", '"'))
                 expires_at = expires_at
             else:
                 del self._assume_role_cache_gs[proxy_group_id]
--- a/fence/blueprints/data/indexd.py
+++ b/fence/blueprints/data/indexd.py
@@ -1214,7 +1214,7 @@ class GoogleStorageIndexedFileLocation(IndexedFileLocation):
                     .first()
                 )
                 if cache and cache.expires_at > expiration_time:
-                    private_key = json.loads(cache.gcp_private_key)
+                    private_key = json.loads(str(cache.gcp_private_key).replace("'", '"'))
                     expires_at = cache.expires_at
                     self._assume_role_cache_gs[proxy_group_id] = (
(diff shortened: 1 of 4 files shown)


fence/blueprints/data/indexd.py
```diff
@@ -1251,18 +1247,18 @@ def _generate_google_storage_signed_url(
...
                 private_key,
-                key_db_entry,
                 key_db_entry.expires,
             )
 
             db_entry = {}
             db_entry["gcp_proxy_group_id"] = proxy_group_id
-            db_entry["gcp_private_key"] = json.dumps(str(private_key))
-            db_entry["gcp_key_db_entry"] = str(key_db_entry)
+            db_entry["gcp_private_key"] = str(private_key)
```
Avantol13, line 1255: why don't we just json.dumps this and then json.loads should work without the need to do that string manipulation?

```
>>> import json
>>> test = {"foo": "bar"}
>>> db_entry = json.dumps(test)
>>> db_entry
'{"foo": "bar"}'
>>> json.loads(db_entry)
{'foo': 'bar'}
```

fence/blueprints/data/indexd.py
```diff
@@ -1214,23 +1214,19 @@ def _generate_google_storage_signed_url(
                     .first()
                 )
                 if cache and cache.expires_at > expiration_time:
-                    rv = (
-                        json.loads(cache.gcp_private_key),
-                        json.loads(cache.gcp_key_db_entry),
-                        cache.expires_at,
+                    private_key = json.loads(
+                        str(cache.gcp_private_key).replace("'", '"')
```
Avantol13, line 1218: I'm not sure I understand why this is necessary now. See next comment

tests/data/test_indexed_file.py
```diff
@@ -441,9 +441,7 @@ def test_internal_get_gs_signed_url_cache_new_key_if_old_key_expired(
             ):
                 indexed_file = IndexedFile(file_id="some id")
                 google_object = GoogleStorageIndexedFileLocation("gs://some/location")
-                keydbentry = UserGoogleAccountToProxyGroup()
-                keydbentry.expires = 10
-                google_object._assume_role_cache_gs = {"1": ("key", keydbentry, 10)}
+                google_object._assume_role_cache_gs = {"1": ("key", 10)}
```
Avantol13, line 444: we could probably add some more unit tests here to regression test this? What do you think , I'm not sure we have to have integration tests

fence/blueprints/data/indexd.py
BinamB: okay yes, done. Tested it out and its working

tests/data/test_indexed_file.py
```diff
@@ -461,6 +459,108 @@ def test_internal_get_gs_signed_url_cache_new_key_if_old_key_expired(
...
+            return_value=("sa_private_key"),
+        ):
+            with mock.patch.object(
+                cirrus.google_cloud.utils,
+                "get_signed_url",
+                return_value="https://cloud.google.com/compute/url",
+            ):
+                indexed_file = IndexedFile(file_id="some id")
+                google_object = GoogleStorageIndexedFileLocation("gs://some/location")
+                google_object._assume_role_cache_gs = {"1": ("key", 10)}
```
Avantol13, line 518: maybe we should use a dictionary as the key to better simulate the real world data (and hopefully catch any issues with the JSON parsing)

tests/data/test_indexed_file.py
```diff
@@ -461,6 +459,108 @@ def test_internal_get_gs_signed_url_cache_new_key_if_old_key_expired(
...
+                    expires_in=0,
+                    user_id=1,
+                    username="some user",
+                    r_pays_project=None,
+                )
+
+                after_cache = db_session.query(AssumeRoleCacheGCP).first()
+
+                assert after_cache
+                assert before_cache != after_cache
```
Avantol13, line 535: Why do we expect these to be different? Also I think we should test the actual contents of the cache, make sure it's a dictionary or JSON (whatever we actually expect in the code)

tests/data/test_indexed_file.py
```diff
@@ -461,6 +459,108 @@ def test_internal_get_gs_signed_url_cache_new_key_if_old_key_expired(
...
+    """
+    Test fence.blueprints.data.indexd.GoogleStorageIndexedFileLocation._generate_google_storage_signed_url does not use cached key if its expired
+
+    create presigned url
+        set cache in db
+    clear cache
+    create presigned url again
+        make sure cache is set correctly
+    """
+    # db_session.add(
```
Avantol13, line 480: should this be commented out?

tests/data/test_indexed_file.py
```diff
@@ -461,6 +465,115 @@ def test_internal_get_gs_signed_url_cache_new_key_if_old_key_expired(
...
+                    user_id=1,
+                    username="some user",
+                    r_pays_project=None,
+                )
+
+                after_cache = db_session.query(AssumeRoleCacheGCP).first()
+
+                assert after_cache
+                assert (
+                    str(type(after_cache))
```
Avantol13, line 537: I'm not sure this assert is worth having. The important thing is that we can json load properly, this type check seems unecessary

tests/data/test_indexed_file.py
```diff
@@ -461,6 +465,115 @@ def test_internal_get_gs_signed_url_cache_new_key_if_old_key_expired(
...
+                    == "<class 'fence.models.AssumeRoleCacheGCP'>"
+                )
+                # check if json loads can properly parse json string stored in cache
+                assert (
+                    str(type(json.loads(after_cache.gcp_private_key)))
```
Avantol13, line 542: I don't think we need this type check either. The assert below will make sure the json.loads works and creates a comparable object to the mock we provided. So I'd remove this line

tests/data/test_indexed_file.py
```diff
@@ -461,6 +465,115 @@ def test_internal_get_gs_signed_url_cache_new_key_if_old_key_expired(
...
+                    == "<class 'dict'>"
+                )
+                assert json.loads(after_cache.gcp_private_key) == sa_private_key
+
+                db_session.delete(after_cache)
+                cleared_cache = db_session.query(AssumeRoleCacheGCP).first()
```
Avantol13, line 548: This isn't the regression test we want, it's not bad to keep this too, but the test we need is:

* Generate url
* Simulate pod rolling (e.g. the **in memory** cache being cleared, **not** the database one)
* Generate url, ensure **new** in memory cache has valid info from the db

tests/data/test_indexed_file.py
```diff
@@ -461,6 +465,115 @@ def test_internal_get_gs_signed_url_cache_new_key_if_old_key_expired(
...
+                    user_id=1,
+                    username="some user",
+                    r_pays_project=None,
+                )
+
+                redo_cache = db_session.query(AssumeRoleCacheGCP).first()
+
+                assert redo_cache
+                assert (
+                    str(type(redo_cache)) == "<class 'fence.models.AssumeRoleCacheGCP'>"
```
Avantol13, line 565: see comments from above. I think we should remove all these type checks and only leave the json.loads assert. And we may want to do something like this to capture the failure correction:

```
try:
  assert json.loads(redo_cache.gcp_private_key) == sa_private_key
except Exception:
  pytest.fail("Could not json.loads(cache)")
```

tests/data/test_indexed_file.py
```diff
@@ -461,6 +465,107 @@ def test_internal_get_gs_signed_url_cache_new_key_if_old_key_expired(
...
+                google_object._generate_google_storage_signed_url(
+                    http_verb="GET",
+                    resource_path="gs://some/location",
+                    expires_in=0,
+                    user_id=1,
+                    username="some user",
+                    r_pays_project=None,
+                )
+
+                try:
```
Avantol13, line 536: you only need the try catch around the json.loads

Avantol13: so here you can just 
```
assert "1" in google_object._assume_role_cache_gs
assert len(google_object._assume_role_cache_gs["1"]) > 1
assert google_object._assume_role_cache_gs["1"][0] == sa_private_key
```

Avantol13: and remove the try/except

BinamB: Oh, saw these comments after i had added the try-catch for the json.loads. Should i still do this? My reasoning was, whatever is stored in the database should be json.loads-able.

tests/data/test_indexed_file.py
```diff
@@ -461,6 +465,107 @@ def test_internal_get_gs_signed_url_cache_new_key_if_old_key_expired(
...
+                google_object._generate_google_storage_signed_url(
+                    http_verb="GET",
+                    resource_path="gs://some/location",
+                    expires_in=0,
+                    user_id=1,
+                    username="some user",
+                    r_pays_project=None,
+                )
+
+                assert google_object._assume_role_cache_gs["1"][0] == sa_private_key
```
Avantol13, line 558:
```suggestion
```

this is already tested below

BinamB (approved):

PR #1042, opened 2022-09-06
Update `GET /data/upload/<GUID>` default filename

paulineribeyre:
The file gets uploaded to `s3://<bucket>/<GUID>/<file name>` so if we default to `file_name = GUID`, we end up with this folder structure: `s3://<bucket>/<GUID prefix>/<GUID ID>/<GUID prefix>/<GUID ID>`

### Improvements
- Update the default filename generated by the `GET /data/upload/<GUID>` endpoint
- Raise errors so usersync does not fail silently


MaribelleHGomez (dismissed):
:) lgtm

SpencerAxelrod (approved):

PR #1044, opened 2022-09-12
Update buildpipeline.yaml

Avantol13:
### Improvements
- temporarily do not block on security flags until we update to Python 3.9


cmlsn (approved):
Looks good

PR #1046, opened 2022-09-13
Create unitTests.yaml

AidanHilt:


.github/workflows/unitTests.yaml
```diff
@@ -0,0 +1,9 @@
+name: Run Fence unit tests
+on: push
+#comment
+jobs:
+  run_tests:
+    uses: uc-cdis/.github/.github/workflows/python_unit_test.yaml@3442040c24d683b335367e7fbbe3500ca8e07319
+    with:
+      setup-script-location: https://raw.githubusercontent.com/AidanHilt/.github/master/unit_test_scripts/fence.sh
```
Avantol13, line 8: is this from your personal .github repo? We can't use that here.

.github/workflows/unitTests.yaml
```diff
@@ -0,0 +1,9 @@
...
+      dependency-file-location: https://raw.githubusercontent.com/AidanHilt/.github/master/requirements/fence_requirements.txt
```
Avantol13, line 9: You should not use an external deps file like this, poetry install in the standard workflow

.travis.yml
```diff
@@ -24,6 +24,7 @@ before_script:
   - cd -
 
 script:
+  - pip list
```
Avantol13, line 27: this is redundant, it already happens above

tests/test-fence-config.yaml
```diff
@@ -411,6 +411,7 @@ dbGaP:
       port: 22
       proxy: ''
       proxy_user: ''
+      encrypted: false
```
Avantol13, line 414: Why change this?

.travis.yml
```diff
@@ -24,6 +24,7 @@ before_script:
   - cd -
 
 script:
```
paulineribeyre, line 26: we can delete this file

pyproject.toml
```diff
@@ -63,6 +63,90 @@ moto = "^1.1.24"
 pytest = "^3.2.3"
 pytest-cov = "^2.5.1"
 pytest-flask = "^0.15.0"
+Flask-Cors = "^3.0.10"
```
paulineribeyre, line 66: These new dependencies don't look right. Some of them are already in `[tool.poetry.dependencies]` so they don't need to be added to `[tool.poetry.dev-dependencies]`, and others seem to be sub-dependencies that should be installed automatically. What problems do you run into exactly when you keep the `pyproject.toml` file as-is?

.github/workflows/unitTests.yaml
```diff
@@ -0,0 +1,9 @@
+name: Run Fence unit tests
+on: push
+#comment
+jobs:
+  run_tests:
+    uses: uc-cdis/.github/.github/workflows/python_unit_test.yaml@826ccf7c3aebe705f22f2a64d36b689a5e7fc936
+    with:
+      setup-script-location: https://raw.githubusercontent.com/uc-cdis/fence/reusable-unit-test-workflow/tests/setup/fence.sh
+      python-versions: "['3.6.7', '3.9']"
```
paulineribeyre, line 9: i'm surprised the tests pass with python 3.9, i thought we were running into issues updating fence to 3.9 :-)

.github/workflows/unitTests.yaml
```diff
@@ -0,0 +1,9 @@
+name: Run Fence unit tests
+on: push
+#comment
+jobs:
+  run_tests:
+    uses: uc-cdis/.github/.github/workflows/python_unit_test.yaml@826ccf7c3aebe705f22f2a64d36b689a5e7fc936
+    with:
+      setup-script-location: https://raw.githubusercontent.com/uc-cdis/fence/reusable-unit-test-workflow/tests/setup/fence.sh
```
paulineribeyre, line 8: if we get a PR that updates this file, we need to make sure we are testing the changes. Could we `git checkout` the current branch and use the branch's `tests/setup/fence.sh` file, instead of the version of the file on master branch?

this should work, since fence is a public repo: https://github.com/uc-cdis/gen3sdk-python/blob/174988b/.github/workflows/docs.yml#L12-L15

tests/test-fence-config.yaml
paulineribeyre: i'm not sure why exactly it was causing issues, but the context is in this thread https://cdis.slack.com/archives/C0294A7KVE0/p1663789686063239

PR #1050, opened 2022-10-03
fix(refresh_tokens): ensure deletion of expired tokens from the database

Avantol13:
### Improvements
- ensure deletion of expired tokens from the database


fence/resources/openid/idp_oauth2.py
```diff
@@ -191,10 +191,17 @@ def get_access_token(self, user, token_endpoint, db_session=None):
             refresh_token = row.refresh_token
```
paulineribeyre, line 191: should we sort the tokens from oldest to newest to make sure the old ones get cleaned up?

paulineribeyre (approved):

fence/resources/openid/idp_oauth2.py
Avantol13: this doesn't break on the first one it finds, it goes through all of them. So either way they'll all get cleaned up

Avantol13: Ohh, I misunderstood what you were saying. Good catch, yeah I'll need to make another PR. Ideally there's only ever 1 expired at a time but this might be problematic for the current state

PR #1048, opened 2022-09-17
Adds bucket parameter to upload_file

bwalsh:
### New Features

* Adds  optional parameter `bucket`, to allow caller to specify destination bucket, a member of fence_config.S3_BUCKETS
* Defaults to DATA_UPLOAD_BUCKET

### Breaking Changes

`None`

### Bug Fixes

`None`

### Improvements

`None`

### Dependency updates

`None`

### Deployment changes

Paired with corresponding PR to gen3sdk https://github.com/uc-cdis/gen3sdk-python/pull/158

PR #1051, opened 2022-10-03
PXP-10361 Allow specifying the upload bucket

paulineribeyre:
Fixes and closes #1048

### New Features
- Allow specifying the bucket to upload to (endpoints `/data/upload` and `/data/upload/<GUID>`)

### Deployment changes
- Configure the `ALLOWED_DATA_UPLOAD_BUCKETS` setting to allow users to upload to buckets other than `DATA_UPLOAD_BUCKET`

### Improvements
- Return a 500 error instead of a 401 error when an S3 bucket is not configured properly

### DIFF
--- a/fence/blueprints/data/blueprint.py
+++ b/fence/blueprints/data/blueprint.py
@@ -306,11 +306,11 @@ def upload_file(file_id):
         s3_buckets = get_value(
             flask.current_app.config,
             "S3_BUCKETS",
-            InternalError("buckets not configured"),
+            InternalError("S3_BUCKETS not configured"),
         )
         if bucket not in s3_buckets:
             logger.debug(f"Bucket '{bucket}' not found in S3_BUCKETS config")
-            raise Unauthorized("permission denied for bucket")
+            raise InternalError("permission denied for bucket")
 
     result = get_signed_url_for_file(
         "upload", file_id, file_name=file_name, bucket=bucket
--- a/fence/blueprints/data/blueprint.py
+++ b/fence/blueprints/data/blueprint.py
@@ -1,6 +1,7 @@
 import flask
 
 from cdislogging import get_logger
+from cdispyutils.config import get_value
 
 from fence.auth import login_required, require_auth_header, current_token, get_jwt
 from fence.authz.auth import check_arborist_auth
@@ -10,7 +11,7 @@ from fence.blueprints.data.indexd import (
     get_signed_url_for_file,
 )
 from fence.config import config
-from fence.errors import Forbidden, InternalError, UserError, Forbidden
+from fence.errors import Forbidden, InternalError, UserError, Forbidden, Unauthorized
 from fence.resources.audit.utils import enable_audit_logging
 from fence.utils import get_valid_expiration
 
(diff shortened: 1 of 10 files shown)


fence/blueprints/data/blueprint.py
```diff
@@ -293,7 +301,20 @@ def upload_file(file_id):
         file_name = str(file_id).replace("/", "_")
         logger.warning(f"file_name not provided, using '{file_name}'")
 
-    result = get_signed_url_for_file("upload", file_id, file_name=file_name)
+    bucket = flask.request.args.get("bucket")
+    if bucket:
+        s3_buckets = get_value(
+            flask.current_app.config,
+            "S3_BUCKETS",
```
Avantol13, line 308: similar to the above, I think we should have another layer of security by explicitly requiring the identification of buckets that are allowed for upload

Avantol13 (changes requested):

fence/blueprints/data/blueprint.py
```diff
@@ -175,11 +176,15 @@ def upload_data_file():
     )
 
     protocol = params["protocol"] if "protocol" in params else None
+    bucket = params.get("bucket")
```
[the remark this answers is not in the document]
paulineribeyre: > We may want to also consider only allowing this new feature when providing authz, as that will be required for passports going forward (so we'll want to deprecate acl even more than we are today).

I'm not sure that's a good idea because the "data upload" flow does not use `authz` and is still supported. Plus some people in the community have written their own code to mimic that flow (such as when they're not using S3 buckets) and i think we should allow them to use this feature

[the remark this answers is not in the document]
Avantol13: that's fair, we'll still have the restriction of requiring authz for passports but I think that's fine and people will just have to adjust upload flows if they want to use that

fence/blueprints/data/indexd.py
```diff
@@ -901,7 +912,8 @@ def get_credential_to_access_bucket(
 
         bucket_cred = s3_buckets.get(bucket_name)
         if bucket_cred is None:
-            raise Unauthorized("permission denied for bucket")
+            logger.debug(f"Bucket '{bucket_name}' not found in S3_BUCKETS config")
```
Avantol13, line 915: we'll never get here for data upload buckets, right? Because now we're erroring on startup. I think we should try to keep the old behavior and defer the error. Because a lot of commons don't actually use the data upload feature but might have a bucket configured (and even though Fence defaults it to '' I think cloud automation might automatically add one)

tests/data/test_blank_index.py
```diff
@@ -130,46 +117,86 @@ def json(self):
...
+        )
+        if expect_success:
+            assert response.status_code == 201, response.json
+            assert "guid" in response.json
+            assert "url" in response.json
+            bucket_in_url = bucket if bucket else config["DATA_UPLOAD_BUCKET"]
+            assert bucket_in_url in response.json["url"]
+        else:
+            # "permission denied for bucket"
+            assert response.status_code == 500, response
```
Avantol13, line 191: if the bucket is not in allowed bucket configuration, I think this should really be a 403. The 500 for default bucket makes sense (b/c it should be configured to work) but for buckets that are in S3_BUCKETS but **not** in the allowed, that's not a server error, that's an intentional configuration to deny access, e.g. a 403

Avantol13 (changes requested):

fence/blueprints/data/indexd.py
paulineribeyre: i'll change this to a warning log

paulineribeyre: although i think it might be better to check the config at startup in general, and then assume the config is correct, but that's a discussion for later :-)

Avantol13 (approved):

PR #1056, opened 2022-11-09
(PXP-10172): Fix invalid sequence warning

BinamB:
### Bug Fixes
- Fixed warning ` DeprecationWarning: invalid escape sequence '\d' ` in usersync while trying to read whitelist filename patterns.


tests/dbgap_sync/test_user_sync.py
```diff
@@ -717,7 +717,7 @@ def mock_merge(dbgap_servers, sess):
 
     # this function will be called once for each sftp server
     # the test config file has 3 dbgap sftp servers
```
MaribelleHGomez, line 719: guessing this changed to 2?

paulineribeyre: why is this change in the tests needed? the code change shouldn't have any impact

fence/sync/sync_users.py
```diff
@@ -370,7 +369,7 @@ def _match_pattern(filepath, id_patterns, encrypted=True):
         Returns:
             bool: whether the pattern matches
         """
-        id_patterns.append("authentication_file_phs(\d{6}).(csv|txt)")
+        id_patterns.append("authentication_file_phs(\\\d{6}).(csv|txt)")
```
paulineribeyre, line 372: doesn't this mean that you expect an actual `\` in the file name?

i fixed the same warning in other repos last week, you can use a raw string to address this warning: `r"..."`

tests/dbgap_sync/test_user_sync.py
BinamB: Lol i dont remember making this change. Changing this back because its obviously breaking the unit tests

fence/sync/sync_users.py
BinamB: Yeah i thought that was the solution too but it doesn't work :/

PR #1057, opened 2022-11-11
PXP-10250 PXP-10268 OIDC clients expliration

paulineribeyre:
Jira Tickets: [PXP-10250](https://ctds-planx.atlassian.net/browse/PXP-10250) and [PXP-10268](https://ctds-planx.atlassian.net/browse/PXP-10268)

Goes with https://github.com/uc-cdis/cloud-automation/pull/2075

### New Features
- Add `--expires-in` parameter to the `fence-create` `client-create` and `client-modify` commands to specify the number of days in which in a client expires
- Add the `fence-create client-delete-expired` command to remove expired OIDC clients and optionally post warnings in Slack

### DIFF
--- a/fence/models.py
+++ b/fence/models.py
@@ -126,6 +126,40 @@ def get_project_to_authz_mapping(session):
     return output
 
 
+def get_client_expires_at(expires_in, grant_types):
+    """
+    Given an `expires_in` value (days from now), return an `expires_at` value (timestamp).
+
+    expires_in (int/float/str): days until this client expires
+    grant_types (str): list of the client's grants joined by "\n"
+    """
+    expires_at = None
+
+    if expires_in:
+        try:
+            expires_in = float(expires_in)
+            assert expires_in > 0
+        except (ValueError, AssertionError):
+            raise UserError(
+                f"Requested expiry must be a positive integer; instead got: {expires_in}"
+            )
+
+        # for backwards compatibility, 0 means no expiration
+        if expires_in != 0:
+            expires_in_secs = expires_in * 24 * 60 * 60  # days to seconds
+            expires_at = (
+                datetime.utcnow() + timedelta(seconds=expires_in_secs)
+            ).timestamp()
+
+    if "client_credentials" in grant_types.split("\n"):
+        if not expires_in or expires_in <= 0 or expires_in > 366:
+            logger.warning(
+                "Credentials with the 'client_credentials' grant which will be used externally are required to expire within 12 months. Use the `--expires-in` parameter to add an expiration."
+            )
+
+    return expires_at
+
+
 class ClientAuthType(Enum):
     """
     List the possible types of OAuth client authentication, which are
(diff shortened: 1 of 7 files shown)


tests/scripting/test_fence-create.py
```diff
@@ -210,6 +215,44 @@ def to_test():
...
+
+    def to_test():
+        saved_client = db_session.query(Client).filter_by(name=client_name).first()
+        assert saved_client.grant_types == grant_types
+        if not expires_in:
+            assert saved_client.expires_at == 0
+        else:
+            expected_expires_at = now + expires_in * 24 * 60 * 60
+            # allow up to 4 seconds variation to account for test execution
+            assert saved_client.expires_at <= expected_expires_at + 4000
```
Avantol13, line 236: isn't expires in seconds not ms? I think this should be +- 4 not `4000`

tests/scripting/test_fence-create.py
```diff
@@ -294,6 +337,93 @@ def test_client_delete_error(app, db_session, cloud_manager, test_user_a):
...
+    ]
+
+    # make sure the service account for the expired client are deleted
+    client_sa_after = (
+        db_session.query(GoogleServiceAccount)
+        .filter_by(client_id="test_client_id_0")
+        .all()
+    )
+    assert len(client_sa_after) == 0
+
```
Avantol13, line 425: you should clean up the non-expired clients here too (if that's not done elsewhere)

tests/scripting/test_fence-create.py
```diff
@@ -1535,3 +1665,43 @@ def test_modify_client_action_modify_append_url(db_session):
...
+    else:
+        modify_client_action(DB=db_session, client=client_name, expires_in=expires_in)
+
+        # make sure the expiration was updated if necessary
+        if not expires_in:
+            assert client.expires_at == original_expires_at
+        else:
+            expected_expires_at = now + expires_in * 24 * 60 * 60
+            # allow up to 4 seconds variation to account for test execution
+            assert client.expires_at <= expected_expires_at + 4000
```
Avantol13, line 1706: see above, I think this is 4000 seconds, not 4 seconds

fence/models.py
```diff
@@ -184,9 +221,11 @@ class Client(Base, OAuth2ClientMixin):
     _default_scopes = Column(Text)
     _scopes = ["compute", "storage", "user"]
 
+    expires_at = Column(Integer, nullable=False, default=0)
```
Avantol13, line 224: how are existing clients getting migrated to this since it's not nullable?

Avantol13 (changes requested):

fence/models.py
paulineribeyre: turns out the `client` table already has the `expires_at` column, it was added to match the authlib model. It's already non-nullable and defaults to 0 (so for backwards compatibility, 0=no expiration)

https://github.com/uc-cdis/fence/blob/61778929b84210eca1da1242fa6006c97b550185/bin/old_migration_script.py#L595

Avantol13: if it already has it, why do we need to add it here?

paulineribeyre: it's in the table but it's not declared in the model. looks like it works fine without declaring it, but isn't it better that it's explicit?

Avantol13: Yeah, it's better. I was just worried if it was necessary to add it here to make things work. But it sounds like if the table already has it, then this is more just for clarity in the future and everything should work fine without needing to run this migration?

paulineribeyre: yes it all works fine with no migration 👍

Avantol13: :+1:

tests/scripting/test_fence-create.py
paulineribeyre: this made me go through a rabbit hole - there was a ~3600s difference. i assumed it was in ms and added 4s

but i found out this only happened for the `expires_in=1000 days` case. Turns out getting a timestamp is not as easy as `now + <n_days> * 24 * 60 * 60`. On March 10 2024 there is an additional hour to take into account, I couldn't really figure out why (the time change is supposed to be on March 12). But `datetime.timestamp()` does the job so i just replaced the manual calculation 😅

tests/scripting/test_fence-create.py
paulineribeyre: the database is emptied after each test so it should be fine: https://github.com/uc-cdis/fence/blob/61778929b84210eca1da1242fa6006c97b550185/tests/conftest.py#L545

tests/scripting/test_fence-create.py
Avantol13: oh shoot, yeah I bet that was fun to look into :sweat_smile:  time is fun. I'm glad `datetime` just does everything for us :D

Avantol13 (approved):

PR #1058, opened 2022-11-15
Update idp_oauth2.py

Avantol13:
### Bug Fixes
- sort expired tokens so all expired ones get deleted and valid one gets used


paulineribeyre:
fixes #1050

paulineribeyre (approved):

PR #1060, opened 2022-11-29
feat(dbenvvar): db connection info from env

jawadqur:

### New Features
Ability to add DB connection string via environment variable


fence/config.py
```diff
@@ -49,6 +49,9 @@ def post_process(self):
         for default in defaults:
             self.force_default_if_none(default, default_cfg=default_config)
 
+        # allow setting DB connection string via env var
+        self["DB"] = os.environ.get("DB", self["DB"])
```
paulineribeyre, line 53: could we add an informative log here to say whether we are using the value from the environment variable or from the config

paulineribeyre (approved):

PR #1061, opened 2022-12-08
feat(cookies): ensure setting secure on cookies

Avantol13:
### Improvements
- ensure secure cookies are set based on configuration


cmlsn (approved):
Changes are needed to secure session cookies - looks like the builds and tests completed

PR #1067, opened 2023-01-06
(PXP-10410): Add Project ID format check in usersync

BinamB:
### Improvements
- Add project id format check in usersync


fence/sync/sync_users.py
```diff
@@ -514,6 +514,17 @@ def _parse_csv(self, file_dict, sess, dbgap_config={}, encrypted=True):
             self.logger.info(
                 f"using study to common exchange area mapping: {study_common_exchange_areas}"
             )
+
+        project_id_patterns = [r"phs(\d{6})"]
+        if dbgap_config.get("allow_non_dbGaP_whitelist", False):
```
paulineribeyre, line 519: could you add `allow_non_dbGaP_whitelist` to `config-default.yaml`

fence/sync/sync_users.py
```diff
@@ -548,6 +559,29 @@ def _parse_csv(self, file_dict, sess, dbgap_config={}, encrypted=True):
...
+                        self.logger.debug(
+                            "Checking pattern:{} with project_id:{}".format(
+                                pattern, dbgap_project
+                            )
+                        )
+                        if re.match(pattern, dbgap_project):
+                            skip = False
+                            break
+                        else:
+                            self.logger.warning(
```
paulineribeyre, line 574: i think this log should be in the next `if` statement instead?
```
if skip:
    [log here]
    continue
```
because if another pattern matches you would still be logging that the project is skipped

fence/config-default.yaml
```diff
@@ -551,6 +551,8 @@ dbGaP:
       # 'studyX': ['/orgA/', '/orgB/']
       # 'studyX.c2': ['/orgB/', '/orgC/']
       # 'studyZ': ['/orgD/']
+    # Allowed patterns for project_ids. The default value in usersync is 'phs(\d{6}) for dbgap projects'
```
paulineribeyre, line 554: this should be more clear about `phs(\d{6})` being the first value in the list regardless, not just a default value - for example if you want to add another pattern, you don't need to add `phs(\d{6})` to your config

paulineribeyre, line 554: it should also be specified that this only works if `allow_non_dbGaP_whitelist` is true. But why are we adding this limitation in the first place? why not allow new patterns in any case?

paulineribeyre: that's actually a bit confusing, maybe we should either set `allowed_project_id_patterns: ["phs(\d{6})"]` as default, or rename it `additional_allowed_project_id_patterns: []`

BinamB: Good question. I guess I was just over-engineering it lol. Should I remove that in this PR?

tests/dbgap_sync/test_user_sync.py
```diff
@@ -74,13 +74,11 @@ def test_sync_incorrect_user_yaml_file(syncer, monkeypatch, db_session):
     assert syncer.arborist_client.create_policy.not_called()
 
 
-@pytest.mark.parametrize("allow_non_dbgap_whitelist", [False, True])
 @pytest.mark.parametrize("syncer", ["google", "cleversafe"], indirect=True)
 @pytest.mark.parametrize("parse_consent_code_config", [False, True])
 def test_sync(
     syncer,
     db_session,
-    allow_non_dbgap_whitelist,
```
paulineribeyre, line 83: need to add the `allow_non_dbGaP_whitelist` feature back as discussed, and add `allow_non_dbGaP_whitelist` to `config-default.yaml`

frankliuao:
Same Jenkins problems? Need more reviews or it's considered done?

paulineribeyre (approved):

paulineribeyre:
ignoring failing "CentralizedAuth" test which is currently broken and failing "Security Pipeline / Poetry_Vulnerabilities" check which will be addressed in a future PR

PR #1083, opened 2023-03-30
(PXP-10737): Add bucket_info region endpoint

BinamB:

### New Features
- Add new endpoint, `/bucket_info/region`, that exposes bucket-region mapping from the fence-config


fence/blueprints/bucket_info.py
```diff
@@ -0,0 +1,26 @@
+import flask
+
+from fence.auth import require_auth_header
+from fence.config import config
+
+blueprint = flask.Blueprint("bucket_info", __name__)
+
+
+@blueprint.route(
+    "/region",
```
Avantol13, line 10: maybe the endpoint should just return all bucket info and not just filter out to the regions

tests/test_bucket_info.py
```diff
@@ -0,0 +1,15 @@
+import fence.blueprints.bucket_info
+
+
+def test_get_bucket_info(client):
+    res = client.get("/bucket_info/region")
```
Avantol13, line 5: docstring please, quickly describing what you're testing

Avantol13 (changes requested):

tests/test_bucket_info.py
```diff
@@ -0,0 +1,15 @@
...
+
+    assert res.status_code == 200
+
+    bucket_info = res.json
+    for cloud, buckets in bucket_info.items():
+        for bucket, info in buckets.items():
+            assert "cred" not in info
+
+    assert len(bucket_info["S3_BUCKETS"]) == 5
+    assert len(bucket_info["GS_BUCKETS"]) == 2
```
paulineribeyre, line 15: can you also check that the actual values match?

fence/blueprints/bucket_info.py
BinamB: We have stuff like `cred` for the bucket in our config https://github.com/uc-cdis/fence/blob/master/fence/config-default.yaml#L634. I dont see it on our public configurations so I havent really seen one in use in any environments so was just trying to be careful.
https://github.com/uc-cdis/fence/blob/master/fence/config-default.yaml#L634

Avantol13: Yeah I guess filtering out the cred is probably smart. But I'd maybe reverse the logic to not only include region, but instead, filter out cred and role-arn. Leaving `endpoint_url` is probably helpful. 

I guess either way making it so that the `endpoint_url` and `region` are there is good

paulineribeyre (dismissed):

frankliuao:
Wait until Alex's PR gets merged before we merge this one. Talked about at the DCF standup on 04.26

fence/blueprints/data/blueprint.py
```diff
@@ -336,3 +337,22 @@ def download_file(file_id):
...
         return flask.jsonify(result)
     return flask.redirect(result["url"])
+
+
+@blueprint.route(
+    "/buckets",
+    methods=["GET"],
+)
+def get_bucket_region_info():
+    S3_BUCKETS = config.get("S3_BUCKETS", {}).copy()
```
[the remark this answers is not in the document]
frankliuao: If I understand correctly the `del` was trying to remove/update the config items from the original config although the operation was done with a copied object? Since the `buckets[bucket] = info_copy` replaces the older object with the new one, I think `copy` should work too? Actually it makes sure `info_copy` is always synced with the original in case something gets updated during this process? But I think I see your point of directly working on the original. Maybe Binam was being very safe here.

tests/test_bucket_info.py
```diff
@@ -0,0 +1,37 @@
...
+            "bucket2": {
+                "region": "us-east-1",
+                "endpoint_url": "https://cleversafe.example.com/",
+            },
+            "bucket3": {"region": "us-east-1"},
+            "bucket4": {"region": "us-east-1"},
+            "bucket5": {"region": "us-east-1"},
+        },
+        "GS_BUCKETS": {
+            "bucket1": {"region": "us-east-1"},
```
frankliuao, line 19: Probably ain't a big deal but I think GS regions go like US-EAST1

BinamB: There's not a standard way of representing bucket region yet. Im probably going to keep it the same as aws for now just for consistency

Avantol13 (approved):

Avantol13 (dismissed):

MaribelleHGomez (approved):

PR #1088, opened 2023-04-17
Add ability to read fence config(s) from env vars

jawadqur:
### Improvements
- Add ability to read fence config(s) from env vars

### DIFF
--- a/fence/config.py
+++ b/fence/config.py
@@ -49,6 +49,15 @@ class FenceConfig(Config):
         for default in defaults:
             self.force_default_if_none(default, default_cfg=default_config)
 
+        # Read in all environment variables starting with FENCE_
+        for env_var, env_val in os.environ.items():
+            if env_var.startswith("FENCE_"):
+                config_key = env_var[6:]  # remove "FENCE_" prefix
+                self[config_key] = env_val
+                logger.info(
+                    f"Found environment variable '{env_var}': overriding '{config_key}' field from config file"
+                )
+
         # allow setting DB connection string via env var
         if os.environ.get("DB"):
             logger.info(


fence/config.py
```diff
@@ -49,6 +49,15 @@ def post_process(self):
         for default in defaults:
             self.force_default_if_none(default, default_cfg=default_config)
 
+        # Read in all environment variables starting with FENCE_
+        for env_var, env_val in os.environ.items():
```
paulineribeyre, line 53: do you think we can revert [this](https://github.com/uc-cdis/fence/blob/8ea18eca17975acb7b479e55e5a8f01cb886033a/fence/config.py#L61-L70) then and just use this logic for the DB url?

paulineribeyre (commented):
The logic lgtm, but tagging @Avantol13 in case he has any concerns about this (this is for Helm charts support)

fence/config.py
Avantol13: +1

Avantol13:
How are these getting to env vars for Helm Charts? 

We put a lot of work into the YAML-based configuration support for default values and allowing overrides.

It's not clear to me how this fully replaces that. How does this handle nested configurations like OIDC providers?

Avantol13 (changes requested):
see comment on overall pr

fence/config.py
jawadqur: > 
> How are these getting to env vars for Helm Charts?
> 
https://github.com/uc-cdis/gen3-helm/blob/master/helm/fence/values.yaml#L200-L255 

> We put a lot of work into the YAML-based configuration support for default values and allowing overrides.
> 
We need a way to be able to mount k8s secrets into specific fields. Env var seems to be the best option, I am all ears for other ideas. 

> It's not clear to me how this fully replaces that. How does this handle nested configurations like OIDC providers?
It does not replace anything, it's just a supplement, and it does not support nested configs as far as I can tell. 

This is to be used for DB passwords, and the INDEXD_PASSWORD as those are secrets created by other charts, and consumed by fence.

jawadqur: I have also removed the DB specific ones, and I am adding an extra env var  in fence helm chart to support the new and old format for backwards compatibility.

paulineribeyre (commented):
lgtm

PR #1091, opened 2023-04-23
Label PRs to run only relevant tests

haraprasadj:

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,2 +1,2 @@
-\@requires-fence:
+'@requires-fence':
   - '**/*'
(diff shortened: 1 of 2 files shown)


paulineribeyre:
looks like the tests did not run

haraprasadj:
> looks like the tests did not run

```
[Pipeline] // dir
[Pipeline] echo
Found changes in a file inside dot folder: .github/labeler.yml
[Pipeline] echo
Found changes in a file inside dot folder: .github/workflows/labeler.yml
[Pipeline] echo
doc-only changes, skipping tests...
```

paulineribeyre (dismissed):
As discussed: my concern is that we need we tag each test with the relevant services and it introduces human error. It’s easy to make a mistake and skip tests that we shouldn’t. For example it’s easy to forget that arborist is needed when we write a test for another service

PR #1092, opened 2023-04-26
allow setting INDEXD_PASSWORD via env var

jawadqur:
### Deployment changes
- Allow setting `INDEXD_PASSWORD` via environment variable.

### DIFF
--- a/fence/config.py
+++ b/fence/config.py
@@ -67,7 +67,7 @@ class FenceConfig(Config):
             )
             self["INDEXD_PASSWORD"] = os.environ["INDEXD_PASSWORD"]
         else:
-            logger.info(
+            logger.debug(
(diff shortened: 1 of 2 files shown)


fence/config.py
```diff
@@ -60,6 +60,17 @@ def post_process(self):
...
             )
 
+        # allow setting INDEXD_PASSWORD via env var
+        if os.environ.get("INDEXD_PASSWORD"):
+            logger.info(
+                "Found environment variable 'INDEXD_PASSWORD': overriding 'INDEXD_PASSWORD' field from config file"
+            )
+            self["INDEXD_PASSWORD"] = os.environ["INDEXD_PASSWORD"]
+        else:
+            logger.info(
```
Avantol13, line 70: you could make this a debug log, I think it's fine excluding this under normal circumstances because the default behavior is that it loads from the config file

Avantol13 (approved):

PR #1094, opened 2023-05-02
PXP-10920 Support deleting data without GUID prefix

paulineribeyre:
Jira Ticket: [PXP-10920](https://ctds-planx.atlassian.net/browse/PXP-10920)

If we hit the Fence `DELETE /data/<GUID>` endpoint with a GUID without prefix, the indexd record is fetched successfully (GUID resolution without prefix) and the file is deleted from the bucket, but fence fails to delete the indexd record itself (the deletion endpoint does not accept a GUID without prefix).

### Improvements
- The `DELETE /data/<GUID>` endpoint now supports resolving and deleting a GUID without its prefix

[PXP-10920]: https://ctds-planx.atlassian.net/browse/PXP-10920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/fence/sync/sync_users.py
+++ b/fence/sync/sync_users.py
@@ -1976,6 +1976,11 @@ class UserSyncer(object):
                         )
                     )
                     continue
+                self.logger.debug(
+                    "updating client `{}` (found {} client IDs)".format(
+                        client_name, len(clients)
+                    )
+                )
                 # there may be more than 1 client with this name if credentials are being rotated,
                 # so we grant access to each client ID
                 for client in clients:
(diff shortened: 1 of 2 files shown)


fence/sync/sync_users.py
```diff
@@ -1976,6 +1976,11 @@ def _update_authz_in_arborist(
                         )
                     )
                     continue
+                self.logger.debug(
```
k-burt-uch, line 1979: What's this? Misc change in the branch or is there a relevance I'm not seeing?

k-burt-uch (approved):

fence/sync/sync_users.py
paulineribeyre: misc change from a previous debugging session 🙂

PR #1095, opened 2023-05-04
feat(backoff): Add backoff to give_service_account_billing_access

k-burt-uch:
### Improvements
- Adds backoff for the give_service_account_billing_access calls

### DIFF
--- a/tests/google/test_utils.py
+++ b/tests/google/test_utils.py
@@ -1,0 +1,34 @@
+from unittest.mock import patch, MagicMock
+
+import pytest
+
+from fence.resources.google.utils import (
+    give_service_account_billing_access_if_necessary,
+    GoogleCloudManager,
+)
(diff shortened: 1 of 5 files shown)


tests/google/test_utils.py
```diff
@@ -0,0 +1,36 @@
...
+            Exception("Something's wrong")
+        )
+
+        with pytest.raises(Exception):
+            give_service_account_billing_access_if_necessary(
+                sa_private_key, r_pays_project
+            )
+
+        assert (
+            mock_enter.return_value.give_service_account_billing_access.call_count == 3
```
Avantol13, line 35: maybe import the default settings and use that value instead of hard-coded 3

Avantol13 (dismissed):

paulineribeyre (approved):

PR #1098, opened 2023-05-17
Handle empty fields in usersync _parse_csv()

paulineribeyre:
```
Traceback (most recent call last):
[...]
File "/fence/fence/sync/sync_users.py", line 1288, in _get_user_permissions_from_csv_list
 user_projects, user_info = self._parse_csv(
  File "/fence/fence/sync/sync_users.py", line 559, in _parse_csv
   phsid = row.get("phsid", "").split(".")
AttributeError: 'NoneType' object has no attribute 'split'
```

This appear to happens with a malformed CSV row missing some commas, for example:
```
user name,login,phsid,permission set,created,updated
user1,user1
```

### Improvements
- More robust CSV parsing in `usersync`


BinamB (approved):
Lgtm

PR #1102, opened 2023-06-01
feat(PXP-10776): Cascading Authorization

k-burt-uch:
Usersync updated to grant access to child studies if a user is granted access to a parent study.

Parent/Child study relationships are defined by
dbGaP.parent_to_child_studies_mapping property in the fence-config.yaml.

Jira Ticket: [PXP-10776](https://ctds-planx.atlassian.net/browse/PXP-10776)

### New Features
- Usersync now supports parent/child study authorizations from dbGaP, granting user access to child studies when they're given access to parent studies. fence-config.yaml should have its dbGaP config updated to include the `parent_to_child_studies_mapping` property to define known parent to child relationships in dbGaP.

### Dependency updates
- Cryptography minimum version changed to 41.0.2 from 3.5. Support for OpenSSL <1.1.1d has been removed, support for LibreSSL <3.6 dropped.  See https://cryptography.io/en/latest/changelog/#v41-0-0 for more information. 

### Deployment changes
Update fence-config.yaml to include `parent_to_child_studies_mapping` to define parent/child relationships within the dbGaP block of the config.  

[PXP-10776]: https://ctds-planx.atlassian.net/browse/PXP-10776?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fence/sync/sync_users.py
```diff
@@ -650,6 +662,41 @@ def _parse_csv(self, file_dict, sess, dbgap_config={}, encrypted=True):
...
+        parent_phsid = dbgap_project.split(".")[0]
+        return parent_phsid in self.parent_to_child_studies_mapping
+
+    def _add_children_for_dbgap_project(
+        self, dbgap_project, privileges, username, sess, user_projects, dbgap_config
+    ):
+        parent_phsid = dbgap_project
+        child_suffix = ""
+        if self.parse_consent_code and re.match(
+            config["DBGAP_ACCESSION_WITH_CONSENT_REGEX"], dbgap_project
```
k-burt-uch, line 675: I'm not sure how to handle instances where we have non-dbGaP authorization providers. If we're not getting a phsId, I just ignore it here. Given consent codes parsing is dbGaP specific, I think this is fine.

fence/sync/sync_users.py
```diff
@@ -335,6 +335,9 @@ def __init__(
         self.is_sync_from_dbgap_server = is_sync_from_dbgap_server
         self.dbGaP = dbGaP
         self.parse_consent_code = dbGaP[0].get("parse_consent_code", True)
+        self.parent_to_child_studies_mapping = dbGaP[0].get(
```
Avantol13, line 338: I don't think any of this works when you have multiple SFTPs configured and the first one is a non-dbgap SFTP.

Nor does this support multiple dbgap SFTPs (that's maybe fine).

I guess I didn't realize we were hard-coding the first entry as the only dbGaP one. 

I think `parse_consent_code` and `parent_to_child_studies_mapping` should be accessed per item in the dbGaP list. So wherever these are used needs that config passed in (I think some places already do). But neither of these should be global across SFTPs imo, they are only within that given dbGaP configuration block.

I realize you didn't add this original code, so open to flagging this as tech debt and not fixing now, but if we do that, then we definitely need to update the default config to make it VERY clear that the first entry in this dbGaP list MUST BE a true dbGaP SFTP. AND that the parse_consent_code and this new mapping is inherited by every other non-dbgap SFTP configured.

fence/sync/sync_users.py
```diff
@@ -650,6 +662,41 @@ def _parse_csv(self, file_dict, sess, dbgap_config={}, encrypted=True):
 
         return user_projects, user_info
 
+    def _has_children(self, dbgap_project):
+        parent_phsid = dbgap_project.split(".")[0]
+        return parent_phsid in self.parent_to_child_studies_mapping
+
+    def _add_children_for_dbgap_project(
```
Avantol13, line 669: I appreciate the fact this is an internal function, but I think it's complex enough to warrant a docstring for future readability/comprehension

fence/sync/sync_users.py
Avantol13: yeah, this is one of the oddities of the way this is currently set up (see my previous comment about hard-coding the first entry of dbGaP config). 

I think we should support doing this mapping even if we're not parsing consent codes. That allows flexibility to non-dbgap SFTPs to use this generically for cascading access. 

I also think we should heavily prioritize the common case of **not** having child studies and returning early on that before much more work. I think you basically already do this with the early return.

See below for some rough pseudo code, this might be overkill but hopefully gets the idea across:

```python
    is_cascade = False
    cascade_parent = None
    
    if self._has_children(dbgap_project):
        is_cascade = True
        cascade_parent = dbgap_project
      
    # I'm honestly not sure if re.match or a simple split is faster, I assume split is... so I think for 
    # the same of speed doing this check is probably okay - even though I like the regex for readability
    # (never thought I'd utter those words)
    if dbgap_project.split(".") > 1:
        # get parts with regex
        phsid_consent = ".".join([phsid_part, consent_part])
        if phsid_consent in parent_to_child_studies_mapping:
            is_cascade = True
            cascade_parent = phsid_consent
        
   if not is_cascade:
       return
       
  children =  parent_to_child_studies_mapping.get("cascade_parent", [])
  # do cascade

```

fence/sync/sync_users.py
```diff
@@ -650,6 +662,41 @@ def _parse_csv(self, file_dict, sess, dbgap_config={}, encrypted=True):
...
+        self, dbgap_project, privileges, username, sess, user_projects, dbgap_config
+    ):
+        parent_phsid = dbgap_project
+        child_suffix = ""
+        if self.parse_consent_code and re.match(
+            config["DBGAP_ACCESSION_WITH_CONSENT_REGEX"], dbgap_project
+        ):
+            phsid_parts = dbgap_project.split(".")
+            parent_phsid = phsid_parts[0]
+            child_suffix = "." + phsid_parts[1]  # include parent consent code
```
Avantol13, line 679: not a huge deal at all, and I guess this isn't part of Black formatting, but we usually only use inline comments when absolutely necessary and tend to prefer them above the line. Or at least I do. Though if Black didn't complain then it's fine

fence/sync/sync_users.py
Avantol13: I actually did a little poking and can confirm split is faster than re.match

Just given that this is run every hour on every dataset, and given that I only expect to have to re.match on a very small subset of the studies, I think the optimization is worth it.

```
>>> timeit.timeit('re.match("(?P<phsid>phs[0-9]+)(.(?P<version>v[0-9]+)){0,1}(.(?P<participant_set>p[0-9]+)){0,1}.(?P<consent>c[0-9]+)", "phs0000123.p1.v2.c1")', number=1000000, setup='import re')
0.9019025509987841
>>> timeit.timeit('"phs0000123.p1.v2.c1".split(".")', number=1000000)
0.11285156600206392
>>> timeit.timeit('re.match("(?P<phsid>phs[0-9]+)(.(?P<version>v[0-9]+)){0,1}(.(?P<participant_set>p[0-9]+)){0,1}.(?P<consent>c[0-9]+)", "phs0000123.p1.v2.c1")', number=1000000, setup='import re')
0.9104178220004542
>>> timeit.timeit('"phs0000123.p1.v2.c1".split(".")', number=1000000)
0.10198385500189033
```

fence/sync/sync_users.py
```diff
@@ -1625,6 +1672,16 @@ def _grant_all_consents_to_c999_users(
                 consent_mapping.setdefault(accession_number["phsid"], set()).add(
                     ".".join([accession_number["phsid"], accession_number["consent"]])
                 )
+                if self._has_children(accession_number["phsid"]):
```
Avantol13, line 1675: minor optimization: if you change this to _get_children and use a local var to store result, you can do a truthiness check and avoid accessing the parent to child dict again

something like

```python
children = self._get_children(accession_number["phsid"])
if children:
  for child in children:
    # do stuff
```

fence/sync/sync_users.py
```diff
@@ -1625,6 +1672,16 @@ def _grant_all_consents_to_c999_users(
...
+                    child_phs_list = self.parent_to_child_studies_mapping[
+                        accession_number["phsid"]
+                    ]
+                    for child_phs in child_phs_list:
+                        consent_mapping.setdefault(child_phs, set()).add(
```
Avantol13, line 1680: note that this suffers from the same limitation as the general c999 handling. If no one has direct access to a given consent, this logic won't know about it, and won't grant access to it (since this is just appending all the project consents it actually sees in dbgap/user.yaml). 

So if, for example, a study has .c1 and .c2 and no one is given direct access to .c2, but someone **does** get access to .c999, this logic will only grant access to .c1 and .c999.

Avantol13 (changes requested):

fence/sync/sync_users.py
k-burt-uch: Updated sync_users to check parse_consent_code from a dbgap_config rather than a constructor arg. 

I also merged the parent_to_child_studies_mappings and validate that no duplicate parent study IDs exist across multiple dbGaP configs.

fence/sync/sync_users.py
k-burt-uch: Done.

fence/sync/sync_users.py
k-burt-uch: Just to clarify:
1. we're doing this logic even if parse_consent_code is false. If parse_consent_code is true, we do the regex match and pull out the parent phsId and consent code if it matches the dbgap accession regex.
2. we early return if the study has no children, regardless of parse_consent_code.

If I missed something else in your pseudo code, lmk and I'll address it.

fence/sync/sync_users.py
k-burt-uch: Done.

fence/sync/sync_users.py
k-burt-uch: Yeah, I'm not sure how to handle this edge case.

fence/sync/sync_users.py
k-burt-uch: Changed the variable name of phsid_parts to parent_phsid_parts so its clearer and removed the comment.

fence/sync/sync_users.py
Avantol13: Not much we can do, I wanted to make the note on the PR for future reference. It might be worth putting this in a comment in the actual code

fence/config.py
```diff
@@ -130,18 +130,35 @@ def post_process(self):
                     "BILLING_PROJECT_FOR_SA_CREDS or BILLING_PROJECT_FOR_SIGNED_URLS is set to a non-None value. "
                     "SESSION_ALLOWED_SCOPES includes `google_credentials`. Removing "
                     "`google_credentials` from USER_ALLOWED_SCOPES as this could allow "
-                    "end-users to indescriminently bill our default project. Clients are inheritently "
+                    "end-users to indiscriminately bill our default project. Clients are inherently "
```
Avantol13, line 133: thanks, I need spell check enabled on my editor

fence/sync/sync_users.py
```diff
@@ -455,6 +458,11 @@ def _get_from_ftp_with_proxy(self, server, path):
         )
         os.system(execstr)
 
+    def _get_parse_consent_code(self, dbgap_config={}):
+        return dbgap_config.get(
+            "parse_consent_code", True
+        )  # Should this really be true?
```
Avantol13, line 464: yes, parsing consent should be on by default imo

Avantol13 (dismissed):

Avantol13 (approved):
new unit test could use a docstring, other than that lgtm

PR #1104, opened 2023-06-08
Chore/enable client creds on downloads

george42-ctds:
Jira Ticket: [HP-1133](https://ctds-planx.atlassian.net/browse/HP-1133)

### Improvements

* Enable client-credentials on /download/{guid} and /ga4gh/drs/v1/objects/{guid}/access/{access_id} endpoints 

### Deployment changes

[HP-1133]: https://ctds-planx.atlassian.net/browse/HP-1133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/data/test_data.py
```diff
@@ -1838,8 +1838,8 @@ def test_download_s3_file_with_client_token(
 ):
     """
     Test that an access token that does not include a `sub` or `context.user.
-    name` (such as a token issued from the `client_credentials` flow) cannot be
-    used to download data from S3.
+    name` (such as a token issued from the `client_credentials` flow) can be
+    used to download data from S3 if the indexd_record has an `authz` field.
```
paulineribeyre, line 1842:
```suggestion
    used to download data from S3 if the indexd_record has an `authz` field, and that the `client_id` is used to sign.
```

paulineribeyre: ^ copying the original comment from https://github.com/uc-cdis/fence/pull/1033/commits/108fa7535364e6ba01299f80fa72f5fff60052bd#diff-026339dfebdb32c48aea006a3f5317b4b67caf2ef87cc7a8127ea811e71af697R1862-R1869

Avantol13 (commented):
Do we want this behind a configuration flag? It's possible some commons may want to allow clients to do this and others only want end-users to access data in this way.

paulineribeyre (commented):
lgtm, and we can add a config flag if necessary

george42-ctds:
Optional config flag CLIENT_CREDENTIALS_ON_DOWNLOAD_ENABLED has been added. Default is `false`. 

Unit tests now test flag=true and flag=false.

fence/config-default.yaml
```diff
@@ -971,3 +971,9 @@ USERSYNC:
   visa_types:
     ras: ['https://ras.nih.gov/visas/v1', 'https://ras.nih.gov/visas/v1.1']
 RAS_USERINFO_ENDPOINT: '/openid/connect/v1.1/userinfo'
+
+# //////////////////////////////////////////////////////////////////////////////////////
+# CLIENT CREDENTIALS
+# //////////////////////////////////////////////////////////////////////////////////////
+# set to true to enable client credentials on download/{guid}
```
Avantol13, line 978: Can you update this comment and add a test for the ga4gh endpoint too? I believe it should automatically work with this b/c it just calls these functions. But we should explicitly document that and have a test for it https://github.com/uc-cdis/fence/blob/3cc55b36c5dbedfb4c96b69016625a3b4fbaff6d/fence/blueprints/ga4gh.py#L1-L44

Avantol13 (changes requested):

fence/config-default.yaml
george42-ctds: Oh, yes, correct, this also now enables the ga4gh endpoint. Test has been added.

Avantol13 (approved):

PR #1106, opened 2023-06-28
feat(DCF-1446): Assign MFA policy on login

k-burt-uch:
Adds the mfa_policy on login when the configured claim and value are present in the user's id_token.

### New Features
- Enables support for identifying when users have authenticated with MFA.  

### Deployment changes
- Deployment changes to enable this feature for commons are described here https://github.com/uc-cdis/fence/blob/e36f6574ade17bb8c9c0c9df8801f9a59de423b1/docs/fence_multifactor_authentication_guide.md

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,12 @@ services:
 addons:
   postgresql: "13"
   apt:
-      packages:
-        - postgresql-13
+    sources:
+      - sourceline: deb http://apt.postgresql.org/pub/repos/apt/ jammy-pgdg main
+          13
+        key_url: https://www.postgresql.org/media/keys/ACCC4CF8.asc
+    packages:
+      - postgresql-13
 install:
   - pip install --upgrade pip
   - pip install poetry
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ sudo: false
 
 cache: pip
 
-addons:
-  postgresql: "9.6"
+services:
+    - postgresql
 
 install:
   - pip install --upgrade pip
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-
+dist: jammy
 python:
   - "3.9"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,11 @@ cache: pip
 services:
     - postgresql
 
+addons:
+  postgresql: "13"
+  apt:
+      packages:
+        - postgresql-13
 install:
   - pip install --upgrade pip
   - pip install poetry
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,15 @@ addons:
         key_url: https://www.postgresql.org/media/keys/ACCC4CF8.asc
     packages:
       - postgresql-13
+
+before_install:
+  # Copy custom configs from the repo because PG-13 isn't set up to run like
+  # it normally does on Travis out of the box.
+  # Source: https://github.com/NCI-GDC/psqlgraph/blob/94f315db2c039217752cba85d9c63988f2059317/.travis.yml
+  - sudo cp travis/postgresql.conf /etc/postgresql/13/main/postgresql.conf
+  - sudo cp travis/pg_hba.conf /etc/postgresql/13/main/pg_hba.conf
+  - sudo pg_ctlcluster 13 main restart
+
 install:
   - pip install --upgrade pip
   - pip install poetry
--- a/docs/fence_multifactor_authentication_guide.md
+++ b/docs/fence_multifactor_authentication_guide.md
@@ -3,7 +3,7 @@
 Fence is capable of using token claims from IdPs to identify when multifactor authentication (MFA) was used during the authentication process.
 
 ## File Level Enforcement
-To restrict access to files to user who've authenticated with MFA, the following resource *MUST* be present for indexd record:
+To restrict access to files to user who've authenticated with MFA, the following resource *MUST* be present in the indexd record's `authz`:
 `/multifactor_auth`
 
 And the following configs must be updated:
--- a/docs/fence_multifactor_authentication_guide.md
+++ b/docs/fence_multifactor_authentication_guide.md
@@ -1,0 +1,64 @@
+# Fence Multifactor Authentication Guide
+
+Fence is capable of using token claims from IdPs to identify when multifactor authentication (MFA) was used during the authentication process.
+
+## File Level Enforcement
+To restrict access to files to user who've authenticated with MFA, the following resource *MUST* be present for indexd record:
+`/multifactor_auth`
+
+And the following configs must be updated:
+- fence-config.yaml
+- user.yaml
+
+### fence-config.yaml changes
+
+MFA claim checking is configured on a per-IdP basis. For a given IdP, define the name of the claim in the id_token and is possible values that indicate MFA. If the id_token claim value matches at least one value in the configured multifactor_auth_claim_info.values, then "/multifactor_auth" resource will be assigned to the user.
+
+For example, Okta may issue the following id_token when MFA is used:
+```
+{
+  "amr": ["otp", "pwd"],
+  "aud": "6joRGIzNCaJfdCPzRjlh",
+  "auth_time": 1311280970,
+  "exp": 1311280970,
+  "iat": 1311280970,
+  "idp": "00ok1u7AsAkrwdZL3z0g3",
+  "iss": "https://$"
+  "jti": "Tlenfse93dgkaksginv",
+  "sub": "00uk1u7AsAk6dZL3z0g3",
+  "ver": 1
+}
+```
+
+And fence-config.yaml is configured as follows:
+```
+OPENID_CONNECT:
+  okta:
+    client_id: 'redacted'
+    client_secret: 'redacted'
+    multifactor_auth_claim_info:
+      claim: 'amr'
+      values: [  "mfa", "otp", "sms" ]
+```
+
+Then fence will assign the "/multifactor_auth" resource to the user in Arborist.
+
+### user.yaml changes
+The `mfa_policy` policy and `multifactor_auth` resource must be added to user.yaml so the appropriate policy and resource are created in arborist when usersync runs.
+
+NOTE: The role_ids provided here are an example and should be changed to the appropriate arborist roles for the commons.
+
+Add the following to the `resources` section:
+```yaml
+  - name: multifactor_auth
+```
+
+Add the following the `policies` section:
+```yaml
+- id: mfa_policy
+  role_ids:
+   - read-storage
+   - read
+   resource_paths:
+   - /multifactor_auth
+```
--- a/fence/blueprints/login/base.py
+++ b/fence/blueprints/login/base.py
@@ -1,4 +1,5 @@
 import flask
+from cdislogging import get_logger
 from flask_restful import Resource
 from urllib.parse import urlparse, urlencode, parse_qsl
 
@@ -6,6 +7,8 @@ from fence.auth import login_user
 from fence.blueprints.login.redirect import validate_redirect
 from fence.config import config
 from fence.errors import UserError
+
+logger = get_logger(__name__)
 
 
 class DefaultOAuth2Login(Resource):
@@ -63,6 +66,7 @@ class DefaultOAuth2Callback(Resource):
         username_field="email",
         email_field="email",
         id_from_idp_field="sub",
+        app=flask.current_app,
     ):
         """
         Construct a resource for a login callback endpoint
@@ -84,6 +88,10 @@ class DefaultOAuth2Callback(Resource):
         self.username_field = username_field
         self.email_field = email_field
         self.id_from_idp_field = id_from_idp_field
+        self.is_mfa_enabled = "multifactor_auth_claim_info" in config[
+            "OPENID_CONNECT"
+        ].get(self.idp_name, {})
+        self.app = app
 
     def get(self):
         # Check if user granted access
@@ -109,7 +117,7 @@ class DefaultOAuth2Callback(Resource):
             return flask.redirect(location=final_redirect_url)
 
         code = flask.request.args.get("code")
-        result = self.client.get_user_id(code)
+        result = self.client.get_auth_info(code)
         username = result.get(self.username_field)
         if not username:
             raise UserError(
@@ -125,6 +133,22 @@ class DefaultOAuth2Callback(Resource):
 
     def post_login(self, user=None, token_result=None, **kwargs):
         prepare_login_log(self.idp_name)
+        if token_result:
+            username = token_result.get(self.username_field)
+            if self.is_mfa_enabled:
+                if token_result.get("mfa"):
+                    logger.info(f"Adding mfa_policy for {username}")
+                    self.app.arborist.grant_user_policy(
+                        username=username,
+                        policy_id="mfa_policy",
+                    )
+                    return
+                else:
+                    logger.info(f"Revoking mfa_policy for {username}")
+                    self.app.arborist.revoke_user_policy(
+                        username=username,
+                        policy_id="mfa_policy",
+                    )
 
 
 def prepare_login_log(idp_name):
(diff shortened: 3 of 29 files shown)


fence/blueprints/login/ras.py
```diff
@@ -114,4 +114,4 @@ def post_login(self, user=None, token_result=None, id_from_idp=None):
             user=user, refresh_token=refresh_token, expires=expires + issued_time
         )
 
-        super(RASCallback, self).post_login(id_from_idp=id_from_idp)
+        super(RASCallback, self).post_login(token_result=token_result)
```
k-burt-uch, line 117: This says token_result but the return value from get_user_id isn't the id_token, its just a dict containing the username or email or both depending on the IdP used. I tagged the mfa property into that dict.

fence/resources/openid/cilogon_oauth2.py
```diff
@@ -42,7 +42,7 @@ def get_user_id(self, code):
             claims = self.get_jwt_claims_identity(token_endpoint, jwks_endpoint, code)
 
             if claims.get("sub"):
-                return {"sub": claims["sub"]}
+                return {"sub": claims["sub"], "mfa": self.has_mfa_claim(claims)}
```
k-burt-uch, line 45: This is the weird part I mentioned in our 1-1 @Avantol13. This has nothing to do with getting the user's id like the method name implies but it is the only place where the idp_oauth2 implementations reach out and grab tokens.

Should I split up this method into two parts: get_tokens and get_user_id? My thinking is since this is used in base.py, it'd be a simple refactor to call get_tokens with the authorization code, return the tokens from there, then call get_user_id with the id_token instead of the authorization_code.

fence/blueprints/login/base.py
```diff
@@ -63,6 +66,7 @@ def __init__(
         username_field="email",
         email_field="email",
         id_from_idp_field="sub",
+        app=flask.current_app,
```
Avantol13, line 69: is there a reason for this vs `from flask import current_app` at the import level of the module?

fence/resources/openid/idp_oauth2.py
```diff
@@ -220,6 +223,22 @@ def get_access_token(self, user, token_endpoint, db_session=None):
...
+        """
+        Determines if the claim denoting whether multifactor authentication was used is contained within the claims
+        of the provided id_token.
+
+        Parameters:
+        - decoded_id_token (dict): The decoded id_token, a dict of claims -> claim values.
+
+        """
+        mfa_claim_info = self.settings.get("multifactor_auth_claim_info")
+        if mfa_claim_info is None:
```
Avantol13, line 236: you could be "truthy" here instead of explicitly checking None. That would early return on an empty dict too.

`if not mfa_claim_info: return False`

fence/resources/openid/idp_oauth2.py
```diff
@@ -220,6 +223,22 @@ def get_access_token(self, user, token_endpoint, db_session=None):
...
+            return False
+        claim_name = mfa_claim_info["claim"]
```
Avantol13, line 238: does the config get validated before this? Or could this fail if someone mistypes the config? 

I'd either: validate before, or be safely trying to access and logging if you can't find something here

tests/login/test_base.py
```diff
@@ -0,0 +1,48 @@
+from fence.blueprints.login import DefaultOAuth2Callback
+from fence.config import config
+from unittest.mock import MagicMock, patch
+
+
+@patch("fence.blueprints.login.base.prepare_login_log")
+def test_post_login_set_mfa(app, monkeypatch):
+    monkeypatch.setitem(
```
Avantol13, line 8: quick dosctring

tests/login/test_base.py
```diff
@@ -0,0 +1,48 @@
...
+    token_result = {"username": "homersimpson", "mfa": False}
+    callback.post_login(token_result=token_result)
+    app.arborist.revoke_user_policy.assert_called_with(
+        username=token_result["username"], policy_id="mfa_policy"
+    )
+
+
+@patch("fence.blueprints.login.base.prepare_login_log")
+def test_post_login_no_mfa_enabled(app, monkeypatch):
+    app.arborist = MagicMock()
```
Avantol13, line 38: quick docstring

fence/resources/openid/cilogon_oauth2.py
Avantol13: Yeah it is a bit odd, the idea I assume was to abstract the different places information can be in tokens to some "standard" return that we can generically read from. The confusing part is a) that it's not just the user_id, you're right and b) that it's a dict and not a well-defined object of some kind.

 I'd almost just rename this to something like `get_standard_authentication_info` or something. Because I do appreciate the abstraction of parsing any needed info from the tokens and returning a simple representation.
 
 idk I'm open to your thoughts here

Avantol13 (changes requested):

fence/config-default.yaml
```diff
@@ -181,6 +181,10 @@ OPENID_CONNECT:
     client_secret: ''
     redirect_url: '{{BASE_URL}}/login/ras/callback'
     scope: 'openid email profile ga4gh_passport_v1'
+#   The id_token claim and claim values for the RAS IdP that indicate MFA was used for authentication.
+#    multifactor_auth_claim_info:
+#      claim: 'acr'
+#      values: [ "https://sts.nih.gov/assurance/aal/2", "urn:gov:gsa:ac:classes:sp:PasswordProtectedTransport:duo" ]
     # if mock is true, will fake a successful login response for login
```
BinamB, line 188: Could you add an example of what the `is-mfa_enabled` param would look like in the default config. Maybe have it disabled in the test config too and enable it during the test, which you've already done

fence/blueprints/login/base.py
```diff
@@ -125,6 +133,20 @@ def get(self):
...
+        if self.is_mfa_enabled:
+            username = token_result.get(self.username_field)
+            if token_result.get("mfa"):
+                logger.info(f"Adding mfa_policy for {username}")
+                self.app.arborist.grant_user_policy(
+                    username=username,
+                    policy_id="mfa_policy",
+                )
+            else:
+                logger.info(f"Revoking mfa_policy for {username}")
```
BinamB, line 145: Open to your opinion but should we remove mfa_policy regardless of whether or not `is_mfa_enabled` is true. Im just thinking about a case where it was turned on and then turned off and the user still has the mfa policy

fence/resources/openid/cilogon_oauth2.py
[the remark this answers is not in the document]
Avantol13: I say leave it, might avoid some confusion in the future

fence/sync/sync_users.py
```diff
@@ -1868,6 +1868,35 @@ def _update_arborist(self, session, user_yaml):
 
         return True
 
+    def _revoke_all_policies_preserve_mfa(self, user):
```
Avantol13, line 1871: where is actually called from?

k-burt-uch: whoops

its called from here https://github.com/uc-cdis/fence/pull/1106/files#diff-5824dff91d0ccd46626b8d7fbb03bcc2eb6882c9c038139c48ccfd8c39082d70R2005

fence/blueprints/login/base.py
k-burt-uch: Revisiting this: I reverted my changes here. I think its better if usersync cleans up the mfa_policy if mfa is turned off rather than during login.

fence/blueprints/login/base.py
```diff
@@ -125,6 +133,22 @@ def get(self):
...
+                if token_result.get("mfa"):
+                    logger.info(f"Adding mfa_policy for {username}")
+                    self.app.arborist.grant_user_policy(
+                        username=username,
+                        policy_id="mfa_policy",
+                    )
+                    return
+                else:
+                    logger.info(f"Revoking mfa_policy for {username}")
+                    self.app.arborist.revoke_user_policy(
```
Avantol13, line 148: If MFA is enabled but they've never had an MFA policy added before and they don't use MFA, does this error out here? e.g. does revoke_user_policy "fail" silently if the policy doesn't exist for the user?

Avantol13 (approved):

tests/login/test_idp_oauth2.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def test_has_mfa_claim_multiple_acr(oauth_client):
+    has_mfa = oauth_client.has_mfa_claim({"acr": "mfa otp duo"})
+    assert has_mfa
+
+
+def test_does_not_has_mfa_claim(oauth_client):
+    has_mfa = oauth_client.has_mfa_claim({"acr": "pwd"})
+    assert not has_mfa
```
paulineribeyre, line 35:
```suggestion
    assert not has_mfa

    has_mfa = oauth_client.has_mfa_claim({"something": "mfa"})
    assert not has_mfa
```

BinamB (dismissed):

fence/blueprints/login/base.py
k-burt-uch: It doesn't error, there's just a log that appears saying the user doesn't have this policy

fence/sync/sync_users.py
```diff
@@ -1868,6 +1868,35 @@ def _update_arborist(self, session, user_yaml):
...
+
+            resources = self.arborist_client.list_resources_for_user(username)
+            # TODO This should be a diff, not a revocation of all policies.
+            self.arborist_client.revoke_all_policies_for_user(username)
+            has_revoked_policies = True
+            if "/multifactor_auth" in resources:
+                self.arborist_client.grant_user_policy(username, "mfa_policy")
+        except Exception as e:
+            self.logger.error(f"Error attempting to preserve MFA policy. {e}")
+            if not has_revoked_policies:
```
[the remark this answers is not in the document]
k-burt-uch: I'm worried (maybe overly so) the call to self.arborist_client.list_resources_for_user(username) may fail and the revoke_all_policies call won't go out.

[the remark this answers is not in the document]
paulineribeyre: if you pull those calls out of the `try` block, i think usersync will just fail in this ^ case

but if you're really worried, i think this would work too:
```
is_mfa_enabled = ...
etc

try:
    resources = self.arborist_client.list_resources_for_user(username)
finally:
    self.arborist_client.revoke_all_policies_for_user(username)

if "/multifactor_auth" in resources:
    try:
        self.arborist_client.grant_user_policy(username, "mfa_policy")
    except Exception as e:
        self.logger.error(f"Error attempting to preserve MFA policy. {e}")
```

not a big deal. i just think shorter `try` blocks are easier to maintain and debug

fence/config-default.yaml
```diff
@@ -111,6 +111,9 @@ OPENID_CONNECT:
     user_id_field: ''  # optional (default "sub"); claims field to get the user_id from
     email_field: ''  # optional (default "email"); claims field to get the user email from
     scope: ''  # optional (default "openid")
+    multifactor_auth_claim_info: # optional, include if you're using arborist to enforce mfa on a per-file level
+      claim: '' # claims field that indicates mfa, usually the acr or acm claim.
```
Avantol13, line 115: this says "usually" but there's a check later that errors if it's not acr or acm

fence/config.py
```diff
@@ -143,6 +143,13 @@ def post_process(self):
                 "Visa parsing on login is enabled but `ENABLE_VISA_UPDATE_CRON` is disabled!"
             )
 
+        for idp in self._configs.get("OPENID_CONNECT", {}).values():
+            mfa_info = idp.get("multifactor_auth_claim_info")
+            if mfa_info and mfa_info["claim"] not in ["amr", "acr"]:
+                logger.warning(
+                    f"multifactor_auth_claim_info is using {mfa_info['claim']}, which is neither AMR or ACR. Unable to determine if a user used MFA. Fence will continue and assume they have not used MFA."
+                )
```
paulineribeyre, line 151: logging the name of the idp causing the issue, for easier debugging

```suggestion
        for idp_id, idp in self._configs.get("OPENID_CONNECT", {}).items():
            mfa_info = idp.get("multifactor_auth_claim_info")
            if mfa_info and mfa_info["claim"] not in ["amr", "acr"]:
                logger.warning(
                    f"IdP '{idp_id}' is using multifactor_auth_claim_info '{mfa_info['claim']}', which is neither AMR or ACR. Unable to determine if a user used MFA. Fence will continue and assume they have not used MFA."
                )
```

docs/fence_multifactor_authentication_guide.md
```diff
@@ -0,0 +1,64 @@
+# Fence Multifactor Authentication Guide
+
+Fence is capable of using token claims from IdPs to identify when multifactor authentication (MFA) was used during the authentication process.
+
+## File Level Enforcement
+To restrict access to files to user who've authenticated with MFA, the following resource *MUST* be present for indexd record:
```
paulineribeyre, line 6:
```suggestion
To restrict access to files to user who've authenticated with MFA, the following resource *MUST* be present in the indexd record's `authz`:
```

fence/config-default.yaml
k-burt-uch: Fixed.

Avantol13 (dismissed):

fence/sync/sync_users.py
```diff
@@ -1868,6 +1868,35 @@ def _update_arborist(self, session, user_yaml):
...
+        except Exception as e:
+            self.logger.error(
+                f"Could not retrieve user's resources, revoking all policies anyway. {e}"
+            )
+        finally:
+            # TODO This should be a diff, not a revocation of all policies.
+            self.arborist_client.revoke_all_policies_for_user(username)
+
+        if "/multifactor_auth" in resources:
+            status_code = self.arborist_client.grant_user_policy(username, "mfa_policy")
```
paulineribeyre, line 1898: sorry for bringing this up late in the review process, but i think this should be more robust:
- we assume that a user having the `/multifactor_auth` resource always means they have the exact `mfa_policy` we expect
- i just noticed in the docs that we assume the `mfa_policy` has been created before enabling MFA and running usersync. If that's not the case, in the current state i think usersync will fail? We should at least make sure the `grant_user_policy()` statement doesn't break usersync if the `mfa_policy` doesn't exist but the user got the `/multifactor_auth` resource some other way.

Why don't we create the policy during usersync?
```
whether "auto_mfa_policy" exists or not, create or update it
revoke_all_policies_for_user()
if "auto_mfa_policy" in gen3authz.get_user()["policies"]:
    grant_user_policy(username, "auto_mfa_policy")
```
The only downside i see is if we need the `mfa_policy` to give different access in different Commons

k-burt-uch: grant_user_policy just logs an error if the request has anything other than 204 response code 
https://github.com/uc-cdis/gen3authz/blob/master/python/src/gen3authz/client/arborist/base.py#L810

So it won't break user_sync if the policy isn't configured in user.yaml. Worst case, we'll fail secure and /multifactor_auth data isn't accessible until the user.yaml is updated.  

> The only downside i see is if we need the mfa_policy to give different access in different Commons

Yeah, I'm trying to avoid hardcoding the permissions to mfa_policy because I wasn't positive of the needs of other commons and DCF's permissions may need to change later. 

Great tip on checking the policies in user though. I updated this code to not rely on the resource.

paulineribeyre (dismissed):

.travis.yml
```diff
@@ -1,14 +1,14 @@
...
   - "3.9"
 
 sudo: false
 
 cache: pip
 
-addons:
-  postgresql: "9.6"
+services:
+    - postgresql
```
paulineribeyre, line 11: since the default is 9.2, do we need to set the version? https://docs.travis-ci.com/user/database-setup/#using-a-different-postgresql-version
i'm not sure which version we use in prod but probably 9.6 since that's what we had on here

Avantol13: Yeah as far as I know, we were on >=9.6. We rely on something specific that was introduced around JSON or BYTES fields, I can't remember

k-burt-uch: I checked dcf prod and we use version 13 - but I couldn't find anything in cloud-automation that has the postgresql version set :(

k-burt-uch: alright I think I got postgresql-13 working with jammy. Had to steal a lot of code from NCI-GDC's repo https://github.com/NCI-GDC/psqlgraph/

paulineribeyre (approved):

PR #1110, opened 2023-08-03
fix(authlib): override authlib install with a forked version with a s…

Avantol13:
…ecurity patch

### Dependency updates
- temporarily override authlib install with a forked version with a security patch


paulineribeyre (approved):
https://github.com/uc-cdis/authlib/compare/v0.11...v0.11_CVE_patch_v1 lgtm

PR #1112, opened 2023-08-15
(PPS-411): Add bucket selection support for multi part upload

BinamB:

cdis-data-client PR (https://github.com/uc-cdis/cdis-data-client/pull/123)

### New Features
- Added upload bucket selection support for multipart upload
- Add support for non-aws s3 buckets for multipart upload


fence/blueprints/data/blueprint.py
```diff
@@ -176,17 +177,9 @@ def upload_data_file():
     )
 
     protocol = params["protocol"] if "protocol" in params else None
-    bucket = params.get("bucket")
+    bucket = params.get("bucket", None)
```
Avantol13, line 180: If should already default to None, you don't need this

fence/blueprints/data/blueprint.py
```diff
@@ -257,12 +256,18 @@ def generate_multipart_upload_presigned_url():
         default=default_expires_in,
     )
 
+    bucket = params.get("bucket", None)
```
Avantol13, line 259: .get() already returns None if it's not found, so the `, None` is redundant. You could just `.get("bucket")`

fence/blueprints/data/blueprint.py
```diff
@@ -257,12 +256,18 @@ def generate_multipart_upload_presigned_url():
...
+    if bucket:
+        _is_allowed_data_upload_bucket_configured(bucket)
+
+    # TODO add bucket param here
```
Avantol13, line 263: Is this a real TODO? like we need to do something in the future, can we do it now?

If this is just a note for the future in case we need to add things, don't use `TODO`, just say `note: ` or something

fence/blueprints/data/blueprint.py
```diff
@@ -285,6 +290,7 @@ def complete_multipart_upload():
         raise UserError("missing required arguments: {}".format(list(missing)))
 
     default_expires_in = flask.current_app.config.get("MAX_PRESIGNED_URL_TTL", 3600)
+    bucket = params.get("bucket", None)
```
Avantol13, line 293: same as prev comments, can just use .get(

fence/blueprints/data/blueprint.py
```diff
@@ -176,17 +177,9 @@ def upload_data_file():
...
-        s3_buckets = get_value(
-            flask.current_app.config,
-            "ALLOWED_DATA_UPLOAD_BUCKETS",
-            InternalError("ALLOWED_DATA_UPLOAD_BUCKETS not configured"),
-        )
-        if bucket not in s3_buckets:
-            logger.debug(f"Bucket '{bucket}' not in ALLOWED_DATA_UPLOAD_BUCKETS config")
-            raise Forbidden(f"Uploading to bucket '{bucket}' is not allowed")
-
+        _is_allowed_data_upload_bucket_configured(bucket)
```
Avantol13, line 182: the prefix "is" usually implies a boolean return. can we rename to something like "verify_data_upload_bucket_configuration" ?

fence/blueprints/data/blueprint.py
Avantol13, line 263: let's not leave any TODOs, if we need to add this now, let's do it. If this is a note for the future in case we have more, don't use `TODO` just say `note: ` or something

fence/blueprints/data/blueprint.py
Avantol13, line 293:
```suggestion
    bucket = params.get("bucket")
```

fence/blueprints/data/indexd.py
```diff
@@ -314,17 +314,18 @@ def init_multipart_upload(key, expires_in=None):
         Returns:
             uploadId(str)
         """
-        try:
-            bucket = flask.current_app.config["DATA_UPLOAD_BUCKET"]
-        except KeyError:
-            raise InternalError(
-                "fence not configured with data upload bucket; can't create signed URL"
-            )
+        if bucket is None:
```
Avantol13, line 317:
```suggestion
        if bucket:
```

If you use Truthiness, you don't have to worry about `None` being forced instead of empty strings

fence/blueprints/data/indexd.py
```diff
@@ -1611,3 +1629,14 @@ def filter_auth_ids(action, list_auth_ids):
...
+
+
+def _is_allowed_data_upload_bucket_configured(bucket):
+    s3_buckets = get_value(
+        flask.current_app.config,
+        "ALLOWED_DATA_UPLOAD_BUCKETS",
+        InternalError("ALLOWED_DATA_UPLOAD_BUCKETS not configured"),
+    )
+    if bucket not in s3_buckets:
+        logger.debug(f"Bucket '{bucket}' not in ALLOWED_DATA_UPLOAD_BUCKETS config")
```
Avantol13, line 1641: you should use .error here, not .debug because this is a legitimate error we want logged regardless of the log level

Avantol13: you could have an additional debug log that spits out the entire ALLOWED config though if you wanted

openapis/swagger.yaml
```diff
@@ -1717,10 +1717,15 @@ components:
             type: string
           description: requested authorization resources to be set on the
             resulting indexed record. You must have proper authorization to set this
+        bucket:
+          type: string
+          required: false
+          description: bucket to upload
```
Avantol13, line 1723:
```suggestion
          description: bucket to upload to
```

openapis/swagger.yaml
```diff
@@ -1738,6 +1743,10 @@ components:
         expires_in:
           type: integer
           description: optional integer specifying the presigned URL lifetime
+        bucket:
+          type: string
+          required: false
+          description: bucket to upload
```
Avantol13, line 1749:
```suggestion
          description: bucket to upload to
```

openapis/swagger.yaml
```diff
@@ -1767,6 +1776,10 @@ components:
...
```
Avantol13, line 1782:
```suggestion
          description: bucket to upload to
```

fence/blueprints/data/indexd.py
```diff
@@ -337,19 +338,33 @@ def complete_multipart_upload(key, uploadId, parts, expires_in=None):
         Returns:
             None if success otherwise an exception
         """
-        try:
-            bucket = flask.current_app.config["DATA_UPLOAD_BUCKET"]
-        except KeyError:
-            raise InternalError(
-                "fence not configured with data upload bucket; can't create signed URL"
+        if bucket:
+            s3_buckets = get_value(
```
Avantol13, line 342: is this logic exactly what `_is_allowed_data_upload_bucket_configured(bucket)` is doing? Can you reuse that here?

Avantol13 (changes requested):

fence/blueprints/data/indexd.py
```diff
@@ -1614,3 +1628,15 @@ def filter_auth_ids(action, list_auth_ids):
         if checked_permission in values:
             authorized_dbgaps.append(key)
     return authorized_dbgaps
+
+
+def verify_data_upload_bucket_configuration(bucket):
```
Avantol13, line 1633: docstring please

fence/blueprints/data/multipart_upload.py
```diff
@@ -3,15 +3,17 @@
...
+from fence.config import config
 from fence.errors import InternalError
 
 MAX_TRIES = 5
 
 logger = get_logger(__name__)
 
 
-def initilize_multipart_upload(bucket, key, credentials):
+def initilize_multipart_upload(bucket_name, key, credentials):
```
Avantol13, line 16: refactor name for typo: `initialize`, not `initilize`

fence/blueprints/data/blueprint.py
```diff
@@ -284,6 +288,7 @@ def complete_multipart_upload():
         raise UserError("missing required arguments: {}".format(list(missing)))
 
     default_expires_in = flask.current_app.config.get("MAX_PRESIGNED_URL_TTL", 3600)
+    bucket = params.get("bucket")
```
Avantol13, line 291: is there a reason you don't run the verify function after this one?

fence/blueprints/data/indexd.py
```diff
@@ -317,17 +317,18 @@ def init_multipart_upload(key, expires_in=None):
...
-        try:
-            bucket = flask.current_app.config["DATA_UPLOAD_BUCKET"]
-        except KeyError:
-            raise InternalError(
-                "fence not configured with data upload bucket; can't create signed URL"
-            )
+        if not bucket:
+            try:
+                bucket = flask.current_app.config["DATA_UPLOAD_BUCKET"]
+            except KeyError:
```
Avantol13, line 323: this error will never happen b/c `DATA_UPLOAD_BUCKET` is part of the default config, so it will always be set, it might just be ''. And this isn't catching the case where the bucket is actually empty ''.  

So you can simplify/fix this whole block with:

```
bucket = bucket or flask.current_app.config["DATA_UPLOAD_BUCKET"]

if not bucket:
    raise InternalError("fence not configured with data upload bucket; can't create signed URL")
```

fence/blueprints/data/indexd.py
```diff
@@ -1614,3 +1628,15 @@ def filter_auth_ids(action, list_auth_ids):
         if checked_permission in values:
             authorized_dbgaps.append(key)
     return authorized_dbgaps
+
+
+def verify_data_upload_bucket_configuration(bucket):
+    s3_buckets = get_value(
+        flask.current_app.config,
```
Avantol13, line 1635: you don't need to use `get_value` b/c `ALLOWED_DATA_UPLOAD_BUCKETS` is part of the default configuration, which guarantees its existence in the config dictionary. If you're trying to check if it's an empty list, then this isn't catching that either.
```
s3_buckets = flask.current_app.config["ALLOWED_DATA_UPLOAD_BUCKETS"]

if not s3_buckets:
    raise InternalError("ALLOWED_DATA_UPLOAD_BUCKETS not configured")
```

fence/blueprints/data/indexd.py
Avantol13: I think you still left a chunk of code you don't need. If you're calling verify you don't need to get the s3_buckets here

fence/blueprints/data/indexd.py
```diff
@@ -340,19 +341,29 @@ def complete_multipart_upload(key, uploadId, parts, expires_in=None):
...
+        if bucket:
+            s3_buckets = get_value(
+                flask.current_app.config,
+                "ALLOWED_DATA_UPLOAD_BUCKETS",
+                InternalError("ALLOWED_DATA_UPLOAD_BUCKETS not configured"),
             )
+            verify_data_upload_bucket_configuration(bucket)
+        else:
+            try:
+                bucket = flask.current_app.config["DATA_UPLOAD_BUCKET"]
```
Avantol13, line 353: see previous comment about default configs. this will never KeyError

fence/blueprints/data/indexd.py
```diff
@@ -364,12 +375,15 @@ def generate_aws_presigned_url_for_part(key, uploadId, partNumber, expires_in):
...
-        except KeyError:
-            raise InternalError(
-                "fence not configured with data upload bucket; can't create signed URL"
-            )
+        if bucket:
+            verify_data_upload_bucket_configuration(bucket)
+        else:
+            try:
+                bucket = flask.current_app.config["DATA_UPLOAD_BUCKET"]
+            except KeyError:
```
Avantol13, line 383: see previous comment

fence/blueprints/data/multipart_upload.py
```diff
@@ -95,7 +125,7 @@ def complete_multipart_upload(bucket, key, credentials, uploadId, parts):
 
 
 def generate_presigned_url_for_uploading_part(
```
Avantol13, line 127: these are all positive tests. I know they were here before, but can you add some for the error cases like when DATA_UPLOAD and/or ALLOWED_DATA_UPLOAD_BUCKETS buckets aren't specified / are empty?

Avantol13 (changes requested):

fence/blueprints/data/indexd.py
BinamB: There's a test that deletes that configuration so this will fail. https://github.com/uc-cdis/fence/blob/master/tests/data/test_blank_index.py#L351
I don't think it gets another default value.
Should I just set the test to empty the field rather than delete the config since that would be more realistic?

fence/blueprints/data/multipart_upload.py
BinamB: Sounds like my previous comment might resolve this test case? 🤔

lbeckman314:
Thanks Binam for your patience, working on the unit tests issues blocking #1115 which targets this branch (let us know if we should specify a different branch to merge into!).

I will update here and in Slack when the tests are running successfully and it's ready for review!

> https://github.com/uc-cdis/fence/pull/1115

MaribelleHGomez (dismissed):

Avantol13 (approved):

PR #1114, opened 2023-09-05
Chore/pps-301

tianj7:
Jira Ticket: https://ctds-planx.atlassian.net/browse/PXP-9212

- Describe what this pull request does.
Rewrite Fence project with work with Authlib 1.2.1

### Breaking Changes

- New Database Migration (using `alembic`) was added and is required. 

### Dependency updates
- Update Authlib Version by Major Version to post-1.0 release: 1.2.1


Avantol13:
update the PR description to include the "Breaking Changes" section, note that a db migration is required

bin/fence_create.py
```diff
@@ -329,6 +329,9 @@ def parse_arguments():
         help='scopes to include in the token (e.g. "user" or "data")',
     )
     token_create.add_argument("--exp", help="time in seconds until token expiration")
+    token_create.add_argument(
```
Avantol13, line 332: is this optional, so it will run without it to work for access tokens?

migrations/models/migration_client.py
```diff
@@ -0,0 +1,67 @@
...
+from sqlalchemy import Boolean, Column, Integer, String, Text, func
+from sqlalchemy.orm import Session, backref, relationship
+from sqlalchemy.schema import ForeignKey
+from userdatamodel import Base
+from userdatamodel.models import User
+import time
+
+# This needs to be in a different file
+# Otherwise SqlAlchemy would import this multiple times and then complain about metadata conflict
+class MigrationClient(Base, OAuth2ClientMixin):
```
Avantol13, line 11: do you need to inherit here? inheriting from base makes sqlalchemy actually create this table in the database and do other initialization steps. Can we just have this be an isolated class of data?

migrations/versions/9b3a5a7145d7_authlib_update_1_2_1.py
```diff
@@ -0,0 +1,246 @@
...
+    op.add_column(
+        "client",
+        sa.Column(
+            "client_secret_expires_at", sa.Integer(), nullable=False, server_default="0"
+        ),
+    )
+
+    # Modify Columns for client Table
+    op.alter_column("client", "issued_at", new_column_name="client_id_issued_at")
+    op.alter_column("client", "client_id", nullable=False, type_=sa.String(48))
```
Avantol13, line 37: I assume these are safe operations that automatically migrate existing data?

migrations/versions/9b3a5a7145d7_authlib_update_1_2_1.py
```diff
@@ -0,0 +1,246 @@
...
+# revision identifiers, used by Alembic.
+revision = "9b3a5a7145d7"  # pragma: allowlist secret
+down_revision = "a04a70296688"  # pragma: allowlist secret
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Add New Columns for client Table
+    op.add_column("client", sa.Column("client_metadata", sa.Text(), nullable=True))
```
Avantol13, line 27: I'm still a little confused on the overall process. So you're:

1. Add New Columns for client Table
2. Modify Columns for client Table
3. Set value of metadata field (this is confusingly named b/c it does waaay more than just set the metadata field. It copies everything into a new migration_client table from client, then modifies the migration_client table, deletes client, recreates client from migration_client)

I think I'm confused b/c some of the migration happens on the client table before you delete it and do the rest on the migration_client. Why not just do the whole migration on the migration_client, read from client and get in-mem current clients, convert to migration_clients and commit them to that table, then do the swap (delete client, replace with migration_client).

OR alternatively, do you even need the migration_client table? can you just load all current client data into memory, modify the client table schema, and then ensure the data in that table gets loaded? There is a bit more risk there... b/c if we remove all the clients or mess up the db and then the app crashes we lose the in-mem representations.. so maybe the second table is a bit safer

fence/models.py
```diff
@@ -295,10 +322,7 @@ def client_type(self):
             return "public"
         return "confidential"
 
-    @property
-    def default_redirect_uri(self):
-        return self.redirect_uris[0]
-
+    ##Deprecated, Not called anywhere currently
```
Avantol13, line 325: let's remove then

pyproject.toml
```diff
@@ -31,7 +23,10 @@ cdiserrors = "<2.0.0"
 cdislogging = "^1.0.0"
 cdispyutils = "^2.0.1"
 cryptography = ">=41.0.2"
-flask = ">=2.2.3"
+#Pegging flask to <= 2.3.3 to maintain compatibilty with pytest-flask
+#Pytest 1.2.0 (latest on 10-09-2023) still doesn't work with flask 3.0.0
+#https://github.com/pytest-dev/pytest-flask/issues/167
```
Avantol13, line 28: this has been fixed in pytest 1.3.0. How much work would it be to bump to 3.0.0? If you think it's too much, maybe we should make a separate PR after we merge this one..

tests/migrations/test_9b3a5a7145d7.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # clean up the client table
+    with app.db.session as db_session:
+        db_session.query(Client).delete()
+
+    # go back to the latest state of the DB
+    alembic_main(["--raiseerr", "upgrade", "head"])
+
+
+def test_upgrade(app):
```
Avantol13, line 26: add a quick docstring please

migrations/versions/9b3a5a7145d7_authlib_update_1_2_1.py
```diff
@@ -0,0 +1,328 @@
...
+revision = "9b3a5a7145d7"  # pragma: allowlist secret
+down_revision = "a04a70296688"  # pragma: allowlist secret
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+
+    # Make a copy of client table
+    copy_client_to_temp_and_clear_data(op)
```
Avantol13, line 29: can you pass in the temp table name? `migration_client`

Just so that it's easy to follow this high level function and it's clear what you're dropping when you 

```
op.drop_table("migration_client")
```

migrations/versions/9b3a5a7145d7_authlib_update_1_2_1.py
```diff
@@ -0,0 +1,328 @@
...
+    set_old_column_values()
+    op.drop_table("migration_client")
+
+    # Remove New Columns for authorization_code Table
+    op.drop_column("authorization_code", "code_challenge")
+    op.drop_column("authorization_code", "code_challenge_method")
+
+
+def copy_client_to_temp_and_clear_data(op):
+    conn = op.get_bind()
```
Avantol13, line 143: let's get some docstrings for these functions

migrations/versions/9b3a5a7145d7_authlib_update_1_2_1.py
```diff
@@ -0,0 +1,328 @@
...
+        data["client_name"] = None
+
+        if client.client_metadata:
+            metadata = json_loads(client.client_metadata)
+            data["i18n_metadata"] = client.client_metadata
+        else:
+            metadata = {}
+            data["i18n_metadata"] = None
+
+        if "redirect_uris" in metadata:
```
Avantol13, line 241: You can simplify these to:

```
data["redirect_uri"] = metadata.get("redirect_uris")
```

migrations/versions/9b3a5a7145d7_authlib_update_1_2_1.py
```diff
@@ -0,0 +1,328 @@
...
+            data["token_endpoint_auth_method"] = metadata["token_endpoint_auth_method"]
+        else:
+            data["token_endpoint_auth_method"] = None
+
+        if "scope" in metadata:
+            data["_allowed_scopes"] = metadata["scope"]
+        else:
+            data["_allowed_scopes"] = None
+
+        if "grant_types" in metadata and metadata["grant_types"]:
```
Avantol13, line 256: I think you can simplify to:

```
data["grant_type"] = "\n".join(metadata.get("grant_types", "")) or ""
```

Avantol13 (changes requested):

fence/oidc/oidc_server.py
```diff
@@ -53,6 +71,54 @@ def init_app(self, app, query_client=None, save_token=None):
...
             self.authenticate_client = ClientAuthentication(query_client)
+
+    # 2023-09-29
+    # Below code replaces authlib functions. It does the same thing as authlib 1.2.1 except it returns grant_scope from
+    # either args or forms. Authlib 1.2.1 forces grant_type to be part of post request body which isn't the current use case.
+    # https://github.com/lepture/authlib/blob/a6e89f8e6cf6f6bebd63dcdc2665b7d22cf0fde3/authlib/oauth2/rfc6749/requests.py#L59C10-L59C10
+    def create_token_response(self, request=None):
+        """Validate token request and create token response.
+
+        :param request: HTTP request instance
```
Avantol13, line 84: Use Google style docstrings

Avantol13 (changes requested):

Avantol13 (dismissed):
Thanks for all the hard work on this. Make sure to clearly state in the `Breaking Changes` section of the PR description that this requires a db migration

MichaelLukowski (approved):

paulineribeyre (approved):

fence/oidc/endpoints.py
```diff
@@ -109,3 +109,25 @@ def create_revocation_response(self):
         finally:
             body = {"error": message} if message != "" else {}
         return (status, body, headers)
+
+
+class JWTToken(object):
+    def __init__(self, token):
+        self.encoded_string = token
+        self.client_id = jwt.decode(
+            token, algorithms=["RS256"], options={"verify_signature": False}
```
pieterlukasse, line 118: why ``"verify_signature": False`?

tianj7: I was just keeping it the same as previous used logic. I believe from this comment that it was done for compatibility reasons https://github.com/uc-cdis/fence/blob/234638e3221ff470fb497113badcfe47ed355654/fence/jwt/validate.py#L130

it's from 3 years ago so we probably need to review this and create a new ticket to address this if appropriate.
https://github.com/uc-cdis/fence/blame/234638e3221ff470fb497113badcfe47ed355654/fence/jwt/validate.py#L130

PR #1115, opened 2023-09-12
Update support for multipart uploads to non-AWS S3 endpoints

lbeckman314:
### New Features
- Updates support for retrieving correct endpoints and credentials for non-AWS bucket parameters.

### Breaking Changes
- None

### Bug Fixes
- Passes bucket parameter if present to fence/blueprints/data/indexd.py:init_multipart_upload function.

### Dependency updates
- None

### Deployment changes
- None

### DIFF
--- a/fence/blueprints/data/multipart_upload.py
+++ b/fence/blueprints/data/multipart_upload.py
@@ -3,7 +3,9 @@ from botocore.exceptions import ClientError
 from retry.api import retry_call
 
 from cdispyutils.hmac4 import generate_aws_presigned_url
+from cdispyutils.config import get_value
 from cdislogging import get_logger
+from fence.config import config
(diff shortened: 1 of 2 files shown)


fence/blueprints/data/multipart_upload.py
```diff
@@ -23,17 +25,30 @@ def initilize_multipart_upload(bucket, key, credentials):
     Returns:
         UploadId(str): uploadId
     """
+    s3_buckets = get_value(
```
lbeckman314, line 28: This section of code is repeated multiple times below and could be streamlined into it's own function.

BinamB (commented):
I was comparing it to the PR I have, could you resolve the conflicts? Everything else looks fine.

PR #1116, opened 2023-09-27
Remove Travis, add GH Action for Unit Test

k-burt-uch:
### Improvements
- Removes travis CI, replace with GH Actions


.github/workflows/ci.yaml
```diff
@@ -1,25 +1,27 @@
 name: CI
-
 on:
-    push:
+  push:
     pull_request:
       types: [opened, reopened]
```
paulineribeyre, line 5: `pull_request` is too indented

paulineribeyre (approved):

.github/workflows/ci.yaml
```diff
@@ -1,25 +1,27 @@
...
+    uses: uc-cdis/.github/.github/workflows/python_unit_test.yaml@master
+    with:
+       python-version: '3.9'
+       test-script: 'tests/ci_commands_script.sh'
+       run-coveralls: true
   ci:
     name: Build Image and Push
-    # TODO Add this line back once we update to Python 3.9 from 3.6
-    # needs: Security
+    needs: Security
```
k-burt-uch, line 24: Do we still need this?

k-burt-uch:
This fell through the cracks :( 

The python unit tests are now passing, so this is technically done but I'm not sure if the security flag on the docker push is necessary.

paulineribeyre (approved):

.github/workflows/ci.yaml
paulineribeyre: i think so - it means we don't build and push if the security check did not succeed. (arguably not great for development, but 🤷‍♀️ )

k-burt-uch: its easy to remove if it becomes a barrier.

paulineribeyre (approved):
rerunning the CI

PR #1117, opened 2023-10-05
Downloads through GA4GH DRS are recorded in audit service

paulineribeyre:
### Improvements
- Downloads through GA4GH DRS are recorded in audit service

### DIFF
--- a/fence/resources/audit/utils.py
+++ b/fence/resources/audit/utils.py
@@ -43,7 +43,7 @@ def create_audit_log_for_request(response):
     TODO The audit service has the ability to record presigned URL "upload" logs but we are not
     currently sending those logs. We would need to:
     - add the `@enable_audit_logging` decorator to `init_multipart_upload` (single upload requests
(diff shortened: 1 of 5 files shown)


fence/resources/audit/utils.py
```diff
@@ -39,6 +39,13 @@ def create_audit_log_for_request(response):
     in `enable_audit_logging` decorator), record an audit log. The data we
     need to record the logs are stored in `flask.g.audit_data` before reaching
     this code.
+
+    TODO The audit service has the ability to record presigned URL "upload" logs but we are not
```
k-burt-uch, line 43: Do we need a ticket for this TODO?

k-burt-uch (approved):

fence/resources/audit/utils.py
paulineribeyre: We can create a ticket but tbh this is more of an explanation for future-me looking for upload audit logs 🙂

PR #1122, opened 2023-11-16
Pull resources from auth_mapping

k-burt-uch:
Link to JIRA Issue: [PXP-11181](https://ctds-planx.atlassian.net/browse/PXP-11181)
### Bug Fixes
- Fixes issue where users not present in arborist would have blank resources and authz fields when they should have values for open access data within the commons. 

### Deployment changes

[PXP-11181]: https://ctds-planx.atlassian.net/browse/PXP-11181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -259,14 +259,14 @@
         "filename": "tests/conftest.py",
         "hashed_secret": "1348b145fa1a555461c1b790a2f66614781091e9",
         "is_verified": false,
-        "line_number": 1556
+        "line_number": 1559
       },
       {
         "type": "Base64 High Entropy String",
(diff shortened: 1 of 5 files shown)


Avantol13 (changes requested):
change looks good but can we add a regression test where we simulate `list_resources_for_user` raising a 404, hitting userinfo, and make sure the keys from auth_mapping are in the response?

Avantol13 (approved):

PR #1124, opened 2023-12-01
Feat/metrics

Avantol13:
### Improvements
- Add a new log that includes data information per signed URL generation which includes the size, bucket used, authz, acl, and requesting user

### DIFF
--- a/fence/blueprints/data/indexd.py
+++ b/fence/blueprints/data/indexd.py
@@ -168,7 +168,7 @@ def get_signed_url_for_file(
 
 
 def _log_signed_url_data_info(index_document, user_sub):
-    size = index_document.get("size")
+    size_in_kibibytes = index_document.get("size", 0) / 1024
     acl = index_document.get("acl")
     authz = index_document.get("authz")
     buckets = set()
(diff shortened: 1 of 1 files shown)


jawadqur (approved):

fence/blueprints/data/indexd.py
```diff
@@ -159,9 +159,41 @@ def get_signed_url_for_file(
...
     return {"url": signed_url}
 
 
+def _log_signed_url_data_info(indexed_file, user_sub, requested_protocol):
+    size_in_kibibytes = indexed_file.index_document.get("size", 0) / 1024
+    acl = indexed_file.index_document.get("acl")
+    authz = indexed_file.index_document.get("authz")
+
+    # the behavior later on is to pick the 1st location as the signed URL if a protocol is not requested
+    protocol = requested_protocol or indexed_file.indexed_file_locations[0].protocol
```
jawadqur, line 177: Would this run into an error situation if the `indexed_file.indexed_file_locations[0]` is empty array? Not sure if that's possible

Avantol13: oh! nice catch. Let me double check, it might be a problem. I think to get to this point there has to be at least 1 (it'll error before this), but I need to double check

Avantol13: okay, we're good. there's no way to get to this point without a file location b/c it will error when it tries to generate the signed url here: https://github.com/uc-cdis/fence/blob/d9f9aac95e2c46d42916461986d9d438dfdcdc91/fence/blueprints/data/indexd.py#L582 which happens before this

jawadqur: Awesome! 👍

PR #1125, opened 2023-12-04
Fix MFA policy management with unknown users

k-burt-uch:
### Bug Fixes
- Fixes issue with usersync and MFA policy management in cases where a user has not been persisted by Fence.

### DIFF
--- a/fence/sync/sync_users.py
+++ b/fence/sync/sync_users.py
@@ -1995,6 +1995,7 @@ class UserSyncer(object):
         for username, user_project_info in user_projects.items():
             self.logger.info("processing user `{}`".format(username))
             user = query_for_user(session=session, username=username)
+            idp = None
             if user:
                 username = user.username
                 idp = user.identity_provider.name if user.identity_provider else None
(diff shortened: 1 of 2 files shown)


paulineribeyre (approved):
looks

paulineribeyre (commented):
we might need a unit test covering the use case of a user that doesn't exist in the db

fence/sync/sync_users.py
```diff
@@ -1886,7 +1890,7 @@ def _revoke_all_policies_preserve_mfa(self, user):
 
         policies = []
         try:
-            policies = self.arborist_client.get_user()["policies"]
```
paulineribeyre, line 1889: not sure how this ever worked... `get_user` is missing the required `username` parameter

tests/dbgap_sync/test_user_sync.py
```diff
@@ -1033,6 +1033,20 @@ def test_user_sync_with_visa_sync_job(
...
+    Test that function returns even when there's no user
+    """
+    # no arborist user with that username
+    user_that_doesnt_exist = "foobar"
+    syncer.arborist_client.get_user.return_value = None
+
+    syncer._revoke_all_policies_preserve_mfa(user_that_doesnt_exist, "mock_idp")
+
+    # we only care that this doesn't error
+    assert True
```
paulineribeyre, line 1048:
```suggestion
    # we only care that this doesn't error
    syncer._revoke_all_policies_preserve_mfa(user_that_doesnt_exist, "mock_idp")
```

paulineribeyre (approved):

PR #1130, opened 2023-12-15
chore(logs): improve signed URL data metrics log by adding client_id …

Avantol13:
…from token

### Improvements
-  improve signed URL data metrics log by adding client_id from token


fence/blueprints/data/indexd.py
```diff
@@ -192,10 +193,21 @@ def _log_signed_url_data_info(indexed_file, user_sub, requested_protocol):
                 break
 
     logger.info(
-        f"Signed URL Generated. size_in_kibibytes={size_in_kibibytes} acl={acl} authz={authz} bucket={bucket} user_sub={user_sub}"
+        f"Signed URL Generated. size_in_kibibytes={size_in_kibibytes} "
+        f"acl={acl} authz={authz} bucket={bucket} user_sub={user_sub} client_id={client_id}"
```
k-burt-uch, line 197: Ahhh, I don't think we should log empty string here if the azp claim isn't present. Seeing that would look like a bug to me. 

Something more explicit like "Unknown Client" would be nicer.

Avantol13: I can do that :+1:

k-burt-uch (approved):

PR #1142, opened 2024-04-18
Update sync_users.py

k-burt-uch:

### Bug Fixes
- Suppress ssh banner when connecting.


k-burt-uch:
The ssh banner was a red herring and this PR is irrelevant

PR #1147, opened 2024-05-21
Upload to existing indexd record

BinamB:
Change reflected in gen3-client https://github.com/uc-cdis/cdis-data-client/pull/131

### Improvements
- `/upload` endpoint accepts optional `guid` in request body. If the `guid` exists in indexd, it does not create a new record, else it creates a blank record with the provided `guid`. 
- `/multipart/init` endpoint accepts optional `guid` in request body. If the `guid` exists in indexd, it does not create a new record, else it creates a blank record with the provided `guid`.


fence/blueprints/data/blueprint.py
```diff
@@ -134,6 +134,8 @@ def upload_data_file():
     authz = params.get("authz")
     uploader = None
 
+    guid = params.get("did")
```
Avantol13, line 137: I would prefer we leave `did` an implementation detail in indexd. Can we call this `guid`?

fence/blueprints/data/blueprint.py
```diff
@@ -165,7 +167,10 @@ def upload_data_file():
         )
 
     blank_index = BlankIndex(
-        file_name=params["file_name"], authz=params.get("authz"), uploader=uploader
+        file_name=params["file_name"],
+        authz=params.get("authz"),
+        uploader=uploader,
+        guid=guid,
```
Avantol13, line 173: What happened if an invalid or non-existent GUID is specified? Should the behavior be to 400 or move forward creating a blank record? I think we need to consider things like this. If I supply a GUID to the public API, I probably don't _want_ a blank record. So if I mistyped the GUID I'd actually expect an error, not for it to move forward and create a blank record anyway

fence/blueprints/data/indexd.py
```diff
@@ -266,6 +272,20 @@ def index_document(self):
...
         """
+
+        if self.guid:
+            index_url = self.indexd.rstrip("/") + "/index/" + self.guid
+            indexd_response = requests.get(index_url)
+            if indexd_response.status_code == 200:
+                document = indexd_response.json()
+                self.guid = document["did"]
+                self.logger.info(f"Record with {self.guid} id found in Indexd.")
+                return document
```
Avantol13, line 283: I'm not a huge fan of early returns in a longer function like this since at a glance it's unclear where things end. Maybe just breaking some of this into a helper function(s) would help. `document = _create_blank_record()` and put everything after 288 in there basically. This is nitpicky so you don't need to do this, I just think the readability suffers with early returns unless the function itself is pretty small

tests/data/test_data.py
```diff
@@ -916,6 +960,26 @@ def test_non_public_authz_and_public_acl_object_upload_file(
     Test that a user can successfully generate an upload url for an Indexd
     record with a non-public authz field and a public acl field.
     """
+    did = str(uuid.uuid4())
```
Avantol13, line 963: I think you need to modify these to ensure you capture these test cases:

- When the guid is provided and it already exists
- When the guid is provided and it doesn't exist
- When the guid is not provided

I believe you're only capturing the second one right now

Avantol13 (changes requested):

fence/blueprints/data/blueprint.py
```diff
@@ -165,7 +167,10 @@ def upload_data_file():
         )
 
     blank_index = BlankIndex(
-        file_name=params["file_name"], authz=params.get("authz"), uploader=uploader
+        file_name=params["file_name"],
+        authz=params.get("authz"),
```
Avantol13, line 171: this was already there, but I think this can just be `authz`

fence/blueprints/data/indexd.py
```diff
@@ -266,6 +272,21 @@ def index_document(self):
                 response from indexd (the contents of the record), containing ``guid``
                 and ``url``
         """
+
+        if self.guid:
+            index_url = self.indexd.rstrip("/") + "/index/" + self.guid
```
Avantol13, line 277: We may need to percent-encode the guid b/c it's user input, so I'm wondering if there's some security concerns here (like what if I pass in slashes, `../admin/` as the guid. I don't know if requests does that by default and this might get flagged as unsanitized user input in our security scans if it hasn't already. Not sure if/how we handle that elsewhere, just a thought

fence/blueprints/data/indexd.py
```diff
@@ -253,8 +259,8 @@ def __init__(
         self.file_name = file_name
         self.authz = authz
 
-        # if a guid is not provided, this will create a blank record for you
-        self.guid = guid or self.index_document["did"]
+        self.guid = guid
+        self.guid = self.index_document["did"]
```
Avantol13, line 263: This is somewhat confusing if you don't realize that index_document is a cached property. I'd add a comment above this to that effect. ".index_document is a cached property with code below, it creates/retrieves the actual record and this line updates the stored GUID to the returned record"

Avantol13: b/c otherwise it looks like you're mistakenly overwriting the GUID that someone passed in

fence/blueprints/data/indexd.py
```diff
@@ -266,6 +272,21 @@ def index_document(self):
                 response from indexd (the contents of the record), containing ``guid``
                 and ``url``
         """
+
+        if self.guid:
+            index_url = self.indexd.rstrip("/") + "/index/" + self.guid
+            indexd_response = requests.get(index_url)
+            if indexd_response.status_code == 200:
+                document = indexd_response.json()
+                self.guid = document["did"]
```
Avantol13, line 281: this is also set outside this cached_property.. I'd remove this and mirror the way the blank ID works (it just returns the whole document and let's the init set the guid)

fence/blueprints/data/indexd.py
Avantol13: Great updates, looks much cleaner now, thanks!

openapis/swagger.yaml
```diff
@@ -565,6 +565,8 @@ paths:
         Previous authorization check requires a more general, global upload permission:
         "file_upload" on "/data_file" resource. When "authz" is *not* provided, this
         endpoint will check for that permission for your user.
+
+        Accepts a "did" field in the request body. If "did" is provided, it checks indexd for an existing record. If not found, a new blank record will be created.
```
Avantol13, line 569: `guid` not did

openapis/swagger.yaml
```diff
@@ -666,7 +668,10 @@ paths:
         flow, Fence needs to provide a list of endpoints for supporting multipart upload presigned url
         This is the first step on the API side for the multipart upload presigned url. This endpoint
         causes fence to make a request to indexd to create a new, blank index record, and returns
-        the GUID for this new record and an uploadId for multipart upload presigned url
+        the GUID for this new record and an uploadId for multipart upload presigned url.
+
+
+        Accepts a "did" field in the request body. If "did" is provided, it checks indexd for an existing record. If not found, a new blank record will be created.
```
Avantol13, line 674: `guid`

tests/data/test_data.py
```diff
@@ -1664,6 +1730,57 @@ def json(self):
         assert "uploadId" in response.json
 
 
+@pytest.mark.parametrize("indexd_200_response", [True, False])
+def test_initialize_multipart_upload_with_guid_in_request(
```
Avantol13, line 1734: docstring please

tests/data/test_data.py
```diff
@@ -1664,6 +1730,57 @@ def json(self):
...
+            "Authorization": "Bearer " + encoded_creds_jwt.jwt,
+            "Content-Type": "application/json",
+        }
+        file_name = "asdf"
+        data = json.dumps({"file_name": file_name, "guid": did})
+        response = client.post("/data/multipart/init", headers=headers, data=data)
+
+        if indexd_200_response:
+            assert response.status_code == 201, response
+            assert "guid" in response.json
```
Avantol13, line 1778: can you also check that the content of the guid is the same value as the one passed into the endpoint?

fence/blueprints/data/indexd.py
```diff
@@ -266,6 +272,21 @@ def index_document(self):
...
+                self.logger.info(f"Record with {self.guid} id found in Indexd.")
+                return document
+            else:
+                raise NotFound("No indexed document found with id {}".format(self.guid))
```
Avantol13, line 285: Nice, yeah I think this is the right logic if I pass a guid 👍

Avantol13: Any reason you're using .format over f-strings here? There are other places with f-strings, I tend to prefer those. Not a big deal, just some inconsistency

Avantol13 (changes requested):

openapis/swagger.yaml
```diff
@@ -565,6 +565,8 @@ paths:
         Previous authorization check requires a more general, global upload permission:
         "file_upload" on "/data_file" resource. When "authz" is *not* provided, this
         endpoint will check for that permission for your user.
+
+        Accepts a "guid" field in the request body. If "guid" is provided, it checks indexd for an existing record. If not found, a new blank record will be created.
```
Avantol13, line 569: It doesn't seem like this is the behavior anymore? If it's not found, it raises a 404

openapis/swagger.yaml
```diff
@@ -1686,6 +1691,10 @@ components:
           description: >-
             the requested bucket to upload to.
             If not provided, defaults to the configured DATA_UPLOAD_BUCKET.
+        guid:
```
Avantol13, line 1694: In addition to this new param, we also need to update the example responses to include the 404 that can now happen

tests/data/test_data.py
```diff
@@ -1664,6 +1730,76 @@ def json(self):
         assert "uploadId" in response.json
 
 
+@pytest.mark.parametrize("indexd_200_response", [True, False])
```
Avantol13, line 1733: these should be separate unit tests. One is testing when the GUID exists, one is testing when it doesn't. If you want to share some code, you can create helper functions, but I'd rather keep the positive/negative testing as separate unit tests

Avantol13 (changes requested):

Avantol13 (approved):

PR #1149, opened 2024-05-23
Modify the register form to assume DUA consent.

frankliuao:
Updating the register form so its asks the user to agree to our DUA when submitting the form. Previously, the form only states that the user is submitting it. 

https://ctds-planx.atlassian.net/browse/MIDRC-754

### DIFF
--- a/fence/templates/register_user.html
+++ b/fence/templates/register_user.html
@@ -15,7 +15,7 @@
 {% block content %}
   <form method="post">
 
-    <h1 class="introduction">Register in order to get access to download data</h1>
+    <h1 class="introduction">Register* in order to get access to download data</h1>
     {{ form.csrf_token }}
     <ul>
       <li> {{ form.firstname.label }} {{ form.firstname }} </li>
@@ -30,6 +30,7 @@
       </li>
     </ul>
     <button class="button-primary-orange" type="submit">Register</button>
+    <p>By registering and submitting this form, you hereby acknowledge and consent to be bound by our terms and conditions, and our data use agreement (DUA), both of which can be found on our main website, or on our data access portal. </p>
   </form>
 {% endblock %}
 
--- a/fence/templates/register_user.html
+++ b/fence/templates/register_user.html
@@ -30,7 +30,7 @@
       </li>
     </ul>
     <button class="button-primary-orange" type="submit">Register</button>
-    <p>By registering and submitting this form, you hereby acknowledge and consent to be bound by our terms and conditions, and our data use agreement (DUA), both of which can be found on our main website, or on our data access portal. </p>
+    <p>By registering and submitting this form, you hereby acknowledge and consent to be bound by our terms and conditions, and our data use agreement (DUA), both of which can be found on our main website, or on our <a href="/dashboard/Public/documentation/DUA.html">data access portal</a>. </p>
   </form>
 {% endblock %}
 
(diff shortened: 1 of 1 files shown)


fence/templates/register_user.html
```diff
@@ -30,6 +30,7 @@ <h1 class="introduction">Register in order to get access to download data</h1>
       </li>
     </ul>
     <button class="button-primary-orange" type="submit">Register</button>
+    <p>By registering and submitting this form, you hereby acknowledge and consent to be bound by our terms and conditions, and our data use agreement (DUA), both of which can be found on our main website, or on our data access portal. </p>
```
paulineribeyre, line 33: it think it would be fine to add a link to `/dashboard/Public/documentation/DUA.html` since the file name is generic

fence/templates/register_user.html
```diff
@@ -15,7 +15,7 @@ <h1>{% block title %}Register{% endblock %}</h1>
 {% block content %}
   <form method="post">
 
-    <h1 class="introduction">Register in order to get access to download data</h1>
+    <h1 class="introduction">Register* in order to get access to download data</h1>
```
paulineribeyre, line 18: why the asterisk?

fence/templates/register_user.html
```diff
@@ -15,7 +15,7 @@ <h1>{% block title %}Register{% endblock %}</h1>
 {% block content %}
```
paulineribeyre, line 15: update PR description as discussed

frankliuao: Done.

fence/templates/register_user.html
frankliuao: A comment to register. 
Comment in `<p>`.

fence/templates/register_user.html
paulineribeyre: the link to the DUA shouldn't be on "data access portal" but on "data use agreement (DUA)"

fence/templates/register_user.html
```diff
@@ -30,6 +30,7 @@ <h1 class="introduction">Register in order to get access to download data</h1>
       </li>
     </ul>
     <button class="button-primary-orange" type="submit">Register</button>
+    <p>By registering and submitting this form, you hereby acknowledge and consent to be bound by our terms and conditions, and our data use agreement (DUA), both of which can be found on our main website, or on our <a href="/dashboard/Public/documentation/DUA.html">data access portal</a>. </p>
```
paulineribeyre, line 33: can we link directly to the terms and conditions as well? i don't see where the link is on the portal

fence/templates/register_user.html
paulineribeyre: asterisks make sens when they refer to something else - the paragraph would need to start with an asterisk as well

but here i don't think it's necessary at all, the text is right next to the submit button and visible enough

although it might be better above the button? so you read it before you click submit

fence/templates/register_user.html
frankliuao: Some terms and conditions are on our main website (www.midrc.org) so it's hard to find a generic url for it. It'll add MIDRC-specific contents to this page which is what I'm trying to avoid. I just checked and I think DUA itself is fine. I've modified it so it only says DUA.

fence/templates/register_user.html
frankliuao: Sure. Edited.

PR #1155, opened 2024-07-02
Feat: only login activated users

pieterlukasse:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/VADC-1242

### New Features
- feat: ensure only activated users can login

### Deployment changes
- depends on a new configuration setting: `AUTO_ACTIVATE_NEW_USER`

### DIFF
--- a/fence/auth.py
+++ b/fence/auth.py
@@ -87,8 +87,9 @@ def login_user(
         Args:
             user (User): User object
         """
-        # Abort login if user is not active:
-        if not user.active:
+        # Abort login if user.active is False (user.active is None or True are both
+        # considered active in this case):
+        if user.active is False:
             raise Unauthorized(
                 "User is known but not yet authorized/activated in the system"
             )
--- a/fence/auth.py
+++ b/fence/auth.py
@@ -87,6 +87,11 @@ def login_user(
         Args:
             user (User): User object
         """
+        # Abort login if user is not active:
+        if not user.active:
+            raise Unauthorized(
+                "User is known but not yet authorized/activated in the system"
+            )
         flask.session["username"] = user.username
         flask.session["user_id"] = str(user.id)
         flask.session["provider"] = user.identity_provider.name
@@ -112,7 +117,7 @@ def login_user(
             return
     else:
         # we need a new user
-        user = User(username=username)
+        user = User(username=username, active=config["AUTO_ACTIVATE_NEW_USER"])
 
         if email:
             user.email = email
--- a/fence/sync/sync_users.py
+++ b/fence/sync/sync_users.py
@@ -1143,6 +1143,9 @@ class UserSyncer(object):
             u.display_name = user_info[username].get("display_name", "")
             u.phone_number = user_info[username].get("phone_number", "")
             u.is_admin = user_info[username].get("admin", False)
+            # the presence of a user in the user sync input will activate the user
+            # by default, unless the user is explicitly being deactivated:
+            u.active = user_info[username].get("active", True)
 
             idp_name = user_info[username].get("idp_name", "")
             if idp_name and not u.identity_provider:
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -275,7 +275,7 @@
         "filename": "tests/conftest.py",
         "hashed_secret": "227dea087477346785aefd575f91dd13ab86c108",
         "is_verified": false,
-        "line_number": 1583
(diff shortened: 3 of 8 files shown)


fence/sync/sync_users.py
```diff
@@ -1143,6 +1143,9 @@ def _upsert_userinfo(self, sess, user_info):
             u.display_name = user_info[username].get("display_name", "")
             u.phone_number = user_info[username].get("phone_number", "")
             u.is_admin = user_info[username].get("admin", False)
+            # the presence of a user in the user sync input will activate the user
+            # by default, unless the user is explicitly being deactivated:
+            u.active = user_info[username].get("active", True)
```
Avantol13, line 1148: So at first I thought this didn't exist on the db table, but it turns out it does and it's just a legacy field. It's used elsewhere here: https://github.com/uc-cdis/fence/blob/d4182fd1c70bd6aa701d4fd9790b4c5db6b9773e/fence/resources/admin/admin_users.py#L125 which is part of the `admin` endpoints which have long since been deprecated/unused. So I think it's okay to repurpose this field, but I'd be concerned that some older databases may have this field populated in some cases. We should double check this, b/c if for some reasons some users in DCF or BDC (for example) have this set to False for whatever reason, we'd break login for them.



though I wish we could perhaps change the name, since I feel "active" is somewhat unintuitive for how it's being used here.

fence/auth.py
```diff
@@ -87,6 +87,12 @@ def set_flask_session_values(user):
         Args:
             user (User): User object
         """
+        # Abort login if user.active is False (user.active is None or True are both
+        # considered active in this case):
+        if user.active is False:
```
Avantol13, line 92: this does not feel like the right spot in the login flow for this. The function you're using is for setting session values in the web framework, I feel like this is somewhat obscuring what should be a higher-level denial of login. Plus, waiting until this point means a lot of stuff in the database gets updated/created regardless of their active status (including an entry in the user table): https://github.com/uc-cdis/fence/blob/d4182fd1c70bd6aa701d4fd9790b4c5db6b9773e/fence/auth.py#L102-L136

Maybe that's desired? But I would assume we want to _not_ update things like their email, idp, etc. unless they're `active`. As such, unless there's a compelling reason not to, I'd suggest reworking the logic to something where the denial happens sooner:

```python
    # line 107
    user = query_for_user(session=current_app.scoped_session(), username=username)
    if user:
        if user.active is False:
            # Abort login if user.active is False (user.active is None or True are both
            # considered active in this case):
            raise Unauthorized(
                "User is known but not yet authorized/activated in the system"
            )
        else:
            # existing logic
    else:
        if not config["AUTO_ACTIVATE_NEW_USER"]:
            # do not create new active users automatically 
            raise Unauthorized(
                "New user is not yet authorized/activated in the system"
            )
        else:
            # existing user creation logic
```

fence/auth.py
```diff
@@ -112,7 +118,7 @@ def set_flask_session_values(user):
             return
     else:
         # we need a new user
-        user = User(username=username)
+        user = User(username=username, active=config["AUTO_ACTIVATE_NEW_USER"])
```
Avantol13, line 121: I'm not sure we need to do this with the modified logic I'm suggesting above. It will probably confuse future people b/c only a subset of the users will have this set (since there's a ton of existing users where it'll remain `null`). 

I almost think it'd be better to just leave the `active` options primarily `null` (the default) and `false` an explicit denial. 

If we want to make the configuration more explicit, we could simply call it `ALLOW_NEW_USERS` and default to true

Avantol13: Another thought: for metrics and auditing, it's probably better _not_ to have disallowed users data in the database. e.g. if you want no new users and only the ones explicitly activated and synced, user counts would be off if we created rows in User for everyone that _tried_ to login and then got .active set to false (in the current logic).

Avantol13 (changes requested):

fence/auth.py
pieterlukasse: agreed. Makes sense to not hide it in the `set_flask_session_values`

fence/auth.py
pieterlukasse: good point. Not adding the user to the DB at all sounds good to me 👍

pieterlukasse:
closing in favor of https://github.com/uc-cdis/fence/pull/1184

PR #1161, opened 2024-07-18
Handle 'None' string in fence config 

AlbertSnows:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/PXP-11305

### Bug Fixes
- fixed initial setup error due to empty property in config-default.yaml file


fence/config.py
```diff
@@ -150,28 +152,37 @@ def post_process(self):
                 logger.warning(
                     f"IdP '{idp_id}' is using multifactor_auth_claim_info '{mfa_info['claim']}', which is neither AMR or ACR. Unable to determine if a user used MFA. Fence will continue and assume they have not used MFA."
                 )
+        dbgap_configs = self._configs["dbGaP"]
+        if isinstance(dbgap_configs, list):
+            corrected_dbgap_configs = dbgap_configs
+        else:  # assume is a dict
```
mfshao, line 158: Instead of assuming, why not actually check if `dbgap_configs` is a dict and raise an exception otherwise

AlbertSnows: Hmm, yeah I thought about that initially but decided to just minimize the changes. If it were ever not a dict though, that would be useful information. I'll update it!

Avantol13 (changes requested):

tests/test_app_config.py
```diff
@@ -138,3 +146,22 @@ def test_app_config_parent_child_study_mapping(monkeypatch):
...
+            "parent_to_child_studies_mapping": {
+                "phs001195": ["phs0015623"],
+                "phs001192": ["phs0001", "phs002"],
+            }
+        },
+    ]
+    try:
+        FenceConfig._validate_parent_child_studies(valid_dbgap_configs)
+    except Exception:
+        pytest.fail("Study validation failed when it should have passed!")
```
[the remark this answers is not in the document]
AlbertSnows: I don't know how this comment never showed up. I apologize for missing this, I'll get it implemented.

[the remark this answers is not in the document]
Avantol13: it was sitting on my side as a draft 🤦‍♂️, sorry about that. I just pushed it through this morning

BinamB (dismissed):

fence/config.py
```diff
@@ -150,28 +152,39 @@ def post_process(self):
...
-            ).keys()
-            conflicts = parent_studies & all_parent_studies
-            if len(conflicts) > 0:
-                raise Exception(
-                    f"{conflicts} are duplicate parent study ids found in parent_to_child_studies_mapping for "
-                    f"multiple dbGaP configurations."
-                )
-            all_parent_studies.update(parent_studies)
+    def _validate_parent_child_studies(dbgap_configs: List[Dict[str, Any]]) -> None:
+        def get_parent_studies_safely(dbgap_config):
```
Avantol13, line 174: I don't really like nesting function definitions inside another function unless you really need to. In this case I think you should break it out into another helper funcion or just embed the code

fence/config.py
```diff
@@ -150,28 +152,39 @@ def post_process(self):
...
+            corrected_dbgap_configs = dbgap_configs
+        elif isinstance(dbgap_configs, dict):
+            corrected_dbgap_configs = [dbgap_configs]
+        else:
+            raise Exception("The dbgap configuration is not a recognized data structure, aborting!")
+        self._validate_parent_child_studies(corrected_dbgap_configs)
 
-        self._validate_parent_child_studies(self._configs["dbGaP"])
+    @staticmethod
+    def find_duplicates(collection):
```
Avantol13, line 165: public functions require docstrings

fence/config.py
```diff
@@ -150,28 +152,39 @@ def post_process(self):
...
-            all_parent_studies.update(parent_studies)
+    def _validate_parent_child_studies(dbgap_configs: List[Dict[str, Any]]) -> None:
+        def get_parent_studies_safely(dbgap_config):
+            study_mapping = dbgap_config.get('parent_to_child_studies_mapping', {})
+            # study mapping could be 'None'
+            return list(study_mapping.keys()) if isinstance(study_mapping, dict) else []
+
+        safe_list_of_parent_studies = [safe_parent_studies
+                                       for dbgap_config in dbgap_configs
+                                       for safe_parent_studies in get_parent_studies_safely(dbgap_config)]
```
Avantol13, line 181: the two for loops in this list comprehension is imo too complicated for easy readability. I'd rather see this written out with one level of nesting if need be. List comprehension is nice when it's really clear what's happening, but I'd rather be explicit in more complicated cases like this and use traditional looping

tests/test_app_config.py
```diff
@@ -120,6 +120,15 @@ def test_app_config():
         patcher.stop()
 
 
+def test_validate_parent_child_studies_against_none_config():
```
Avantol13, line 123: tests require docstrings

tests/test_app_config.py
```diff
@@ -120,6 +120,15 @@ def test_app_config():
...
+    none_string_config = [{"parent_to_child_studies_mapping": 'None'},
+                          {"parent_to_child_studies_mapping": 'None'},]
+    try:
+        FenceConfig._validate_parent_child_studies(none_string_config)
+    except Exception:
+        pytest.fail("Study validation failed when given 'None' mapping!")
```
Avantol13, line 129: you can instead do:

```
with pytest.raises(Exception):
    FenceConfig._validate_parent_child_studies(none_string_config)
```

not a big deal but for consistency, the other assertions in the tests do it this way

fence/config.py
```diff
@@ -150,28 +152,39 @@ def post_process(self):
                 logger.warning(
                     f"IdP '{idp_id}' is using multifactor_auth_claim_info '{mfa_info['claim']}', which is neither AMR or ACR. Unable to determine if a user used MFA. Fence will continue and assume they have not used MFA."
                 )
+        dbgap_configs = self._configs["dbGaP"]
```
Avantol13, line 155: I actually liked this logic built into the validation so it doesn't clutter this outer scope of just loading config

Avantol13 (changes requested):

tests/test_app_config.py
AlbertSnows: docstring added

fence/config.py
AlbertSnows: I am not the biggest fan of list comprehensions either. However, speaking as someone who only learned the syntax recently, I have been able to get use to the syntax without much trouble. There are also other reasons I find compelling to use this over traditional for loops, but I will refrain from listing them all for sake of brevity. As such, I am going to keep the change as is. However, if you are not convinced or otherwise would like to insist let me know and I can make the switch.

fence/config.py
AlbertSnows: docstring added

tests/test_app_config.py
AlbertSnows: If I am understanding correctly, that line would cause the test to *pass* if an exception is thrown. This is not what we want. We currently treat 'None' as empty and thus ignore it. So this test should pass and not throw an exception. In which case, I think we just want to use a regular try catch. Let me know if I've misunderstood. I will leave this unresolved just in case.

fence/config.py
AlbertSnows: Yeah I'm sympathetic to that view. If you view this as the 'loading/processing' function and the validation function as just generically validating, then that would be the expected approach. The change was motivated by a separation of concern. The validation function is only cares about checking that there are no duplicates, and thus it would be appropriate to remove any logic that asks, "Are you giving me the correct data structure to validate?" 
However, there is an alternative approach. The compromise would be to generalize the validation function, call it V, and have V consist of two steps: checking that the dbgap config is structured correctly, and then check that there are no duplicates. 
I would have done that first, but I didn't want to be overly ambitious since this is just a bug fix. Let me know if you'd rather see that implementation instead though.

fence/config.py
AlbertSnows: function moved out

fence/config.py
Avantol13: Yeah, I would rather see that compromise. imo the validation isn't intended to take the "correct" data structure, it's intended to take the configuration as-is and determine if it's valid. So, to me it makes more sense to have it all encapsulated there

fence/config.py
Avantol13: I'm not convinced brevity > readability (and I know that readability is somewhat subjective). Unless that brevity actually results in a significant efficiency gain (and even that needs to be weighed against readability). It's rare to see list comprehensions with multiple for loops (even though you technically can do this) so even for people familiar with the basics, this requires a double take to trace what's actually happening. So I hold my original opinion that we should not do a double for loop in a list comprehension.

tests/test_app_config.py
Avantol13: Ah, yes, you're totally right, I misread this 👍

Avantol13 (changes requested):

fence/config.py
AlbertSnows: I think you may have misunderstood. Brevity was just referring to listing motivations for the change, it's not a property of the change itself per se. I don't think I see/share your issues with readability, but given the context I'll switch it over.

fence/config.py
AlbertSnows: makes sense to me, i'll implement the change

fence/config.py
```diff
@@ -150,28 +152,76 @@ def post_process(self):
                 logger.warning(
                     f"IdP '{idp_id}' is using multifactor_auth_claim_info '{mfa_info['claim']}', which is neither AMR or ACR. Unable to determine if a user used MFA. Fence will continue and assume they have not used MFA."
                 )
+        self._validate_dbgap_config(self._configs["dbGaP"])
 
-        self._validate_parent_child_studies(self._configs["dbGaP"])
+    @staticmethod
+    def find_duplicates(collection):
```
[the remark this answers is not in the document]
Avantol13: Yeah and actually, since these are static methods of this class, they're not really going to be used elsewhere in a meaningful way. I'd say you should probably preface these with `_` OR *not* have these in the class and then having them "public" is fine (either here or in utils)

fence/config.py
```diff
@@ -150,28 +152,76 @@ def post_process(self):
...
+            [k1, k2, ...]
+        """
+        study_mapping = dbgap_config.get('parent_to_child_studies_mapping', {})
+        # study mapping could be 'None'
+        safe_studies = list(study_mapping.keys()) if isinstance(study_mapping, dict) else []
+        return safe_studies
 
     @staticmethod
-    def _validate_parent_child_studies(dbgap_configs):
+    def validate_parent_child_studies(dbgap_configs: List[Dict[str, Any]]) -> None:
```
Avantol13, line 190: actually, I think keeping the _ makes sense. This won't be used elsewhere

fence/config.py
```diff
@@ -150,28 +152,76 @@ def post_process(self):
...
+            safe_list_of_parent_studies.extend(safe_parent_studies)
+
+        duplicates = FenceConfig.find_duplicates(safe_list_of_parent_studies)
+        if len(duplicates) > 0:
+            raise Exception(
+                f"{duplicates} are duplicate parent study ids found in parent_to_child_studies_mapping for "
+                f"multiple dbGaP configurations.")
+
+    @staticmethod
+    def enforce_is_array(dbgap_configs):
```
Avantol13, line 212: I think this should be _ prefaced. Not really used elsewhere

Avantol13 (changes requested):

fence/config.py
```diff
@@ -150,28 +152,76 @@ def post_process(self):
...
-    def _validate_parent_child_studies(dbgap_configs):
+    def validate_parent_child_studies(dbgap_configs: List[Dict[str, Any]]) -> None:
+        """
+        Given a list of dictionaries that contain a parent -> child study mapping,
+        this function will check that all parents are unique and not duplicated
+        Args:
+            dbgap_configs: [ { parent_to_child_studies_mapping: { k1: v1, ... } }, ... ]
+
+        Returns:
+            Nothing. This function will throw if duplicates are found.
```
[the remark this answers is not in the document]
AlbertSnows: changed to just a note

fence/config.py
AlbertSnows: added back _

fence/config.py
```diff
@@ -150,28 +152,76 @@ def post_process(self):
...
+        """
+        item_with_occurrences = Counter(collection)
+        duplicates = {item
+                      for item, occurrences in item_with_occurrences.items()
+                      if occurrences > 1}
+        return duplicates
+
+    @staticmethod
+    def get_parent_studies_safely(dbgap_config):
+        """
```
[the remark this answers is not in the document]
AlbertSnows: sentence added

fence/config.py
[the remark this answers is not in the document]
AlbertSnows: will dm clarifying questions

fence/config.py
AlbertSnows: will dm with a question

fence/config.py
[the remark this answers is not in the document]
AlbertSnows: resolved by adding todo's for future work and keeping them in config for now

fence/config.py
AlbertSnows: resolved by redesigning functions a bit, keeping them in config, and adding to todos

tests/test_utils.py
```diff
@@ -0,0 +1,36 @@
+import pytest
+
+from fence.config import FenceConfig
+
+
+def test_find_duplicates():
```
Avantol13: docstrings on all functions, including tests, even just a simple "Test the output of the duplicates utility function with expected input"

tests/test_utils.py
```diff
@@ -0,0 +1,36 @@
...
+    dup_case = FenceConfig._find_duplicates([1, 1, 2])
+    unique_case = FenceConfig._find_duplicates([1, 2, 3])
+    tuple_dup_case = FenceConfig._find_duplicates((1, 1, 1))
+    tuple_unique_case = FenceConfig._find_duplicates((1, 2, 3))
+    assert (dup_case == {1} and unique_case == set() and
+            tuple_dup_case == {1} and tuple_unique_case == set())
+
+
+def test__coerce_to_array():
```
Avantol13, line 15: the __ is a little odd. I know why you did it, but I'd say you keep what you did above and just call it `test_coerce_to_array`

fence/utils.py
```diff
@@ -1,3 +1,9 @@
+"""
+This file is for functions that are generalized enough to be used in varied places that need not be related.
```
Avantol13, line 2: thanks for adding this

fence/utils.py
```diff
@@ -57,97 +60,6 @@ def generate_client_credentials(confidential):
...
-            token_endpoint_auth_method=auth_method,
-            expires_in=expires_in,
-        )
-        s.add(client)
-        s.commit()
-
-    return client_id, client_secret
-
-
-def hash_secret(f):
```
Avantol13, line 126: was this not used anywhere?

fence/utils.py
```diff
@@ -428,6 +294,8 @@ def get_SQLAlchemyDriver(db_conn_url):
 
 
 # Default settings to control usage of backoff library.
+# TODO(fix-it): this variable should be moved to config.py to remove the reliance on the config
```
Avantol13, line 297: I'm nervous to add a bunch of TODO's for future work that isn't scheduled. I'd say either we expand the scope to include this or you really try to get it done as part of Fix it Friday or something, otherwise this will just stay like this for years

Avantol13 (changes requested):

tests/test_utils.py
AlbertSnows: I think i just missed these, should be added now!

tests/test_utils.py
AlbertSnows: can do

fence/utils.py
AlbertSnows: you are welcome!

fence/utils.py
AlbertSnows: correct, I did not find any uses of it so I removed it, but I can always add it back!

fence/utils.py
AlbertSnows: I will make a fix-it ticket for this, and try to get it done on friday.
as such, i will say let's keep the scope focused. i will mark this as resolved when ticket is made

AlbertSnows: https://ctds-planx.atlassian.net/browse/PXP-11392

fence/config.py
```diff
@@ -145,28 +147,114 @@ def post_process(self):
...
-            ).keys()
-            conflicts = parent_studies & all_parent_studies
-            if len(conflicts) > 0:
-                raise Exception(
-                    f"{conflicts} are duplicate parent study ids found in parent_to_child_studies_mapping for "
-                    f"multiple dbGaP configurations."
-                )
-            all_parent_studies.update(parent_studies)
+    def _some(predicate, iterable, found_nothing_value=None):
+        """ Returns the first element that satisfies the predicate, else fail value
```
BinamB, line 171:
```suggestion
        """
         Returns the first element that satisfies the predicate, else fail value
```
Small nitpick

fence/utils.py
```diff
@@ -57,97 +60,6 @@ def generate_client_credentials(confidential):
     return client_id, client_secret, hashed_secret
 
 
-def create_client(
```
BinamB, line 60: Why is create_client moved?

AlbertSnows: i think it was a circular dependency, let me check

AlbertSnows: couldn't recreate so I've moved it back.

BinamB (approved):

PR #1166, opened 2024-07-24
Fix/metrics

emalinowski:
### Bug Fixes
Fix the tests and single process metrics

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -38,9 +38,6 @@ from sqlalchemy.ext.compiler import compiles
 CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
 os.environ["FENCE_CONFIG_PATH"] = os.path.join(CURRENT_DIR, "test-fence-config.yaml")
 
-# Set the prometheus working directory *before* loading any fence app files or `prometheus_client`
-os.environ["PROMETHEUS_MULTIPROC_DIR"] = "/var/tmp/uwsgi_flask_metrics/"
-
 import prometheus_client
 
 import fence
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,7 +15,6 @@ import mock
 import uuid
 import random
 import string
-import tempfile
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -39,10 +38,6 @@
 CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
 os.environ["FENCE_CONFIG_PATH"] = os.path.join(CURRENT_DIR, "test-fence-config.yaml")
 
-# Set the prometheus working directory *before* loading any fence app files or `prometheus_client`
-PROMETHEUS_TMP_COUNTER_DIR = tempfile.TemporaryDirectory()
-os.environ["prometheus_multiproc_dir"] = PROMETHEUS_TMP_COUNTER_DIR.name
-
 import prometheus_client
```
paulineribeyre, line 41: looks like `import prometheus_client` can be moved back to line 34 then 😄

tests/test_metrics.py
```diff
@@ -730,11 +730,10 @@ def test_record_prometheus_events(prometheus_metrics_before, client):
...
     """
-    # NOTE: To update later. The metrics utils don't support this yet. The gauges are not handled correctly.
-    # resp = client.get("/metrics")
-    # assert resp.status_code == 200
-    # # no metrics have been recorded yet
-    # assert_prometheus_metrics(prometheus_metrics_before, resp.text, [])
+    resp = client.get("/metrics")
+    assert resp.status_code == 200
+    # no metrics have been recorded yet
+    assert_prometheus_metrics(prometheus_metrics_before, resp.text, [])
```
paulineribeyre, line 736: this doesn't actually do anything (because of the empty list param which means nothing is checked), could you revert and comment it again? to enable this test case we'll need to explicitly check that no metrics have been recorded. Which doesn't work right now because the test utils don't handle gauges correctly

paulineribeyre (approved):

PR #1167, opened 2024-07-26
chore(metrics): init Metrics in app init, cleanup setup code to one b…

Avantol13:
…ash script, begin migration of common code to library

⚠️ Work in progres

### Improvements
- Cleanup and move some of the new metrics logic using prometheus to the cdispyutils library

### DIFF
--- a/fence/__init__.py
+++ b/fence/__init__.py
@@ -27,7 +27,7 @@ config.load(
 )
 
 from fence.auth import logout, build_redirect_url
-from fence.metrics import metrics
+from fence.metrics import Metrics
 from fence.blueprints.data.indexd import S3IndexedFileLocation
 from fence.blueprints.login.utils import allowed_login_redirects, domain
 from fence.errors import UserError
@@ -97,6 +97,11 @@ def app_init(
     logger.info(
         f"Prometheus metrics are{'' if config['ENABLE_PROMETHEUS_METRICS'] else ' NOT'} enabled."
     )
+    if config["ENABLE_PROMETHEUS_METRICS"]:
+        # Initialize the Metrics instance
+        app.metrics = Metrics()
+    else:
+        app.metrics = Metrics(enabled=False)
 
 
 def app_sessions(app):
@@ -207,7 +212,7 @@ def app_register_blueprints(app):
         /!\ There is no authz control on this endpoint!
         In cloud-automation setups, access to this endpoint is blocked at the revproxy level.
         """
-        data, content_type = metrics.get_latest_metrics()
+        data, content_type = flask.current_app.metrics.get_latest_metrics()
         return flask.Response(data, content_type=content_type)
 
 
--- a/fence/metrics.py
+++ b/fence/metrics.py
@@ -7,12 +7,10 @@ for example).
 - The new method should call the `_increment_counter` and/or `_set_gauge` methods with the
 appropriate metric name and labels.
 - Call the new method from the code where relevant, for example:
-    from fence.metric import metrics
-    metrics.add_login_event(...)
+    from flask import current_app
+    current_app.metrics.add_login_event(...)
 - Add unit tests to the `tests/test_metrics` file.
 """
-
-
 import os
 import pathlib
 
@@ -26,13 +24,11 @@ from prometheus_client import (
     CONTENT_TYPE_LATEST,
 )
 
-from fence.config import config
-
 
 logger = get_logger(__name__)
 
 
-class Metrics:
+class Metrics(object):
     """
     Class to handle Prometheus metrics
     Attributes:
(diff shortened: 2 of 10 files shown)


.github/workflows/ci.yaml
```diff
@@ -1,8 +1,9 @@
+# push will run on every pushed commit to any branch (so this will rerun the tests
+# once a branch gets merged to main in addition to any new commits on any branch)
+on: push
+
 name: CI
-on:
-  push:
-  pull_request:
-    types: [opened, reopened]
```
paulineribeyre, line 5: There can be edge cases where the "automatic merge" messes something up, and the CI succeeds on the "branch run", but fails on the "PR run". In that case you have to update your branch manually and fix the conflict that git didn't see. It has happened to me a handful of times, not sure it's worth running the CI twice on every PR, but 🤷‍♀️

fence/__init__.py
```diff
@@ -97,6 +97,11 @@ def app_init(
     logger.info(
         f"Prometheus metrics are{'' if config['ENABLE_PROMETHEUS_METRICS'] else ' NOT'} enabled."
     )
+    if config["ENABLE_PROMETHEUS_METRICS"]:
+        # Initialize the Metrics instance
+        app.metrics = Metrics()
+    else:
+        app.metrics = Metrics(enabled=False)
```
paulineribeyre, line 104:
```suggestion
    # Initialize the Metrics instance
    app.metrics = Metrics(enabled=config["ENABLE_PROMETHEUS_METRICS"])
```

fence/metrics.py
```diff
@@ -26,37 +24,49 @@
...
+        """
+        Create a metrics class.
+
+        Args:
+            enabled (bool): If this is false, the class functions will be no-ops (no operations), effectively
+                            doing nothing. This is the behavior when metrics are disabled. Why? So application code
+                            doesn't have to check, it always tries to log a metric.
+        """
+        self.enabled = enabled
+        if enabled:
```
paulineribeyre, line 49: can we do `if not enabled: return` and reduce the indentation instead?

tests/ci_commands_script.sh
```diff
@@ -1,5 +1,20 @@
 #!/usr/bin/env bash
 
-mkdir -p /var/tmp/uwsgi_flask_metrics/ || true
-export PROMETHEUS_MULTIPROC_DIR="/var/tmp/uwsgi_flask_metrics/"
-poetry run pytest -vv --cov=fence --cov-report xml tests
+# Get the directory of the current script
+SCRIPT_DIR="$(dirname "$(realpath "$0")")"
+
+# assumes Fence repo folder structure
+RELATIVE_PATH="../deployment/scripts/metrics/setup_prometheus"
```
paulineribeyre, line 7: do you have to run this script now to run the tests, or does it work out of the box with just `pytest tests`?

tests/test_metrics.py
```diff
@@ -709,13 +709,25 @@ def test_login_log_push_to_sqs(
 ######################
 
 
-def test_disabled_prometheus_metrics(client, monkeypatch):
+@pytest.fixture(scope="function")
+def disable_metrics_app(app):
+    """
+    temporarily disable metrics on the session-scoped app for this function
```
paulineribeyre, line 715: i'm wondering if we should do the opposite... 🤔 set `ENABLE_PROMETHEUS_METRICS` to false in the tests by default and only enable metrics on the tests that need them?

tests/ci_commands_script.sh
Avantol13: I couldn't get it to work without requiring the bash script to set things ahead of time 😢 so you do have to use this now. pytest by itself will not effectively setup the environment for prometheus to be happy

tests/test_metrics.py
Avantol13: well, I think it's more likely that we'll have this on for our deployments, so I actually like that it's on for unrelated tests, making sure whatever we add for metrics doesn't accidentally break something existing

.github/workflows/ci.yaml
Avantol13: Does it happen in other repos? which automatic merge are you referring to?

paulineribeyre: it can happen in any repo

i can't find documentation about it but this is my understanding from what i observed:
when 2 checks run (for push and for PR) they're not necessarily running on the same code. The "push" check runs on the branch's code, and the "PR" check runs on the code that would be on master after merging the PR. If there are any changes on master that aren't on the branch, it performs a merge automatically and runs the CI on that. It's the same code that would end up on master if you choose to merge without updating your branch first - git/github finds no conflict and merges. Occasionally it does it wrong, and it can be caught by the "PR" CI check

tests/ci_commands_script.sh
paulineribeyre: in that case maybe we should add a line in the readme about how to run the tests - but i know there's another PR in flight that's refactoring the docs, so maybe we can add that there

.github/workflows/ci.yaml
Avantol13: We should be protected against people merging without first updating their branch, at which point the push should run again. I do remember issues in the past vaguely... but I can't find any documentation in GH Actions that leads me to believe it's necessary to have both of these. Unless you have a strong opinion, I would rather not run everything twice and just get feedback on the code on `main` when the branch-creator merges it in and pushes those commits.

fence/metrics.py
```diff
@@ -26,24 +28,41 @@
...
 
 
-class Metrics:
+class Metrics(object):
     """
     Class to handle Prometheus metrics
     Attributes:
-        registry (CollectorRegistry): Prometheus registry
+        _registry (CollectorRegistry): Prometheus registry
         metrics (dict): Dictionary to store Prometheus metrics
```
paulineribeyre, line 40: the attributes list is outdated, `metrics` is now `_metrics` and i added `_counter_descriptions` and `_gauge_descriptions` in the previous PR (and i did not update the docstring 🤪)

pyproject.toml
```diff
@@ -21,7 +21,7 @@ botocore = "*"
 cached_property = "^1.5.1"
 cdiserrors = "<2.0.0"
 cdislogging = "^1.0.0"
-cdispyutils = "^2.0.1"
+cdispyutils = {git = "https://github.com/uc-cdis/cdis-python-utils", rev = "feat/metrics"}
```
paulineribeyre, line 24: to update before merging

paulineribeyre (approved):

Dockerfile
```diff
@@ -41,7 +41,7 @@ RUN poetry config virtualenvs.create false \
 COPY . /$appname
 COPY ./deployment/uwsgi/uwsgi.ini /etc/uwsgi/uwsgi.ini
 COPY ./deployment/uwsgi/wsgi.py /$appname/wsgi.py
-COPY clear_prometheus_multiproc /$appname/clear_prometheus_multiproc
+COPY ./deployment/scripts/metrics/setup_prometheus /$appname/setup_prometheus
```
paulineribeyre, line 44: i think we can delete this file now https://github.com/uc-cdis/fence/blob/7b0aa60/clear_prometheus_multiproc

fence/metrics.py
```diff
@@ -26,24 +28,48 @@
...
+    def __init__(self, enabled=True, prometheus_dir="/var/tmp/prometheus_metrics"):
+        """
+        Create a metrics class.
+
+        Args:
+            enabled (bool): If this is false, the class functions will be no-ops (no operations), effectively
+                            doing nothing. This is the behavior when metrics are disabled. Why? So application code
+                            doesn't have to check, it always tries to log a metric.
+            prometheus_dir (str): Directory to use when setting PROMETHEUS_MULTIPROC_DIR env var (which prometheus requires
+                                  for multiprocess metrics collection). Note that this the prometheus client is very
```
paulineribeyre, line 59:
```suggestion
                                  for multiprocess metrics collection). Note that the prometheus client is very
```

paulineribeyre (approved):

PR #1172, opened 2024-08-07
Remove Foreign Key Constraints for Service Account

BinamB:
Pre-alembic era created a foreign key clent_id(from the client table) on the google_service_account table. This foreign key was then removed from the schema but any commons created before the constraint was removed still held the foreign key. The previous alembic migration tuncates the client table but this fails if the foreign key constraint still persists therefore failing the migration. This migration checks for the existence of the foreign key constraint and removes it if it exists. There is no downgrade path for this since not having the foreign key constraint is the correct schema throughout all versions. This migration is specifically for commons that were created before the foreign key constraint was removed.

### Bug Fixes
- Create new migration to remove foreign key constraint if it exists on google_service_account table
### Deployment changes
- Requires database migration


tests/migrations/test_e695e299f6a6.py
k-burt-uch, line 1: Do we have a way to test that the full alembic migration succeeds? IIRC the migration fails here: 
https://github.com/uc-cdis/fence/blob/master/migrations/versions/9b3a5a7145d7_authlib_update_1_2_1.py

Because of this FK constraint. But this only tests if the migration to remove the FK constraint succeeds, not that the collection of migration scripts succeeds.

I think a full migration check is beyond the scope of what you're doing here, but it might be best to move the removal of the FK to the previous migration https://github.com/uc-cdis/fence/blob/master/migrations/versions/9b3a5a7145d7_authlib_update_1_2_1.py here instead of adding a new migration.

migrations/versions/e695e299f6a6_remove_foreign_key_constraint_googleSA.py
k-burt-uch, line 1: Can't we update this migration
https://github.com/uc-cdis/fence/blob/master/migrations/versions/9b3a5a7145d7_authlib_update_1_2_1.py

To do this instead of making a new migration?

k-burt-uch (changes requested):

migrations/versions/e695e299f6a6_remove_foreign_key_constraint_googleSA.py
Avantol13: yeah, this won't work. We can't have a new migration b/c the previous one fails. We must update the previous one

Avantol13 (changes requested):

migrations/versions/9b3a5a7145d7_authlib_update_1_2_1.py
```diff
@@ -294,3 +298,33 @@ def set_old_column_values():
...
+
+    if fk_exists:
+        logger.info("Foreign key client_id exists. Removing constraint...")
+        op.drop_constraint(
+            "google_service_account_client_id_fkey",
+            "google_service_account",
+            type_="foreignkey",
+        )
+    else:
+        logger.info("Foreign key client_id does not exist.")
```
Avantol13, line 330:
```suggestion
        logger.debug("Foreign key client_id does not exist. This is expected for newer versions of the service.")
```

Avantol13 (dismissed):

k-burt-uch (approved):

PR #1175, opened 2024-08-16
Fix bulk_google_group_update for RAS Visa Sync

k-burt-uch:
Updates sync_single_user_visas to stop dropping existing google group membership. This preserves group access when a user with a visa logs in.

### Bug Fixes
- Updates sync_single_user_visas to stop dropping existing google group membership. This preserves group access when a user with a visa logs in.

### DIFF
--- a/fence/resources/google/access_utils.py
+++ b/fence/resources/google/access_utils.py
@@ -42,7 +42,7 @@ class GoogleUpdateException(Exception):
     pass
 
 
-def bulk_update_google_groups(google_bulk_mapping):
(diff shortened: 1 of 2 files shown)


fence/resources/google/access_utils.py
```diff
@@ -42,7 +42,7 @@ class GoogleUpdateException(Exception):
     pass
 
 
-def bulk_update_google_groups(google_bulk_mapping):
+def bulk_update_google_groups(google_bulk_mapping, keep_existing=False):
```
BinamB, line 45: Can you add `keep_existing` in the Args in the doc string?

PR #1177, opened 2024-08-26
fix(PXP-11385): Adds single user google group updates

k-burt-uch:
Fixes issue with passports where a single user google group updates would clear existing membership from the GBAG.

### Bug Fixes
- Fixes issue where google group updates for passport sync would clear existing membership from Google Bucket Access Groups.

### DIFF
--- a/fence/resources/google/access_utils.py
+++ b/fence/resources/google/access_utils.py
@@ -4,7 +4,6 @@ registration.
 """
 import backoff
 import time
-import flask
 from urllib.parse import unquote
 import traceback
 
@@ -17,7 +16,7 @@ import fence
 from cdislogging import get_logger
 
 from fence.config import config
-from fence.errors import NotFound, NotSupported, UserError
+from fence.errors import NotFound, NotSupported
 from fence.models import (
     User,
     Project,
@@ -29,7 +28,6 @@ from fence.models import (
 )
 from fence.resources.google.utils import (
     get_db_session,
-    get_users_from_google_members,
     get_monitoring_service_account_email,
     is_google_managed_service_account,
 )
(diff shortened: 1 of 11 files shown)


fence/resources/google/access_utils.py
```diff
@@ -42,7 +42,83 @@ class GoogleUpdateException(Exception):
...
+
+        if google_update_failures:
+            raise GoogleUpdateException(
+                f"FAILED TO UPDATE GOOGLE GROUPS FOR USER {user_email} (see previous errors)."
+            )
+
+
+def bulk_update_google_groups(google_bulk_mapping, keep_existing=False):
     """
     Update Google Groups based on mapping provided from Group -> Users.
```
Avantol13, line 123: lets add `keep_existing` to the docstring

fence/resources/google/access_utils.py
```diff
@@ -42,7 +42,83 @@ class GoogleUpdateException(Exception):
...
+                gcm.remove_member_from_group(user_email, group)
+            except Exception as exc:
+                logger.error(
+                    f"ERROR: FAILED TO REMOVE USER {user_email} FROM "
+                    f"GOOGLE GROUP {group}! This sync will continue to update other groups. Exc: "
+                    f"{traceback.format_exc()}"
+                )
+                google_update_failures = True
+
+        if google_update_failures:
```
Avantol13, line 115: Thanks for all this error handling and clear logging messages 💯

fence/sync/sync_users.py
```diff
@@ -993,14 +990,14 @@ def sync_to_storage_backend(self, user_project, user_info, sess, expires):
...
             user_project_lowercase,
             sess,
-            google_bulk_mapping=google_bulk_mapping,
+            google_bulk_mapping=google_group_user_mapping,
             expires=expires,
         )
 
         if config["GOOGLE_BULK_UPDATES"]:
-            self.logger.info("Doing bulk Google update...")
-            bulk_update_google_groups(google_bulk_mapping)
```
Avantol13, line 1002: So this is called elsewhere by usersync, in this function: https://github.com/uc-cdis/fence/blob/e2a43234785ca09eb05aa49771ede5561c477472/fence/sync/sync_users.py#L1627

and `bulk_update_google_groups` still gets called there: https://github.com/uc-cdis/fence/blob/e2a43234785ca09eb05aa49771ede5561c477472/fence/sync/sync_users.py#L934

which means it'll wipe out groups.

Actually, I don't see where `sync_to_storage_backend` is called at all

fence/resources/google/access_utils.py
k-burt-uch: Oh, I need to remove this entirely. The above method should be used in place of that.

Avantol13: okay that makes a lot more sense actually. There's another place where this is called (see my other comment)

Avantol13 (approved):

PR #1179, opened 2024-08-28
Fix/presigned url addressing style

mfshao:
### Bug Fixes
- Use `auto` addressing style for presigned url generation (attempts to use `virtual`, but falls back to `path` if necessary)
- Pass `endpoint_url` from Fence config to boto3 client for presigned url creation

### DIFF
--- a/fence/blueprints/data/multipart_upload.py
+++ b/fence/blueprints/data/multipart_upload.py
@@ -142,12 +142,18 @@ def generate_presigned_url_for_uploading_part(
         presigned_url(str)
     """
     try:
+        s3_buckets = get_value(
+            config, "S3_BUCKETS", InternalError("S3_BUCKETS not configured")
+        )
+        bucket = s3_buckets.get(bucket_name)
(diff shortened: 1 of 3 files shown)


fence/blueprints/data/multipart_upload.py
```diff
@@ -148,7 +148,7 @@ def generate_presigned_url_for_uploading_part(
             aws_secret_access_key=credentials["aws_secret_access_key"],
             aws_session_token=credentials.get("aws_session_token", None),
             region_name=region,
-            config=Config(s3={"addressing_style": "path"}, signature_version="s3v4"),
+            config=Config(signature_version="s3v4"),
```
paulineribeyre, line 151: no need to add `endpoint_url` here?

mfshao: I'm not sure. It appears to be not there before, and the clever safe tests in CI are for download only.
But I can try to add it later 🤔

paulineribeyre (approved):
lgtm

PR #1184, opened 2024-09-04
Feat: only allow login of onboarded users

pieterlukasse:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/VADC-1255

### New Features
- feat: new config option to allow only existing OR active users to login

### Deployment changes
- activation of the feature depends on a new configuration setting: `ALLOW_NEW_USER_ON_LOGIN`

### DIFF
--- a/fence/auth.py
+++ b/fence/auth.py
@@ -100,26 +100,37 @@ def login_user(
 
     user = query_for_user(session=current_app.scoped_session(), username=username)
     if user:
-        _update_users_email(user, email)
-        _update_users_id_from_idp(user, id_from_idp)
-        _update_users_last_auth(user)
+        if user.active is False:
+            # Abort login if user.active is False (user.active is None or True are both
+            # considered active in this case):
+            raise Unauthorized(
+                "User is known but not authorized/activated in the system"
+            )
+        else:
+            _update_users_email(user, email)
+            _update_users_id_from_idp(user, id_from_idp)
+            _update_users_last_auth(user)
 
-        #  This expression is relevant to those users who already have user and
-        #  idp info persisted to the database. We return early to avoid
-        #  unnecessarily re-saving that user and idp info.
-        if user.identity_provider and user.identity_provider.name == provider:
-            set_flask_session_values(user)
-            return
+            #  This expression is relevant to those users who already have user and
+            #  idp info persisted to the database. We return early to avoid
+            #  unnecessarily re-saving that user and idp info.
+            if user.identity_provider and user.identity_provider.name == provider:
+                set_flask_session_values(user)
+                return
     else:
-        # we need a new user
-        user = User(username=username)
+        if not config["ALLOW_NEW_USER_ON_LOGIN"]:
+            # do not create new active users automatically
+            raise Unauthorized("New user is not yet authorized/activated in the system")
+        else:
+            # add the new user
+            user = User(username=username)
 
-        if email:
-            user.email = email
+            if email:
+                user.email = email
 
-        if id_from_idp:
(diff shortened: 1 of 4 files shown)


fence/config.py
```diff
@@ -47,6 +47,7 @@ def post_process(self):
             "WHITE_LISTED_GOOGLE_PARENT_ORGS",
             "CLIENT_CREDENTIALS_ON_DOWNLOAD_ENABLED",
             "DATA_UPLOAD_BUCKET",
+            "ALLOW_NEW_USER_ON_LOGIN",
```
paulineribeyre, line 50: you shouldn't need this, `ALLOW_NEW_USER_ON_LOGIN` should already by True by default

fence/auth.py
```diff
@@ -100,26 +100,37 @@ def set_flask_session_values(user):
...
+        else:
+            # add the new user
+            user = User(username=username)
+
+            if email:
+                user.email = email
+
+            if id_from_idp:
+                user.id_from_idp = id_from_idp
+                # TODO: update iss_sub mapping table?
```
paulineribeyre, line 133: you could reduce the indentation and lower the number of lines changed in your PR by removing the 2 "else" statements, since raising an error will stop the execution
```
if not config["ALLOW_NEW_USER_ON_LOGIN"]:
    raise Unauthorized
user = User(username=username)
[...]
```

fence/config.py
pieterlukasse: I think it is needed for backwards compatibility purposes or if for whatever reason someone decides to not include a setting for this in their config. See also the lines that follow directly below the line you commented on:
```python
for default in defaults:
```

pieterlukasse: thanks. Further code inspection showed that it is indeed not needed 👍 
Should be resolved now.

tests/login/test_login_user.py
```diff
@@ -33,6 +36,22 @@ def test_login_user_already_in_db(db_session):
...
+     and logs in, the login returns an Unauthorized error.
+    """
+    email = "testuser@gmail.com"
+    provider = "Test Provider"
+    id_from_idp = "Provider_ID_0001"
+
+    test_user = User(username=email, is_admin=False, active=False)
+    db_session.add(test_user)
+    db_session.commit()
+    with pytest.raises(Unauthorized):
```
paulineribeyre, line 51: let's use `pytest.raises(Unauthorized, match="...")` to check that we get the right error message (out of the 2 error messages you added)

tests/login/test_login_user.py
```diff
@@ -85,6 +104,20 @@ def test_login_new_user(db_session):
...
+    """
+    Test that when ALLOW_NEW_USER_ON_LOGIN config is False,
+    and a user that is not in the database logs in, an
+    Unauthorized error is returned.
+    """
+    monkeypatch.setitem(config, "ALLOW_NEW_USER_ON_LOGIN", False)
+    email = "testuser@gmail.com"
+    provider = "Test Provider"
+    id_from_idp = "Provider_ID_0001"
+    with pytest.raises(Unauthorized):
```
paulineribeyre, line 117: let's check the err message here too

paulineribeyre (commented):
looks good, just a minor comment

tests/login/test_login_user.py
pieterlukasse: done

tests/login/test_login_user.py
pieterlukasse: done

paulineribeyre (approved):

PR #1188, opened 2024-09-26
Feature/OIDC oauth groups

flashguerdon:
### New Features

- Support for OIDC Group Claims
- Implementation of token refresh


tests/login/test_idp_oauth2.py
```diff
@@ -38,12 +45,6 @@ def test_has_mfa_claim_acr(oauth_client_acr):
     has_mfa = oauth_client_acr.has_mfa_claim({"acr": "mfa"})
     assert has_mfa
 
-
-def test_has_mfa_claim_acr(oauth_client_acr):
```
Avantol13, line 42: note to self / future reviewer: removed b/c duplicate

tests/login/test_idp_oauth2.py
```diff
@@ -83,3 +84,357 @@ def test_does_not_has_mfa_claim_amr(oauth_client_amr):
 def test_does_not_has_mfa_claim_multiple_amr(oauth_client_amr):
     has_mfa = oauth_client_amr.has_mfa_claim({"amr": ["pwd, trustme"]})
     assert not has_mfa
+
+@pytest.fixture
+def mock_app():
```
Avantol13, line 89: these could go in conftest

fence/resources/openid/idp_oauth2.py
```diff
@@ -33,13 +40,39 @@ def __init__(
         )
         self.idp = idp  # display name for use in logs and error messages
         self.HTTP_PROXY = HTTP_PROXY
+        self.check_groups = config.get("CHECK_GROUPS", False)
+        self.groups = self.settings.get("groups", None)
+        self.read_group_information = False
+        self.groups_from_idp = []
+        self.verify_aud = self.settings.get("verify_aud", False)
+        self.audience = self.settings.get("audience", self.settings.get("client_id"))
+        self.is_mfa_enabled = "multifactor_auth_claim_info" in self.settings
```
Avantol13, line 49: Do you need this? I'm not seeing why this is necessary outside of the existing logic

fence/blueprints/login/base.py
```diff
@@ -129,11 +138,107 @@ def get(self):
...
-        self.post_login(user=flask.g.user, token_result=result, id_from_idp=id_from_idp)
+
+        expires = self.extract_exp(refresh_token)
+
+        # if the access token is not a JWT, or does not carry exp, default to now + REFRESH_TOKEN_EXPIRES_IN
+        if expires is None:
+            expires = int(time.time()) + config["REFRESH_TOKEN_EXPIRES_IN"]
+
+        # # Store refresh token in db
+        if self.check_groups:
```
Avantol13, line 149: perhaps check if there's actually config for the group syncing for this client too, not just the global feature flag

fence/config-default.yaml
```diff
@@ -94,6 +94,7 @@ DB_MIGRATION_POSTGRES_LOCK_KEY: 100
 #   - WARNING: Be careful changing the *_ALLOWED_SCOPES as you can break basic
 #              and optional functionality
 # //////////////////////////////////////////////////////////////////////////////////////
+CHECK_GROUPS: false
```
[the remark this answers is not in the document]
Avantol13: Okay, giving this some more thought, maybe it makes sense to *not* have this global feature flag, and instead just have the per-client config. That could be a lot cleaner

tests/test-fence-config.yaml
```diff
@@ -141,6 +142,17 @@ OPENID_CONNECT:
...
       authorization_endpoint: 'https://generic2/authorization_endpoint'
+  generic3:
+    name: 'generic3'  # optional; display name for this IDP
+    client_id: ''
+    client_secret: ''
+    redirect_url: '{{BASE_URL}}/login/generic3/login'  # replace IDP name
+    # use `discovery` to configure IDPs that do not expose a discovery
+    # endpoint. One of `discovery_url` or `discovery` should be configured
+    discovery_url: 'http://localhost/realms/generic3/.well-known/openid-configuration'
+    groups:
```
Avantol13, line 153: The naming here makes this a little confusing in the code. Take a look at my previous comments around potential renaming. 

I think we need `is_enabled` true/false if we keep this under a `groups` config, but I feel like it would improve readability if there was a top-level flag turning this on and *then* you can read the config. So have :

```
generic3:
  is_authz_groups_sync_enabled: true
  authz_groups_sync:
    group_prefix: /covid
```

Avantol13 (changes requested):
Thanks folks! At a high-level this is a great direction.

Initial round of feedback complete. Light review of the testing at this point (since I expect some of it to change). Suggesting some reorganization and other minor cleanup (see comments).

Please feel free to respond to comments for clarification and/or make updates and let me know once everything has been addressed. I'll do another round of review after that.

fence/blueprints/login/base.py
```diff
@@ -93,6 +96,8 @@ def __init__(
             "OPENID_CONNECT"
         ].get(self.idp_name, {})
         self.app = app
+        self.check_groups = config.get("CHECK_GROUPS", False)
+        self.app = app if app is not None else flask.current_app
```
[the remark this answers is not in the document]
flashguerdon: The reason was flask.current_app wasn't available on some of the test. But I agree, it is redundant now.

fence/resources/openid/idp_oauth2.py
```diff
@@ -72,19 +108,39 @@ def get_jwt_keys(self, jwks_uri):
             return None
         return resp.json()["keys"]
 
+    def decode_token(self, token_id, keys):
```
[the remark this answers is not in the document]
flashguerdon: This function validates audience.

fence/resources/openid/idp_oauth2.py
```diff
@@ -33,13 +40,39 @@ def __init__(
...
         self.HTTP_PROXY = HTTP_PROXY
+        self.check_groups = config.get("CHECK_GROUPS", False)
+        self.groups = self.settings.get("groups", None)
+        self.read_group_information = False
+        self.groups_from_idp = []
+        self.verify_aud = self.settings.get("verify_aud", False)
+        self.audience = self.settings.get("audience", self.settings.get("client_id"))
+        self.is_mfa_enabled = "multifactor_auth_claim_info" in self.settings
+
+        self.arborist = ArboristClient(
```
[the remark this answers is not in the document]
flashguerdon: Requires further investigation, as the arborist client here in not in the app context; hence new client.
RuntimeError: Working outside of application context

fence/error_handler.py
```diff
@@ -27,6 +28,9 @@ def get_error_response(error):
         )
     )
 
+
+    #raise error
```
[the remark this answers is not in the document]
flashguerdon: Issue: Error messages are obfuscated, the line needs be uncommented when troubleshooting errors. Breaks tests if not commented out / removed. I have added a TODO so this can be looked at.

fence/resources/openid/idp_oauth2.py
```diff
@@ -33,13 +40,39 @@ def __init__(
         )
         self.idp = idp  # display name for use in logs and error messages
         self.HTTP_PROXY = HTTP_PROXY
+        self.check_groups = config.get("CHECK_GROUPS", False)
+        self.groups = self.settings.get("groups", None)
+        self.read_group_information = False
+        self.groups_from_idp = []
+        self.verify_aud = self.settings.get("verify_aud", False)
+        self.audience = self.settings.get("audience", self.settings.get("client_id"))
```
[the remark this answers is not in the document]
flashguerdon: This is in the base class, the settings are for a specific client, so inconsistent clients are covered

fence/resources/openid/idp_oauth2.py
flashguerdon: removed.

fence/config-default.yaml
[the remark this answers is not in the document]
flashguerdon: Makes sense. Removed the global feature flag

tests/test-fence-config.yaml
flashguerdon: agreed

fence/resources/openid/idp_oauth2.py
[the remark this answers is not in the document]
Avantol13: Are you ever running this class outside of the flask app? If it's happening in tests, you may need to inspect that. The arborist client for Fence gets instantiated here: https://github.com/uc-cdis/fence/blob/0efce95888718161e58359312e2f345e652db446/fence/__init__.py#L478-L480 it's based on a configuration, so you may need to add that to your instance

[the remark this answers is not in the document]
flashguerdon: The issue is with [fence/job/access_token_updater.py](https://github.com/uc-cdis/fence/pull/1188/files#diff-e96202d6bcb84a435f7a97b898fb222fc1cc15597fc58a92a7a812b57769a44a), when running `fence_create.py update-visas` script, get out of the app context via cron job when executing `update_user_authorization` in Oauth2ClientBase, which requires arborist client. Any ideas?

fence/job/access_token_updater.py
```diff
@@ -44,17 +44,33 @@ def __init__(
...
         else:
-            self.ras_client = RASClient(
+            ras_client = RASClient(
                 oidc["ras"],
                 HTTP_PROXY=config.get("HTTP_PROXY"),
                 logger=logger,
             )
+            self.oidc_clients_requiring_token_refresh.append(ras_client)
+
+        # Initialise a client for each OIDC client in oidc, which does has gis_authz_groups_sync_enabled set to true and add them
```
Avantol13, line 63:
```suggestion
        # Initialise a client for each OIDC client in oidc, which does has is_authz_groups_sync_enabled set to true and add them
```

fence/job/access_token_updater.py
```diff
@@ -44,17 +44,33 @@ def __init__(
...
+        # to oidc_clients_requiring_token_refresh
+        for oidc_name in oidc:
+            if oidc.get(oidc_name).get("is_authz_groups_sync_enabled", False):
```
Avantol13, line 66:
```suggestion
            if oidc.get(oidc_name, {}).get("is_authz_groups_sync_enabled", False):
```

fence/job/access_token_updater.py
```diff
@@ -160,30 +176,38 @@ async def updater(self, name, updater_queue, db_session):
...
+        Select OIDC client based on identity provider.
         """
+        self.logger.info(f"Selecting client for user {user.username}")
         client = None
-        if (
-            user.identity_provider
-            and getattr(user.identity_provider, "name") == self.ras_client.idp
-        ):
-            client = self.ras_client
+        for oidc_client in self.oidc_clients_requiring_token_refresh:
```
Avantol13, line 202: this could be slightly more efficient if `oidc_clients_requiring_token_refresh` was a dict of `{"oidc_client.idp": oidc_client}`

fence/resources/openid/idp_oauth2.py
```diff
@@ -161,17 +230,45 @@ def get_auth_info(self, code):
...
+                    group_prefix = self.settings.get("authz_groups_sync", {}).get(
+                        "group_prefix", ""
+                    )
+                except (AttributeError, TypeError) as e:
+                    self.logger(
+                        f"Error: is_authz_groups_sync_enabled is enabled, required values not configured: {e}"
+                    )
+                    raise Exception(e)
+                except KeyError as e:
+                    self.logger(
```
Avantol13, line 256: self.logger.error

fence/blueprints/login/base.py
```diff
@@ -129,11 +142,141 @@ def get(self):
...
-        self.post_login(user=flask.g.user, token_result=result, id_from_idp=id_from_idp)
+
+        expires = self.extract_exp(refresh_token)
+
+        # if the access token is not a JWT, or does not carry exp, default to now + REFRESH_TOKEN_EXPIRES_IN
+        if expires is None:
+            expires = int(time.time()) + config["REFRESH_TOKEN_EXPIRES_IN"]
+
+        # Store refresh token in db
+        if self.is_read_authz_groups_from_tokens_enabled:
```
Avantol13, line 153: This could also be done if we simply want to persist the refresh token but not use this read authz groups feature.

So I would propose 2 things:
- Ensure it is obvious in the configuration that enabling `is_read_authz_groups_from_tokens_enabled` also makes it so that Fence persists refresh tokens from that IdP
- Provide a _separate_ config, perhaps `persist_refresh_token`, which is defaulted to False and overrides `is_read_authz_groups_from_tokens_enabled`. This allows an idp who is NOT reading authz groups to request storage of refresh tokens. While this doesn't immediately benefit us here, it is a more flexible configuration for future extensibility

fence/blueprints/login/base.py
```diff
@@ -129,11 +142,141 @@ def get(self):
...
+        This function assumes that the refresh token is valid and does not perform any JWT validation.
+        For any JWT coming from an OpenID Connect (OIDC) provider, validation should be done using the
+        public keys provided by the IdP (from the JWKS endpoint) before using this function to extract
+        the expiration time (`exp`). Without validation, the token's integrity and authenticity cannot
+        be guaranteed, which may expose your system to security risks.
+
+        Ensure validation is handled prior to calling this function, especially in any public or
+        production-facing contexts.
+
+        Parameters:
```
Avantol13, line 192: we use "Google-style" docstrings, see other examples in Fence or reference the docs: https://google.github.io/styleguide/pyguide.html

fence/resources/openid/idp_oauth2.py
```diff
@@ -72,19 +100,60 @@ def get_jwt_keys(self, jwks_uri):
...
+            return decoded_token
+
+        except JWTClaimsError as e:
+            self.logger.error(f"Claim error: {e}")
+            raise JWTClaimsError(f"Invalid audience: {e}")
+        except JWTError as e:
+            self.logger.error(f"JWT error: {e}")
+            raise JWTError(f"JWT error occurred: {e}")
+
     def get_jwt_claims_identity(self, token_endpoint, jwks_endpoint, code):
```
Avantol13, line 143: you're modifying the original intent of this function, which was to return the validated claims from the identity (id) token. I understand why you're adding this functionality here (to avoid an additional API call to the token endpoint), but I think we need to update the function name and docstring and consider returning similar formats for the id and refresh token (e.g. one is validated and decoded and one is NOT). 

The fact that the returned refresh token here is not validated is problematic, b/c it's unclear to the caller that they're getting 1 validated token and one non-validated one. And since this is such a high-level and used function, we should rethink this. 

Ideally, this should validate and return the decoded contents. Or rather, perhaps we need to split the logic up here. A function to get the _non-validated, raw_ tokens, and another function to get the ID token claims by passing in the raw token. Then we don't have to make the return of this function so messy

Avantol13 (changes requested):

fence/resources/openid/idp_oauth2.py
```diff
@@ -187,14 +284,13 @@ def get_access_token(self, user, token_endpoint, db_session=None):
         """
         Get access_token using a refresh_token and store new refresh in upstream_refresh_token table.
         """
+        ###this function is not correct. use self.session.fetch_access_token, validate the token for audience and then return the validated token. Still store the refresh token. it will be needed for periodic re-fetching of information.
```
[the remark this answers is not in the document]
flashguerdon: Resolved.

michaelfitzo:
Hello! Before moving ahead with any final approvals, the engineering team feels that it would be extremely useful to us to hear a bit more background on the request.  We have recently implemented a requirement that PRs must typically include a Community Feature Document, which gives us a bit more background on the request, your requirements, and use cases.  

https://docs.google.com/document/d/1P2dfqnSH-e7OX1Hw62sDL8zcR7gZp4d152TlDBlomDc/edit?usp=sharing

The document looks long, but I would ask you to focus on the requirements and use cases sections along with whatever else you can.  Apologies again for not sending sooner, but I think this will unblock any final reviews required.  Let me know if you have any questions about this

PR #1189, opened 2024-10-07
Feat: add support for soft delete user

pieterlukasse:
Comment on breaking changes: see also changes in https://github.com/uc-cdis/fence/pull/1189/commits/bfd381f8a1f3c229ffdaf0274241a6f8fc11e15f . Based on my code review (and partly on internal discussion that can be found in the JIRA ticket), the admin part does not seem to be used. So impact is likely none.

Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/VADC-1357

### New Features
- new endpoint to soft-delete users in Fence.

### Improvements
- consolidate on use of `username` instead of a mix of `name` and `username` in the /admin/user endpoints.

### Breaking Changes
- some of the admin endpoints and methods now expect `username` instead of `name`.


fence/blueprints/admin.py
```diff
@@ -133,6 +133,20 @@ def delete_user(username):
     return response
 
 
+@blueprint.route("/users/<username>/soft", methods=["DELETE"])
```
pieterlukasse, line 136: username = name elsewhere... TODO - align variable names

fence/blueprints/admin.py
```diff
@@ -78,7 +78,7 @@ def create_user():
 
     Returns a json object
     """
-    username = request.get_json().get("name", None)
+    username = request.get_json().get("username", None)
```
paulineribeyre, line 81: Changing this in "create" and "update" is a breaking change, it's probably fine since it's not used, but let's state it under "breaking changes" in the PR description

fence/blueprints/admin.py
```diff
@@ -133,6 +135,20 @@ def delete_user(username):
     return response
 
 
+@blueprint.route("/users/<username>/soft", methods=["DELETE"])
+@blueprint.route("/user/<username>/soft", methods=["DELETE"])
```
paulineribeyre, line 139: The openapi doc should be updated to include all these admin endpoints: https://github.com/uc-cdis/fence/blob/master/openapis/swagger.yaml

fence/resources/admin/admin_users.py
```diff
@@ -359,6 +360,17 @@ def raise_unavailable(gpg_email):
         logger.info("Done with Google deletions.")
 
 
+def soft_delete_user(current_session, username):
+    """
+    Soft-remove the user by marking it as active=False.
+    """
+    logger.debug("Soft-delete user.")
```
paulineribeyre, line 367:
```suggestion
    logger.debug(f"Soft-delete user '{username}'")
```

tests/admin/test_admin_users.py
```diff
@@ -98,6 +99,34 @@ def test_delete_user(db_session, awg_users, cloud_manager):
...
+    """
+    Tests adm.soft_delete_user() by querying an existing User,
+    asserting it is not inactive, and then checking it became inactive
+    after it was soft-deleted.
+    """
+    username = "awg_user"
+    user = db_session.query(User).filter(User.username == username).first()
+    assert user != None
+    assert user.username == username
+    assert user.active is None
```
paulineribeyre, line 112: Why is `active` not True here? Shouldn't it be True by default? I find it strange that the endpoint would return anything else than True or False

tests/admin/test_admin_users_endpoints.py
```diff
@@ -685,6 +691,52 @@ def assert_google_proxy_group_data_deleted(db_session):
...
+    load_non_google_user_data,
+):
+    """
+    Test soft-delete user endpoint by checking that the result is an
+    deactivated user.
+    """
+    username = "test_user_d"
+    user = db_session.query(User).filter_by(username=username).one()
+    assert user.username == username
+    assert user.active is None
```
paulineribeyre, line 707: same question

tests/admin/test_admin_users.py
pieterlukasse: `None` is the default. See also https://github.com/uc-cdis/fence/pull/1184/files#diff-43b58b02e1a8813f615b44a626767d504b26ab4879207e121ade0cc3649550caR103-R108

Maybe I can formalize that here: https://github.com/uc-cdis/fence/pull/1189/files#diff-0874526385bbe4c898757d3eed899922a2f4f8e1c79155d997bdb63ef989866fR94

tests/admin/test_admin_users_endpoints.py
pieterlukasse: fixed.

tests/admin/test_admin_users.py
pieterlukasse: fixed the "same" next question comment below. But this one is harder, as None is just what is stored in the test DB. It is also what will be found in most `User` records in the wild, as None is the default value for this field. Shall I just add a comment next to the None? Or better to fix the test data?

paulineribeyre: I see a couple options
- When getting "active=None" from the DB, the code interprets it as "True" and replaces it with "True" (similar to you fix for the other comment)
- Write a DB migration to replace all "active=None" with "active=True", and make "active" non-nullable in the DB and in the User model

The 1st option is more of a quick fix / code smell in my opinion. The 2nd option would be preferable but is a bit more work (the DB migration should be quick to write, but the User model is in the "userdatamodel" library so it has to be updated separately).

fence/blueprints/admin.py
pieterlukasse: see 
https://petstore.swagger.io/?url=https://raw.githubusercontent.com/uc-cdis/fence/feat/add_support_for_soft_delete_user/openapis/swagger.yaml#/admin%2Fuser

pieterlukasse: I added 3 endpoints there to the "admin/user" section, focusing (for now) only on the ones we care about. Not sure if the other endpoints should be documented, or perhaps just removed from the code itself...

migrations/versions/3a5712474808_make_user_active_field_non_nullable.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def upgrade():
+    op.execute('UPDATE "User" SET active = True WHERE active IS NULL')
+    op.alter_column("User", "active", nullable=False, server_default="True")
+
+
+def downgrade():
+    op.alter_column("User", "active", nullable=True, server_default=None)
+    op.execute('UPDATE "User" SET active = NULL WHERE active = True')
```
paulineribeyre, line 26: I think the last line is unnecessary, the values can stay "true" when we downgrade

pieterlukasse: maybe...but I'd prefer to do a strict downgrade that restores it to exact same state it was before.

paulineribeyre: The problem is that it doesn't restore it to the same state it was in before - it assumes all "trues" were "null". If any `active` values were set to "true" before migrating, they will be changed to "null". We can't tell what the value was before the DB migration, so of these 2 non-ideal solutions (set to "null" or not), i would rather keep "true" which makes sense for this field, than keep "null" which doesn't

tests/admin/test_admin_users.py
pieterlukasse: implemented 2nd option.

migrations/versions/3a5712474808_make_user_active_field_non_nullable.py
pieterlukasse: See https://github.com/uc-cdis/fence/pull/1189/commits/d76c071b6d78809d246cde4ff3e1e22537184492

pieterlukasse:
> "breaking changes" in the PR description as mentioned in a previous comment, and update userdatamodel to the latest version

Done.

paulineribeyre (dismissed):

paulineribeyre (approved):

PR #1193, opened 2024-10-28
Fix/bucket name

mfshao:

### Bug Fixes
- Fixed a bug cause Fence to unable to generate presigned URL for S3 buckets in fence config that has wildcard characters in their names

### Deployment changes
- any env that has S3 buckets in fence config that has wildcard characters should consider adpot this release


Avantol13 (approved):
approved but suggesting a cleanup PR to split into two functions, `get_bucket_cfg` and `get_bucket_name` to avoid confusion in the future and remove the special handling

PR #1199, opened 2024-11-14
Automated copy feature/OIDC oauth groups

Avantol13:
new version: https://github.com/uc-cdis/fence/pull/1233

duplicated from here: https://github.com/uc-cdis/fence/pull/1188

### New Features

- Support for OIDC Group Claims
- Implementation of token refresh


flashguerdon:
@Avantol13 -- Will you be able to give me access to the detailed integration test report? I get 401 when I click on the link.

fence/job/access_token_updater.py
```diff
@@ -3,23 +3,26 @@
...
 from fence.models import User
 from fence.resources.openid.ras_oauth2 import RASOauth2Client as RASClient
+from fence.resources.openid.idp_oauth2 import Oauth2ClientBase as OIDCClient
 
 
 logger = get_logger(__name__, log_level="debug")
 
 
-class Visa_Token_Update(object):
+class AccessTokenUpdater(object):
```
Avantol13, line 17: I'm wondering now if this really captures what this class does. Perhaps we should just call it `TokenAndAuthUpdater`, since it's not just updating access tokens.

fence/blueprints/login/base.py
```diff
@@ -129,11 +156,157 @@ def get(self):
         id_from_idp = result.get(self.id_from_idp_field)
 
         resp = _login(username, self.idp_name, email=email, id_from_idp=id_from_idp)
-        self.post_login(user=flask.g.user, token_result=result, id_from_idp=id_from_idp)
+
+        expires = self.extract_exp(refresh_token)
+
+        # if the access token is not a JWT, or does not carry exp,
+        # default to now + REFRESH_TOKEN_EXPIRES_IN
+        if expires is None:
```
Avantol13, line 164: we should add a log here so it's clear in an audit trail that we assumed a specific expiration

fence/resources/openid/idp_oauth2.py
```diff
@@ -161,17 +266,40 @@ def get_auth_info(self, code):
...
+            claims, refresh_token = self.get_jwt_claims_identity(
+                token_endpoint, jwks_endpoint, code
+            )
+
+            groups = None
+            group_prefix = None
+
+            if self.read_authz_groups_from_tokens:
+                try:
+                    groups = claims.get("groups")
```
Avantol13, line 282: I still think the field name in the claims should be configurable per IdP.

fence/resources/openid/idp_oauth2.py
```diff
@@ -187,14 +315,15 @@ def get_access_token(self, user, token_endpoint, db_session=None):
         """
         Get access_token using a refresh_token and store new refresh in upstream_refresh_token table.
         """
+        # this function is not correct. use self.session.fetch_access_token,
```
Avantol13, line 318: can you expand on this or fix the underlying issue? I'm not sure I'm following

fence/resources/openid/idp_oauth2.py
```diff
@@ -274,3 +403,150 @@ def store_refresh_token(self, user, refresh_token, expires, db_session=None):
         current_db_session = db_session.object_session(upstream_refresh_token)
         current_db_session.add(upstream_refresh_token)
         db_session.commit()
+
+    def get_groups_from_token(self, decoded_id_token, group_prefix=""):
+        """Retrieve and format groups from the decoded token."""
```
Avantol13, line 408: need vars in the docstring

fence/resources/openid/idp_oauth2.py
```diff
@@ -274,3 +403,150 @@ def store_refresh_token(self, user, refresh_token, expires, db_session=None):
...
+        authz_groups_from_idp = decoded_id_token.get("groups", [])
```
Avantol13, line 409: See previous comment re: this being a configuration per IdP

fence/resources/openid/idp_oauth2.py
```diff
@@ -274,3 +403,150 @@ def store_refresh_token(self, user, refresh_token, expires, db_session=None):
...
+                # Add user to all matching groups from IDP
+                for arborist_group in arborist_groups:
+                    if arborist_group["name"] in idp_group_names:
+                        self.logger.info(
+                            f"Adding {user.username} to group: {arborist_group['name']}, sub: {user.id} exp: {exp}"
+                        )
+                        self.arborist.add_user_to_group(
+                            username=user.username,
+                            group_name=arborist_group["name"],
+                            expires_at=exp,
```
Avantol13, line 517: The expiration here is the `access_token` expiration, was that intentional? Usually those are very short lived, like on the order of minutes, not hours. I wonder if we want to make the expiration configurable per IdP (so I could set it to be 8 hours for example) up to the maximum allowed in the refresh_token (which could be like 30 days?)?

fence/resources/openid/idp_oauth2.py
```diff
@@ -274,3 +403,150 @@ def store_refresh_token(self, user, refresh_token, expires, db_session=None):
...
+                        )
+
+                # Remove user from groups in Arborist that they are not part of in IDP
```
Avantol13, line 520: Unfortunately this design will not work generally. IdP's are not the only source of user membership, so you can't remove users from groups that don't match simply by querying for all groups and diffing with what's in a token, b/c there are other, external sources of other group memberships (admin user.yamls, RAS Passports) that need to be able to maintain their own policies. This logic would wipe users from those groups.

Avantol13: Either we rely on policy expiration to cleanup things, or you need a smarter comparison by knowing the list of groups managed by and IdP

fence/resources/openid/idp_oauth2.py
flashguerdon: The function is not retrieving the expires_at from token_response. I'm thinking we should be storing the new expires retrieved from token_response.

fence/resources/openid/idp_oauth2.py
flashguerdon: yeah, setting per IdP will be

fence/error_handler.py
```diff
@@ -27,6 +28,11 @@ def get_error_response(error):
         )
     )
 
+    # TODO: Issue: Error messages are obfuscated, the line below needs be
```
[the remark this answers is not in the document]
flashguerdon: OK

PR #1204, opened 2024-11-22
Feat/al2 old dockerfile

nss10:


fence/jwt/keys.py
```diff
@@ -148,13 +148,26 @@ def from_directory(cls, keys_dir, naming_function=None):
         prv_filepath = os.path.join(keys_dir, "jwt_private_key.pem")
 
         if not os.path.isfile(pub_filepath):
-            raise EnvironmentError(
-                "missing public key file; expected file to exist: " + pub_filepath
-            )
+            # Generate public key from private key
+            with open(prv_filepath, "r") as f:
```
Avantol13, line 152: what are we doing this in the Python code during runtime? Can we just require that the public key be provided like before?

AlbertSnows (approved):
Approved assuming the pipeline issues are resolved. can you add a description to the PR so I or other reviewers have a better understanding of the purposes of these changes? I don't see any glaring issues, but I highly recommend getting more eyes just in case.

PR #1209, opened 2025-01-10
move backoff settings to config

AlbertSnows:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/PXP-11392

This PR moves config properties and functions out of utils into config. Utils should be low-level, and thus not reference other namespaces for sake of not inducing circular dependencies.

### Dependency updates
- moved properties and functions out of utils to config or models to resolve cyclic dependencies


piotrsenkow (approved):
Everything looks good to me, nice refactoring!

fence/config.py
```diff
@@ -170,3 +175,71 @@ def _validate_parent_child_studies(dbgap_configs):
...
+    from userdatamodel.driver import SQLAlchemyDriver
+
+    # override userdatamodel's `setup_db` function which creates tables
+    # and runs database migrations, because Alembic handles that now.
+    # TODO move userdatamodel code to Fence and remove dependencies to it
+    SQLAlchemyDriver.setup_db = lambda _: None
+    return SQLAlchemyDriver(db_conn_url)
+
+
+def send_email(from_email, to_emails, subject, text, smtp_domain):
```
paulineribeyre, line 202: It doesn't make sense to me to move `send_email` from `utils.py` to `config.py`, that function has nothing to do with the config. If it can't be in `utils` for some reason, you can move it to `fence/scripting/google_monitor.py` which is the only file that uses it

AlbertSnows: good call, changed

fence/scripting/google_monitor.py
```diff
@@ -514,7 +517,7 @@ def _send_emails_informing_service_account_removal(
                 for reason in removal_reasons:
                     content += "\n\t\t\t - {}".format(reason)
 
-    return utils.send_email(from_email, to_emails, subject, content, domain)
+    return fence.config.send_email(from_email, to_emails, subject, content, domain)
```
paulineribeyre, line 520: you'll need to update the calls as well

AlbertSnows: ah the config already has send_email options so they didn't get changed. Thanks for catching this!

AlbertSnows:
PR blocked until ongoing audit issue is resolved

PR #1221, opened 2025-01-28
Fix typo in ClientAuthenication log

k-burt-uch:

### Improvements
- Fix typo in ClientAuthenication log


tianj7 (approved):

k-burt-uch:
If the integration tests are actually failing due to this change then I will write ten thousand RAS TDDs as punishment.

PR #1230, opened 2025-02-25
(Fix): Usersync ccdecrypt output not parsing

BinamB:
While changing from mcrypt to ccrypt, we didn't add `stdout` for the result of `ccdecrypt`. So we would decrypt the file but not output the file to somewhere its readable. 

### Bug Fixes
- Fixed usersync decrypted result not parsing

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "fence"
-version = "11.0.1"
+version = "11.1.0"
 description = "Gen3 AuthN/AuthZ OIDC Service"
 authors = ["CTDS UChicago <cdis@uchicago.edu>"]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "fence"
-version = "11.0.0"
+version = "11.0.1"
```
nss10, line 3: I think the current release for fence is `11.0.2` which doesn't have AL2 base image, the new release we may make might be `11.1.0` since the change is a major. WDYT?

BinamB: that makes sense. I'll push the change

Avantol13 (approved):

PR #1234, opened 2025-03-15
Feat: record /admin/user endpoint calls in the logs

pieterlukasse:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/VADC-1450

### New Features
-  add request log support for /admin/ endpoints
- add request log decorator to /admin/ endpoints

### Dependency updates
- depends on https://github.com/uc-cdis/audit-service/pull/74


fence/resources/audit/utils.py
```diff
@@ -80,6 +80,17 @@ def create_audit_log_for_request(response):
                     request_url=request_url,
                     **audit_data,
                 )
+        elif endpoint.startswith("/admin/"):
+            request_url = _clean_authorization_request_url(
+                request_url
+            )  # TODO - more than this will be needed for POST requests...
+            flask.current_app.audit_service_client.create_admin_action_log(
```
paulineribeyre, line 87: The `create_admin_action_log` method doesn't exist: https://github.com/uc-cdis/fence/blob/dccf559/fence/resources/audit/client.py

This code should fail, I'm very surprised that the existing "admin" unit tests passed. How?!

fence/blueprints/admin.py
```diff
@@ -391,6 +409,7 @@ def add_project_to_groups(projectname):
 
 @blueprint.route("/projects/<projectname>/bucket/<bucketname>", methods=["POST"])
 @admin_login_required
+@enable_audit_logging
 def create_bucket_in_project(projectname, bucketname):
     """
     DEPRECATED: This endpoint is deprecated and will be removed in a future release.
```
paulineribeyre, line 415: why are we adding audit logging to deprecated endpoints?

fence/resources/audit/utils.py
```diff
@@ -80,6 +80,17 @@ def create_audit_log_for_request(response):
...
+                status_code=response.status_code,
+                request_url=request_url,
+                action="admin_action_" + method,
```
paulineribeyre, line 90: I don't think audit-logging the request method is useful at all. "admin_action_POST" isn't going to tell you what was changed in the system 🤔

paulineribeyre (commented):
See feedback in [this thread](https://cdis.slack.com/archives/C01A12E8NUR/p1742502120384229?thread_ts=1741894783.892329&cid=C01A12E8NUR)

fence/resources/audit/utils.py
```diff
@@ -88,11 +88,76 @@ def create_audit_log_for_request(response):
...
+            endpoint,
+            audit_data,
+            request_url,
+            status_code,
+        )
+
+    except Exception:
+        # TODO monitor this somehow
+        traceback.print_exc()
+        logger.error(f"!!! Unable to create audit log! Returning response anyway...")
```
paulineribeyre, line 120: i think we could remove the try/except block for this function. I think it's fair to expect this function to always succeed

fence/resources/audit/utils.py
pieterlukasse: code removed. Local logging only now.

fence/resources/audit/utils.py
pieterlukasse: code removed. Local logging only now.

fence/resources/audit/utils.py
pieterlukasse: code was simplified. Local logging only now.

fence/resources/audit/utils.py
```diff
@@ -88,11 +93,54 @@ def create_audit_log_for_request(response):
...
-    record audit logs. It should not be added to non-audited endpoints,
-    so that performance is not impacted.
+    This decorator should be added to any API endpoint for which we want to
+    push audit logs into the **audit-service**. The audit-service logs serve a very
+    specific use case where we need some of the audit log entries available via audit-service.
+    For most of the cases, local logs will suffice, and therefore one should consider using
+    the enable_request_logging decorator instead.
+
+    This decorator should also not be added to non-audited endpoints, so that performance is not impacted.
     The `create_audit_log_for_request_decorator` decorator is only added
```
paulineribeyre, line 144:
```suggestion
    This decorator should also not be added to non-audited endpoints, so that performance is not impacted.

    The `create_audit_log_for_request_decorator` decorator is only added
```

fence/resources/audit/utils.py
```diff
@@ -88,11 +93,54 @@ def create_audit_log_for_request(response):
     return response
 
 
+def create_log_for_request(request: Request):
+    """
+    Right before processing the request (see `enable_local_audit_logging` decorator),
```
paulineribeyre, line 98:
```suggestion
    Right before processing the request (see `enable_request_logging` decorator),
```

fence/resources/audit/utils.py
```diff
@@ -88,11 +93,54 @@ def create_audit_log_for_request(response):
...
+    record an audit log. The data we need to record the logs are stored in
+    `flask.g.audit_data` before reaching this code.
+    """
+    username = get_current_user().username
+    method = request.method
+    endpoint = request.path
+    audit_data = getattr(flask.g, "audit_data", {})
```
paulineribeyre, line 105: i don't think we need `audit_data` in this function

fence/resources/audit/utils.py
```diff
@@ -88,11 +93,54 @@ def create_audit_log_for_request(response):
...
+    request_url = endpoint
+    if request.query_string:
+        # could use `request.url` but we don't want the root URL
+        request_url += f"?{request.query_string.decode('utf-8')}"
+    request_url = _clean_authorization_request_url(request_url)
+    logger.info(
+        f"Request log: admin method called by user. Details: user=%s, method=%s, endpoint=%s, audit_data=%s, request_url=% ",
```
paulineribeyre, line 112:
```suggestion
        f"Incoming request: user=%s, method=%s, endpoint=%s, audit_data=%s, request_url=% ",
```
Since you made it generic, let's not assume it's an admin endpoint

tests/admin/test_admin_users_endpoints.py
```diff
@@ -37,9 +37,9 @@ def mock_arborist(mock_arborist_requests):
 
 @pytest.fixture
 def admin_user(db_session):
-    test_user = db_session.query(User).filter_by(username="admin_user").first()
+    test_user = db_session.query(User).filter_by(username="test").first()
```
paulineribeyre, line 40: can you explain why you needed to change the username?

tests/admin/test_admin_users_endpoints.py
```diff
@@ -693,6 +693,7 @@ def assert_google_proxy_group_data_deleted(db_session):
 
 def test_soft_delete_user_username(
     client,
+    admin_user,
```
paulineribeyre, line 696: why do we need to add `admin_user` to `test_soft_delete_user_username` and `test_soft_delete_user_user_not_found`?

tests/admin/test_admin_users_endpoints.py
pieterlukasse: yes, I was still going through that myself...I don't think we'll need this. Things started breaking because of the newly added `username = get_current_user().username`...but I don't think this `get_current_user()` call is correct and am now considering reading user from the token instead, since not all calls are necessarily in done in the context of a browser login session.

tests/admin/test_admin_users_endpoints.py
pieterlukasse: this is needed because the new line 
```
username = get_user_from_claims(claims).username
```
added in `fence/resources/audit/utils.py` triggers a query to the DB

tests/admin/test_admin_users_endpoints.py
pieterlukasse: Ok, I reverted this part...

fence/resources/audit/utils.py
pieterlukasse: Ok, I can remove.

pieterlukasse: done

tests/admin/test_admin_users_endpoints.py
```diff
@@ -223,11 +252,14 @@ def test_get_user_username_nonexistent(
...
-def test_get_user_username_noauth(client, db_session):
+def test_get_user_username_noauth(client, db_session, log_capture):
     """GET /users/<username>: [get_user] but without authorization"""
-    # This creates a "test" user, so don't remove db_session fixture
+    # This creates a "test" user, so don't remove db_session fixture (this will show that 401 is because of noauth and
+    # not because user doesn't exist in DB):
     r = client.get("/admin/users/test_a")
     assert r.status_code == 401
+    # also assert that the logs are not recorded if noauth:
+    assert len(log_capture) == 0
```
pieterlukasse, line 262: So, non-authenticated requests are not logged...

paulineribeyre: Would it still be the case if you switch the order of the decorators on the admin functions?
i.e.
```
@enable_request_logging
@admin_login_required
```
instead of:
```
@admin_login_required
@enable_request_logging
```
So the logging triggers before the auth check? (not sure if that's the case, just spitballing)

paulineribeyre (approved):
lgtm if the non-authenticated call thing is cleared by security

tests/admin/test_admin_users_endpoints.py
pieterlukasse: good question. I can try. I think it will fail, since the first lines are: 
```
    claims = validate_jwt()
    username = get_user_from_claims(claims).username
```

pieterlukasse: So, tested it now. The order doesn't matter. The call to `validate_jwt()` results in the `raise JWTError(e.message)` below:

```python
    if encoded_token is None:
        try:
            encoded_token = get_jwt_header()
        except Unauthorized as e:
            raise JWTError(e.message)
```

PR #1235, opened 2025-03-24
Add default arborist url

jawadqur:

### Improvements
- Set default arborist url in default configuration. 
-

### DIFF
--- a/fence/config-default.yaml
+++ b/fence/config-default.yaml
@@ -751,7 +751,7 @@ AZ_BLOB_CREDENTIALS:
 AZ_BLOB_CONTAINER_URL: 'https://myfakeblob.blob.core.windows.net/my-fake-container/'
 
 # url where authz microservice is running
-ARBORIST: null
+ARBORIST: 'http://arborist-service'
 
 # url where the audit-service is running
 AUDIT_SERVICE: 'http://audit-service'


k-burt-uch:
This is a reasonable default, but not without potential issue. At the very least, we should update "Deployment changes" so commons that are not running arborist are aware they'll need to explicitly set this to null in their fence-config.yaml secret. 

Otherwise, this line:
https://github.com/uc-cdis/fence/blob/dccf559c70fc87851c9a63daa89a1ab9cda439c6/fence/__init__.py#L480
will cause a lot of issues if arborist isn't reachable or deployed.

m0nhawk (approved):

fence/config-default.yaml
```diff
@@ -751,7 +751,7 @@ AZ_BLOB_CREDENTIALS:
 AZ_BLOB_CONTAINER_URL: 'https://myfakeblob.blob.core.windows.net/my-fake-container/'
 
 # url where authz microservice is running
-ARBORIST: null
+ARBORIST: 'http://arborist-service'
```
paulineribeyre, line 754: This is fine but it's a breaking chance, by default fence will assume arborist is used. Can you update the PR description

jawadqur: > This is fine but it's a breaking chance, by default fence will assume arborist is used. Can you update the PR description

Updated breaking changes section.

jawadqur:
> This is a reasonable default, but not without potential issue. At the very least, we should update "Deployment changes" so commons that are not running arborist are aware they'll need to explicitly set this to null in their fence-config.yaml secret.
> 
> Otherwise, this line:
> 
> https://github.com/uc-cdis/fence/blob/dccf559c70fc87851c9a63daa89a1ab9cda439c6/fence/__init__.py#L480
> 
> 
> will cause a lot of issues if arborist isn't reachable or deployed.

Updated deployment changes.

paulineribeyre (approved):

k-burt-uch (approved):

PR #1238, opened 2025-03-28
Replace AutoAddPolicy with RejectPolicy and load known hosts to prevent man in the middle attacks

piotrsenkow:
[pps-1936](https://ctds-planx.atlassian.net/browse/PPS-1936)
### Breaking Changes
Usersync SFTP now rejects unknown host keys. We switched from paramiko.AutoAddPolicy() to paramiko.RejectPolicy() and load keys from ~/.ssh/known_hosts. If that file does not contain the dbGaP/SFTP host key, the usersync dbGaP sync will fail (previously it auto-trusted and added the key)

### Deployment changes
- Provide and mount a known_hosts file in the usersync pod/container at /root/.ssh/known_hosts (the process runs as root).
- Helm: update to the gen3-helm change that adds the fence-ssh-known-hosts ConfigMap and mounts it into the usersync job. After upgrading the chart, populate that ConfigMap with the correct host keys.


k-burt-uch:
Do not merge this without updating the deployment changes and breaking changes sections.

k-burt-uch (approved):

PR #1239, opened 2025-04-04
feat(deny): add global username denial regex

Avantol13:

### New Features
- add global username denial regex to configuration


fence/auth.py
```diff
@@ -3,6 +3,7 @@
 from datetime import datetime
 from functools import wraps
 import urllib.request, urllib.parse, urllib.error
+import re
```
george42-ctds, line 6: Can this line go above the line for `urllib`?

PR #1243, opened 2025-04-16
switching back to quay.

cmlsn:
### Dependency updates
Updated to use Quay.io instead of ECR


frankliuao (approved):
Use quay for community convenience. Approved.

paulineribeyre (approved):

PR #1244, opened 2025-04-17
[BDC-488] Add helper to check for valid token

SpencerAxelrod:
https://ctds-planx.atlassian.net/browse/BDC-488
### Bug Fixes
add token check when enqueing user in visa_update_cronjob


k-burt-uch:
Tested in BDC preprod. I asked @AlbertSnows to log in so I could verify his visa is synced.

Logs look MUCH cleaner now.

PR #1250, opened 2025-05-05
Fix/registration digest method

k-burt-uch:

### Bug Fixes
- Fix user registration by overriding the default digest method used by itsdangerous for flask wtf


fence/__init__.py
```diff
@@ -1,3 +1,9 @@
+# Override the default_digest_method for Signer before flask and flask_wtf are loaded. 
+import hashlib
+from itsdangerous import Signer
+
+Signer.default_digest_method = hashlib.sha256 # Explicitly set to sha256 as the default will break in FIPS environments.
```
paulineribeyre:
```suggestion
Signer.default_digest_method = hashlib.sha256 # Explicitly set to sha256 as the default (sha1) will break in FIPS environments.
```

fence/__init__.py
```diff
@@ -1,3 +1,9 @@
+# Override the default_digest_method for Signer before flask and flask_wtf are loaded. 
+import hashlib
+from itsdangerous import Signer
+
+Signer.default_digest_method = hashlib.sha256 # Explicitly set to sha256 as the default will break in FIPS environments when flask_wtf attempts to process a user registration form.
```
paulineribeyre: if the issue is indeed related to https://github.com/pgadmin-org/pgadmin4/issues/7979 could you add that link

fence/__init__.py
k-burt-uch: Done.

fence/__init__.py
k-burt-uch: Yup, added it a few other links

paulineribeyre (approved):

PR #1251, opened 2025-05-13
Info log instead of debug log for missing bucket

paulineribeyre:

### Improvements
- When a bucket is missing from the `S3_BUCKET` configuration, print an info log instead of a debug log to facilitate understanding issues in production


george42-ctds (approved):
Looks good

k-burt-uch (approved):

PR #1260, opened 2025-05-23
PPS-2006 Generic IdP discovery support

paulineribeyre:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/PPS-2006


tests/login/test_login_blueprint.py
```diff
@@ -0,0 +1,134 @@
+from urllib.parse import urlencode
```
paulineribeyre, line 1: This is not actually a new file - I renamed and updated `tests/login/test_default_login.py`

fence/blueprints/login/shib.py
```diff
@@ -70,7 +74,12 @@ def get(self):
         entityID = flask.session.get("entityID")
 
         # if eppn not available or logging in through NIH
-        if not username or not entityID or entityID == "urn:mace:incommon:nih.gov":
+        if (
+            not username
+            or not entityID
+            or entityID == "urn:mace:incommon:nih.gov"
```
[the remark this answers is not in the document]
paulineribeyre: I don't think so, I updated it for completeness (we manually created a new, working NIH Shib IdP a while back and the name changed) but this file is actually only used in fenceshib. So my update won't even go live lol

fence/resources/user/__init__.py
```diff
@@ -100,8 +100,11 @@ def get_user_info(current_session, username):
         "message": "",
     }
 
-    if "fence_idp" in flask.session:
-        info["fence_idp"] = flask.session["fence_idp"]
+    if "upstream_idp" in flask.session:
+        info["upstream_idp"] = flask.session["upstream_idp"]
+        # [Backwards compatibility for Fence multi-tenant login / Shibboleth legacy configuration]
```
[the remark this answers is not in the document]
paulineribeyre: We need it as long as we use login.bionimbus for NIH/InCommon login. If and when we switch to Satosa (or something else) across the board, we can get rid of all the Shibboleth logic in Fence - not just these code blocks. Imo that should be its own separate PR when the time comes

pyproject.toml
```diff
@@ -22,6 +22,7 @@ cached_property = "^1.5.1"
 cdiserrors = "<2.0.0"
 cdislogging = "^1.0.0"
 cdispyutils = "^2.0.1"
+defusedxml = "<1"
```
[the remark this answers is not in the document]
paulineribeyre: This was Snyk's recommended "how to fix" 🤷‍♀️

tests/data/incommon_mdq_data_extract.xml
```diff
@@ -0,0 +1,388 @@
...
+    <OrganizationDisplayName xml:lang="en">Northern Michigan University</OrganizationDisplayName>
+    <OrganizationURL xml:lang="en">http://www.nmu.edu</OrganizationURL>
+  </Organization>
+  <ContactPerson contactType="technical">
+    <GivenName>Karl Mulder</GivenName>
+    <EmailAddress>mailto:kmulder@nmu.edu</EmailAddress>
+  </ContactPerson>
+  <ContactPerson contactType="technical">
+    <GivenName>John Marra</GivenName>
+    <EmailAddress>mailto:jmarra@nmu.edu</EmailAddress>
```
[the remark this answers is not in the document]
paulineribeyre: Oops, yes definitely, i'll remove these from our file. But yes it's from a public API exposed by InCommon http://mdq.incommon.org/entities/idps/all

fence/blueprints/login/__init__.py
```diff
@@ -216,19 +296,19 @@ def get_login_providers_info():
...
     try:
         all_provider_info = [
-            provider_info(login_details) for login_details in login_options
+            get_provider_info(login_details) for login_details in login_options
         ]
     except KeyError as e:
         raise InternalError("LOGIN_OPTIONS misconfigured: cannot find key {}".format(e))
 
-    # if several login_options are defined for this default IDP, will
+    # if several login_options are defined for this default IdP, will
```
Avantol13, line 304: I know this was here already, but we may want a warning log saying this is happening. We don't expect this to be common, right?

fence/blueprints/login/__init__.py
```diff
@@ -360,37 +440,115 @@ def default_login():
...
+    Discofeed and replaced by us manually). So we most likely could use the InCommon list and the generic `get_all_upstream_idps` function for Shibboleth instead of duplicating logic.
     """
-    url = config["OPENID_CONNECT"].get("fence", {}).get("shibboleth_discovery_url")
-    if not url:
+    if format == "shibboleth":
+        return get_all_shib_idps(discovery_url)
+    elif format == "mdq-xml":  # InCommon Metadata Query Protocol
+        all_idps = []
+        xml_data = fetch_url_data(discovery_url, "text")
+        for element in ElementTree.fromstring(xml_data).iter():
```
Avantol13, line 463: it's possible this fails to parse, right? Or does fromstring just return empty or something? Either way, we should probably try to catch this and raise an internal error with some more focused error log info (we can also log the traceback)

fence/blueprints/login/__init__.py
```diff
@@ -360,37 +440,115 @@ def default_login():
...
+                not element.tag.endswith("EntityDescriptor")
+                or "entityID" not in element.keys()
+            ):
+                continue
+            idp = element.get("entityID")
+
+            # get the IdP's display name from IDPSSODescriptor.Extensions.UIInfo.DisplayName
+            display_names = []
+            for idps_so_descriptor in element.findall(
+                "{urn:oasis:names:tc:SAML:2.0:metadata}IDPSSODescriptor"
```
Avantol13, line 474: are all these fields defined by InCommon and guaranteed for a version or something? I'm again wondering if we should put some of this in configuration in case these change. If we're confident they will stay the same until a huge refactor is needed, then I'm okay with them in here like this

fence/blueprints/login/__init__.py
paulineribeyre: It wouldn't be a stretch, with multi-tenant fence set up with orcid and incommon for example, the default would be whichever of the 2 is first in the config. We just happen to not have that problem because we usually default to google 🤷‍♀️ 
It's documented [here](https://github.com/uc-cdis/fence/blob/4c9a0a0e23e24a890325ecdca2c98d7b99ec9f79/fence/config-default.yaml#L403-L405) but i can add a warning log too

fence/blueprints/login/__init__.py
paulineribeyre: The info is a bit hard to find but this is what I have:

According to [this page](https://spaces.at.internet2.edu/display/federation/metadata-saml#:~:text=InCommon%20metadata%20is%20the%20instantiation,For%20more%20information%2C%20see:), "The SAML representation of InCommon metadata is defined in:"
- https://groups.oasis-open.org/higherlogic/ws/public/document?document_id=56785
- https://wiki.oasis-open.org/security/SAML2MetadataAttr
- https://wiki.oasis-open.org/security/SAML2MetadataUI

In the last link, it says "Version 1.0" "Final standardization occurred on 24 October 2019." So it looks pretty stable. I don't love the idea of bloating the configuration with these fields, and if the format changes, i'm guessing we would need to change more than just the field names. Maybe a compromise would be to specify in the config that the format is version 1.0?

Avantol13: eh, I don't even think that compromise is necessary b/c that's not something we'd automate logic on. That is, unless you can parse the version before checking all these fields and just bail if it's not 1.0. otherwise, maybe just adding the info in your comment to either a comment somewhere relevant in the config or in the code would make me feel better.

paulineribeyre: I couldn't find an endpoint that returns the version.
I'll change [this](https://github.com/uc-cdis/fence/blob/374660403e2913249a874d2e3f2ab500cde291e9/fence/blueprints/login/__init__.py#L460) to "xml-mdq-v1.0" so that we can at least add support for another version later if needed

fence/blueprints/login/__init__.py
paulineribeyre: Tested and add error handling

PR #1274, opened 2025-06-23
Add ip audit logging and cache call to audit service schema endpoint

SpencerAxelrod:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/PD-16

This PR adds support for audit schema versioning and caching, allowing audit logs to include new fields like IP address based on the current audit service schema versions

### New Features
- The login audit log now conditionally includes the user's IP information if the audit schema version for login is 2.0 or greater.
- Introduced AUDIT_SCHEMA_CACHE to cache audit service schema version/models, cache is checked and updated before creating presigned URL and login audit logs.
- Added _get_audit_schema and _set_schema_models_cache methods to AuditServiceClient to fetch and cache the audit schema from the audit-service’s /_schema endpoint.
    - Falls back to a default v1 schema if the endpoint returns a 404 (older audit-service versions)

### Improvements
- Added and adjusted tests for new features


tests/test_metrics.py
```diff
@@ -545,6 +548,14 @@ def test_login_log_login_endpoint(
         get_auth_info_patch.start()
 
     with audit_service_mocker as audit_service_requests:
+        AUDIT_SCHEMA_CACHE.clear()
+        AUDIT_SCHEMA_CACHE.set(
```
k-burt-uch, line 552: Can you write a test for when the request to the audit service for the schema fails with a 404?

k-burt-uch (changes requested):

k-burt-uch (changes requested):

tests/test_metrics.py
SpencerAxelrod: added tests above

k-burt-uch (approved):

PR #1275, opened 2025-06-23
Fix passport request schema in swagger docs

k-burt-uch:


george42-ctds (approved):
Looks good to me.

PR #1291, opened 2025-08-22
Track X-forwarded headers at login

jacob50231:
Link to JIRA ticket if there is one: https://ctds-planx.atlassian.net/browse/PD-36 https://ctds-planx.atlassian.net/browse/PD-37

### New Features
Track X-forwarded headers for CADR reporting enhancements

### DIFF
--- a/.github/workflows/integration_tests.yaml
+++ b/.github/workflows/integration_tests.yaml
@@ -9,10 +9,6 @@ jobs:
     with:
       SERVICE_TO_TEST: fence
     secrets:
-      CI_AWS_ACCESS_KEY_ID: ${{ secrets.CI_AWS_ACCESS_KEY_ID }}
-      CI_AWS_SECRET_ACCESS_KEY: ${{ secrets.CI_AWS_SECRET_ACCESS_KEY }}
-      JENKINS_API_TOKEN: ${{ secrets.JENKINS_API_TOKEN }}
-      QA_DASHBOARD_S3_PATH: ${{ secrets.QA_DASHBOARD_S3_PATH }}
       CI_TEST_ORCID_USERID: ${{ secrets.CI_TEST_ORCID_USERID }}
       CI_TEST_ORCID_PASSWORD: ${{ secrets.CI_TEST_ORCID_PASSWORD }}
       CI_TEST_RAS_EMAIL: ${{ secrets.CI_TEST_RAS_EMAIL }}


fence/blueprints/data/indexd.py
```diff
@@ -123,9 +123,15 @@ def get_signed_url_for_file(
...
             username, user = next(iter(users_from_passports.items()))
+            x_forwarded_headers = [
+                f"{header}:{value}"
+                for header, value in flask.request.headers
+                if "X-Forwarded" in header
+            ]
             flask.g.audit_data = {
                 "username": username,
                 "sub": user.id,
+                "additional_data": x_forwarded_headers,
```
Avantol13, line 134: this is just updating the flask cache, you need to also update the logic where this is sent to the audit service:

https://github.com/uc-cdis/fence/blob/518e4b40c3b40e8a34329f09901415a79e74155f/fence/resources/audit/client.py#L143-L173

And we need to do this in a backwards compatible way, or ensure we update both fence and audit at the same time and tell ppl not to update one without the other

fence/blueprints/data/indexd.py
```diff
@@ -134,6 +142,7 @@ def get_signed_url_for_file(
         flask.g.audit_data = {
             "username": auth_info["username"],
             "sub": auth_info["user_id"],
+            "additional_data": x_forwarded_headers,
```
k-burt-uch, line 145: You need to do this https://github.com/uc-cdis/fence/blob/master/fence/resources/audit/client.py#L286 to make sure its backwards compatible. 

See https://github.com/uc-cdis/audit-service/blob/master/src/audit/routes/system.py#L14-L23 for the audit schema version where additional_data was added.

tests/test_metrics.py
```diff
@@ -370,6 +371,97 @@ def test_presigned_url_log_acl(
...
+                "additional_data": [],
+            },
+        )
+
+
+@pytest.mark.parametrize(
+    "indexd_client_with_arborist", ["s3_and_gs_acl_no_authz"], indirect=True
+)
+@pytest.mark.parametrize("endpoint", ["download", "ga4gh-drs"])
+def test_presigned_url_log_x_forwarded_headers(
```
k-burt-uch, line 383: This should not be in test_metrics.py. Please move it to test_audit.py

k-burt-uch (approved):

tests/test_metrics.py
k-burt-uch: Disregard this comment - not sure what I was looking at but its clear from looking at the other tests in this file that this is the appropriate place. My mistake 😅

k-burt-uch (changes requested):

PR #1315, opened 2025-12-19
Changes to run fence in OpenShift

paulmurdoch19:


deployment/wsgi/gunicorn.conf.py
```diff
@@ -1,9 +1,15 @@
...
 bind = "0.0.0.0:8000"
 workers = 1
 preload_app = True
-user = "gen3"
-group = "gen3"
+# Set user/group to the current user's UID and GID, unless UID is 0 (root), then set to 'gen3'
+if getuid() == 0:
+    user = "gen3"
+    group = "gen3"
+else:
```
paulmurdoch19: OpenShift wont run this container with uid and gid == gen3 or root so if were not running the container as the root user then we cannot switch to gen3 so use the uid and gid set by the container

dockerrun.bash
```diff
@@ -14,5 +14,9 @@ if [ -f /fence/jwt-keys.tar ]; then
   )
 fi
 
-nginx
+if [ "${OPENSHIFT}" = "true" ]; then
```
paulmurdoch19: check if a Variable called OPENSHIFT is set to true, if so run the nginx execute that doesn't have the set cap added in the base image

michaelfitzo:
Closing this PR since openshift support has been added

PR #1335, opened 2026-03-06
Add idp logout for cognito

k-burt-uch:
### Improvements
- Add Cognito `/logout` endpoint


.pre-commit-config.yaml
```diff
@@ -3,6 +3,7 @@ repos:
     rev: 25.1.0
     hooks:
     -   id: black
+        language_version: python3.13
```
Avantol13: why are we overriding the default? it should use system python which should be fine? https://pre-commit.com/#overriding-language-version

BinamB: oh i dont remember why i did that. Let me revert that.

fence/auth.py
```diff
@@ -252,17 +253,40 @@ def logout(next_url, force_era_global_logout=False):
...
     if force_era_global_logout or provider == IdentityProvider.itrust:
         safe_url = urllib.parse.quote_plus(next_url)
         provider_logout = config["ITRUST_GLOBAL_LOGOUT"] + safe_url
     elif provider == IdentityProvider.fence:
         base = config["OPENID_CONNECT"]["fence"]["api_base_url"]
         provider_logout = base + "/logout?" + urllib.parse.urlencode({"next": next_url})
+    elif provider == "cognito":
+        idp_openid_connect = config["OPENID_CONNECT"]["cognito"]
+        well_known_url = idp_openid_connect["discovery_url"]
+        well_known_resp = requests.get(well_known_url)
```
Avantol13: this needs to be backoff-ed and retried, it's a network call that could time out / fail / etc

fence/auth.py
```diff
@@ -252,17 +253,40 @@ def logout(next_url, force_era_global_logout=False):
...
-        flask.redirect(provider_logout or urllib.parse.unquote(next_url))
-    )
+    try:
+        redirect_response = flask.make_response(
+            flask.redirect(provider_logout or urllib.parse.unquote(next_url))
+        )
+        redirect_response.headers["Access-Control-Allow-Origin"] = "*"
+        redirect_response.headers["Access-Control-Allow-Credentials"] = "true"
+    except Exception as e:
+        print(f"Error logging out: {e}")
```
Avantol13: don't print, use logging and likely we need to reraise this error? This seems fatal

fence/auth.py
```diff
@@ -252,17 +253,40 @@ def logout(next_url, force_era_global_logout=False):
...
+                        "logout_uri": next_url,  # NOTE: This needs to be set up in the cognito console for an allowed sign-out url
+                    }
+                )
+            )
 
     flask.session.clear()
-    redirect_response = flask.make_response(
-        flask.redirect(provider_logout or urllib.parse.unquote(next_url))
-    )
+    try:
```
Avantol13, line 331: idk why we weren't before, but we should backoff and retry on this..

Avantol13, line 331: idk why we weren't before, but we should backoff and retry on this..

Avantol13 (changes requested):

fence/auth.py
BinamB: this url would be going to the browser so it wouldn't make sense to redirect though right?

Avantol13:
we should use responses .raise_for_status instead of manually checking for 200 / ok. The early return should be an error case, not the positive case, at least that is generally the pattern I would promote. Only early return on errors, the final return should be the positive case.

When you log the error with logging now, there's still an implicit ignoring of the exception, we just continue execution. We need to either reraise if we can't recover or we need to be very explicit that we're covering up an exception and handling it.

Avantol13:
Also if we know the error type we should use that and not generally catch the generic "Exception"

Avantol13 (changes requested):
We also need unit/regression tests for this

BinamB:
@Avantol13 Added better error handling + unit tests

fence/auth.py
```diff
@@ -252,17 +269,60 @@ def logout(next_url, force_era_global_logout=False):
...
+    elif provider == "cognito":
+        idp_openid_connect = config["OPENID_CONNECT"]["cognito"]
+        try:
+            well_known_resp = get_openid_config_for_idp(idp_openid_connect)
+            well_known = well_known_resp.json()
+        except requests.exceptions.HTTPError as e:
+            logger.error(
+                f"Well-known endpoint returned an error status after multiple retries, Cognito Session not invalidated, Logging out of Gen3. Error: {e}"
+            )
+            well_known = None
```
Avantol13: just set this once before all the try excepts

tests/test_logout.py
```diff
@@ -93,3 +96,59 @@ def test_logout_fence(app, client, user_with_fence_provider, monkeypatch):
         parsed_url = urllib.parse.urlparse(r.location)
         result_redirect = urllib.parse.parse_qs(parsed_url.query).get("next")[0]
         assert result_redirect == redirect
+
+
+def test_logout_cognito(client, db_session):
```
Avantol13, line 101: docstrings required for all tests

fence/auth.py
```diff
@@ -252,17 +269,60 @@ def logout(next_url, force_era_global_logout=False):
...
 
     flask.session.clear()
-    redirect_response = flask.make_response(
-        flask.redirect(provider_logout or urllib.parse.unquote(next_url))
-    )
+    try:
+        redirect_response = flask.make_response(
+            flask.redirect(provider_logout or urllib.parse.unquote(next_url))
+        )
+        redirect_response.headers["Access-Control-Allow-Origin"] = "*"
```
Avantol13: We need someone from security to validate this is okay. We may need to have some allow list here.

fence/auth.py
Avantol13: yeah, you're right, I missed this was the redirect

fence/auth.py
BinamB: Actually, we might not need this. I was trying stuff out before i realized it was a front end issue. Gonna remove this and test.

fence/auth.py
```diff
@@ -252,17 +269,56 @@ def logout(next_url, force_era_global_logout=False):
...
     if force_era_global_logout or provider == IdentityProvider.itrust:
         safe_url = urllib.parse.quote_plus(next_url)
         provider_logout = config["ITRUST_GLOBAL_LOGOUT"] + safe_url
     elif provider == IdentityProvider.fence:
         base = config["OPENID_CONNECT"]["fence"]["api_base_url"]
         provider_logout = base + "/logout?" + urllib.parse.urlencode({"next": next_url})
+    elif provider == "cognito":
+        idp_openid_connect = config["OPENID_CONNECT"]["cognito"]
+        try:
+            well_known = None
```
Avantol13: it's generally best practice to only inlude the lines that could exception in the try block. So I would actually more well_known = known outside of this. and if the .json() is something you are not catching an exception for, move that outside too. But that one might make sense to stay - though arguably we should have a specific JSON decode error catch with maybe the invalid json in the log so we could debug

BinamB: added a json decode error. Hope the way i did it makes sense?

fence/auth.py
BinamB: confirmed it works without the headers

fence/auth.py
```diff
@@ -252,17 +270,60 @@ def logout(next_url, force_era_global_logout=False):
...
+        except JSONDecodeError as e:
+            logger.error(
+                f"Invalid JSON resonse from well-known, Cognito Session not invalidated, Logging out of Gen3. Error: {e}"
+            )
+        except Exception as e:
+            logger.error(
+                f"Error occured trying to get well-known, Cognito Session not invalidated, Logging out from Gen3. Error: {e}"
+            )
+        if well_known:
+            end_session_endpoint = well_known.get("end_session_endpoint")
```
Avantol13, line 307: we may want to check the domain in end_session_endpoint against an allowlist to be extra safe

fence/auth.py
```diff
@@ -252,17 +274,67 @@ def logout(next_url, force_era_global_logout=False):
...
+            if (
+                domain(end_session_endpoint) not in allowed_login_redirects()
+            ):  # NOTE: discovery url for cognito is different than the cognito api domain url. Check the domain for the APIs like end_session_endpoint or authorization_endpoint found in the well-know openid config
```
Avantol13: this comment is long enough that it should not be in-line, we should put it above the conditional

Avantol13 (approved):

PR #1340, opened 2026-04-02
Catch 404 when deleting user in arborist

k-burt-uch:

### Bug Fixes
- Usersync no longer fails when deleting a user that does not exist in arborist.


paulineribeyre (changes requested):
I think we should only ignore 404s, and raise any other type of error 😬

PR #1344, opened 2026-04-15
Feat/add bulk presigned url

jacob50231:


fence/blueprints/data/blueprint.py
```diff
@@ -347,6 +350,35 @@ def download_file(file_id):
...
+        raise UserError("wrong Content-Type; expected application/json")
+
+    if "guids" not in params:
+        raise UserError("missing required argument `guids`")
+
+    guids = params["guids"]
+    results = bulk_get_signed_url_for_file(guids)
+
+    return flask.jsonify(results)
+    # results = {}
```
k-burt-uch: Is this intended to be used? If not, please remove this commented out code.

fence/blueprints/data/indexd.py
```diff
@@ -182,6 +183,57 @@ def get_signed_url_for_file(
...
+
+    indexed_files = BulkIndexedFiles(file_ids)
+
+    default_expires_in = config.get("MAX_PRESIGNED_URL_TTL", 3600)
+    expires_in = get_valid_expiration_from_request(
+        max_limit=default_expires_in,
+        default=default_expires_in,
+    )
+
+    # prepare_presigned_url_audit_log(requested_protocol, indexed_file)
```
k-burt-uch: Uncomment - we need to ensure that presigned-urls are logged in this flow. I don't believe the indexed_files.get_signed_urls implementation will do this.

fence/blueprints/data/indexd.py
```diff
@@ -470,6 +522,303 @@ def generate_aws_presigned_url_for_part(
...
+
+        # BULK AUTHORIZATION: Collect all unique authz that need checking
+        authz_needing_check = set()
+        authz_to_file_ids = defaultdict(list)
+        for file_id in self.file_ids:
+            file_authz = self.index_document.get(file_id).get("authz")
+            if file_authz:
+                # Convert authz list to tuple for hashing
+                authz_key = (
+                    tuple(file_authz) if isinstance(file_authz, list) else file_authz
```
k-burt-uch: Do we really have cases where authz is either a list or a single entry? That's wild.

fence/blueprints/data/indexd.py
```diff
@@ -470,6 +522,303 @@ def generate_aws_presigned_url_for_part(
...
+        users = []
+        failed_file_ids: list[dict] = []
+
+        # BULK AUTHORIZATION: Collect all unique authz that need checking
+        authz_needing_check = set()
+        authz_to_file_ids = defaultdict(list)
+        for file_id in self.file_ids:
+            file_authz = self.index_document.get(file_id).get("authz")
+            if file_authz:
+                # Convert authz list to tuple for hashing
```
k-burt-uch: Will this work if there's more than 2 authz attributes for a given indexd record?

fence/blueprints/data/indexd.py
```diff
@@ -470,6 +522,303 @@ def generate_aws_presigned_url_for_part(
...
+                    authz_to_file_ids[authz_key].append(file_id)
+
+        # BULK AUTHORIZATION: Get authorization results for all authz at once
+        authz_to_authorized_username = {}
+        if authz_needing_check:
+            if users_from_passports:
+                # Get auth_mapping for each passport user (1 call per unique user)
+                for username in users_from_passports.keys():
+                    try:
+                        auth_mapping = flask.current_app.arborist.auth_mapping(username)
```
k-burt-uch: Do we sync the passport to arborist before this call?

fence/blueprints/data/indexd.py
```diff
@@ -470,6 +522,303 @@ def generate_aws_presigned_url_for_part(
...
+                try:
+                    token = get_jwt()
+                except Unauthorized:
+                    token = None
+
+                for authz_key in authz_needing_check:
+                    authz_list = (
+                        list(authz_key) if isinstance(authz_key, tuple) else authz_key
+                    )
+                    authorized = flask.current_app.arborist.auth_request(
```
k-burt-uch: Re-use the checks we're doing for the passport flow where we pull the mapping then check the permissions individually. Also, backlog a ticket on BDC to implement bulk checks in arborist and gen3authz.

fence/blueprints/data/indexd.py
```diff
@@ -470,6 +522,303 @@ def generate_aws_presigned_url_for_part(
...
+                        methods="read-storage",
+                        resources=authz_list,
+                    )
+                    if authorized:
+                        authz_to_authorized_username[authz_key] = None
+
+        # STEP 2: Process files using cached authorization results
+        for file_id in self.file_ids:
+            authorized_user = None
+            file_authz = self.index_document.get(file_id).get("authz")
```
k-burt-uch: Do we have to get this list of authz again? Can't we just reuse the authz list we got earlier?

fence/blueprints/data/indexd.py
```diff
@@ -470,6 +522,303 @@ def generate_aws_presigned_url_for_part(
...
+                    tuple(file_authz) if isinstance(file_authz, list) else file_authz
+                )
+
+                if authz_key not in self.auth_roles:
+                    # Check cached authorization result
+                    if authz_key in authz_to_authorized_username:
+                        # User is authorized
+                        username = authz_to_authorized_username[authz_key]
+                        authorized_user = (
+                            users_from_passports.get(username) if username else None
```
k-burt-uch: I don't think this is correct. What do we do for the non-passport case in the non-bulk case for presigned urls? If we're setting the authorized_user to none in that case, then sure but it looks odd to me :)

fence/blueprints/data/indexd.py
```diff
@@ -470,6 +522,303 @@ def generate_aws_presigned_url_for_part(
...
+                        failed_file_ids.append(
+                            {
+                                "error_code": error_code,
+                                "object_ids": [file_id],
+                            }
+                        )
+                        continue
+
+            signed_url, user = (
+                self._get_signed_urls(
```
k-burt-uch: You need to catch the not found errors and add them to failed_file_ids. You should also have a catch-all exception that will add file_ids to error_code: 500.

fence/blueprints/data/indexd.py
```diff
@@ -470,6 +522,303 @@ def generate_aws_presigned_url_for_part(
...
+                            users_from_passports.get(username) if username else None
+                        )
+                        self.auth_roles.append(authz_key)
+                    else:
+                        # User is not authorized
+                        error_code = 403 if users_from_passports else 401
+                        failed_file_ids.append(
+                            {
+                                "error_code": error_code,
+                                "object_ids": [file_id],
```
k-burt-uch: This is not correct, but close. failed_file_ids should really only have one error_code per type with a list of all object_ids that fit that error code. So you need one 403 error code object with a list of all file_ids that the user didn't have permission for.

k-burt-uch (changes requested):

fence/blueprints/data/indexd.py
jacob50231: In all honesty, I don't know. According to the blame, this was written by Michael Lukowski before you started working on it and I figured it was for a good reason so I kept it.

fence/blueprints/data/indexd.py
jacob50231: Yup tuples can be any length they're just immutable lists. Did add ordering though to make sure we didn't get the same set of authz as duplicate keys (i.e. (authz1,authz2) (authz2, authz1) being registered differently)

fence/blueprints/data/indexd.py
jacob50231: Good catch. Added that logic to bulk_get_signed_url_for_file.

fence/blueprints/data/indexd.py
jacob50231: authz_to_file_ids had the actual authz tuple as the key in Michael's code with the value as a list of file_id's that correspond to it. That means we would have to iterate through the dictionary in order to find each corresponding file_id and authz pair which would probably be slower than calling index_document.get again. If speed is a priority over memory size I can create another dictionary that's just each file_id as a key and the authz's repeated over and over again as the values. Just let me know, I wouldn't have to add more than another line of code

fence/blueprints/data/blueprint.py
```diff
@@ -347,6 +350,25 @@ def download_file(file_id):
     return flask.redirect(result["url"])
 
 
+@blueprint.route("/download/bulk", methods=["POST"])
+def download_bulk_files():
```
k-burt-uch, line 357: Update the swagger / openapi yamls to include this endpoint.

fence/blueprints/ga4gh.py
```diff
@@ -42,4 +51,144 @@ def get_ga4gh_signed_url(object_id, access_id):
...
+    for access_id, object_ids in access_to_object_ids.items():
+        try:
+            result = bulk_get_signed_url_for_file(
+                file_ids=object_ids, requested_protocol=access_id
+            )
+
+            # Process resolved URLs
+            urls = result.get("urls", [])
+            for url_response in urls:
+                if isinstance(url_response, dict) and "url" in url_response:
```
SpencerAxelrod, line 205: `bulk_get_signed_url_for_file` currently returns a list of strings in the `"urls"` key, eg: 
```
"urls": ["https://...amazonaws.com/...", "https://...amazonaws.com/..."],
```

So this check for `isinstance(url_response, dict)` will drop everything. It currently passes the `test_bulk_drs_access` ecause you are mocking the `bulk_get_signed_url_for_file` response to actaully e a dict.

```
mock_response = {
    "urls": [
        {"drs_object_id": "guid1", "url": "https://signed1"},   # ← dict
        {"drs_object_id": "guid2", "url": "https://signed2"},   # ← dict
    ],
    "failed_file_ids": [],
}

with patch(
    "fence.blueprints.ga4gh.bulk_get_signed_url_for_file",
    return_value=mock_response,
```

I think you can just fix `get_signed_urls` to be

```
signed_urls.append({"drs_object_id": file_id, "url": signed_url})
```

instead of 

```
signed_urls.append(signed_url)
```

Then the tests need to be updated to not mock the helper at the route level. At minimum we should have a test that mocks the storage layer (boto3/GCS) and runs through the real `BulkIndexedFiles` path so the string->dict shape mismatch can be caught.

fence/blueprints/ga4gh.py
```diff
@@ -42,4 +51,144 @@ def get_ga4gh_signed_url(object_id, access_id):
...
+
+    access_to_object_ids = bulk_request.map_access_to_object_ids()
+    resolved_urls = []
+    unresolved_by_code = defaultdict(list)
+
+    # Get URLs for each access protocol
+    for access_id, object_ids in access_to_object_ids.items():
+        try:
+            result = bulk_get_signed_url_for_file(
+                file_ids=object_ids, requested_protocol=access_id
```
SpencerAxelrod: we need to pass passports in here and make sure they are supported. 
`ga4gh_passports=bulk_request.passports`

And we should reject if a request tries to give both a bearer token and a passport. See how the single-object drs endpoint works as reference e.g.

`if request.headers.get("Authorization") and bulk_request.passports:
    raise Forbidden("Sending both a bearer token and passports is not allowed")`

SpencerAxelrod: Also unmock bulk_signed_url_for_file in tests and test both token and passports

fence/blueprints/ga4gh.py
```diff
@@ -42,4 +51,144 @@ def get_ga4gh_signed_url(object_id, access_id):
...
+    return jsonify(result)
+
+
+class BulkObjectAccessIds(BaseModel):
+    bulk_object_id: str
+    bulk_access_ids: list[str]
+
+
+class BulkObjectAccessRequest(BaseModel):
+    passports: Optional[list[str]]
```
SpencerAxelrod: this has to be `passports: Optional[list[str]] = None` or else it will still require the field I think. Write a test to that doesn't pass a `passports` key just to make sure.

fence/blueprints/data/indexd.py
```diff
@@ -470,6 +565,330 @@ def generate_aws_presigned_url_for_part(
...
+        return indexd_server.rstrip("/")
+
+    @cached_property
+    def index_document(self):
+        index_dict = {}
+        for file_id in self.file_ids:
+            indexd_server = config.get("INDEXD") or config["BASE_URL"] + "/index"
+            url = indexd_server + "/index/"
+            try:
+                res = requests.get(url + file_id)
```
SpencerAxelrod: I think the main idea was to use a single POST to indexd's `/bulk/documents` endpoint here

fence/blueprints/ga4gh.py
```diff
@@ -42,4 +51,144 @@ def get_ga4gh_signed_url(object_id, access_id):
...
+            logger.debug(f"UserError fetching signed URLs for {access_id}: {e}")
+            unresolved_by_code[403].extend(object_ids)
+        except (Unauthorized, Forbidden) as e:
+            # Authentication/authorization errors
+            logger.debug(f"Auth error fetching signed URLs for {access_id}: {e}")
+            error_code = 401 if isinstance(e, Unauthorized) else 403
+            unresolved_by_code[error_code].extend(object_ids)
+        except NotFound as e:
+            logger.debug(f"NotFound fetching signed URLs for {access_id}: {e}")
+            unresolved_by_code[404].extend(object_ids)
```
SpencerAxelrod, line 233: A request for `[valid_guid, missing_guid]` under the same access ID returns both as unresolved even though one could have been resolved.

use indexd's bulk documents endpoint and split returned/missing IDs into resolved and unresolved lists

fence/blueprints/data/indexd.py
```diff
@@ -182,6 +184,65 @@ def get_signed_url_for_file(
     return {"url": signed_url}
 
 
+@enable_audit_logging
```
SpencerAxelrod, line 190: the function that runs within this decorator ( /fence/resources/audit/utils.py -> `create_audit_log_for_request()`) only will send logs for certain requests. We need to add support for the new endpoints to actually sned the logs to audit service

jacob50231: Still looking into this and would like a bit of clarification on what it usually looks like in our codebase. Should have pushed changes for the rest of the comments

k-burt-uch: https://github.com/uc-cdis/fence/blob/master/fence/resources/audit/utils.py#L66-L95 This segment of code checks the path and makes decisions on auditing based on it, so you'll need to update the handler here to account for it.

k-burt-uch (changes requested):

fence/blueprints/data/indexd.py
```diff
@@ -470,6 +571,332 @@ def generate_aws_presigned_url_for_part(
...
+        for file_id in self.file_ids:
+            authorized_user = None
+            if file_id not in self.index_document:
+                failed_file_ids_map[404].append(file_id)
+                continue
+            file_authz = self.index_document.get(file_id).get("authz")
+
+            if file_authz:
+                authz_key = (
+                    tuple(file_authz) if isinstance(file_authz, list) else file_authz
```
SpencerAxelrod: I think we need `tuple(sorted(file_authz))` so it's the same as the line 644

fence/blueprints/data/indexd.py
```diff
@@ -182,6 +184,71 @@ def get_signed_url_for_file(
...
+
+    indexed_files = BulkIndexedFiles(file_ids)
+
+    default_expires_in = config.get("MAX_PRESIGNED_URL_TTL", 3600)
+    expires_in = get_valid_expiration_from_request(
+        max_limit=default_expires_in,
+        default=default_expires_in,
+    )
+
+    if ga4gh_passports and not config["GA4GH_PASSPORTS_TO_DRS_ENABLED"]:
```
SpencerAxelrod, line 277: we should move this check to the blueprint level before calling `bulk_get_signed_url_for_file`, ill put example there

fence/blueprints/ga4gh.py
```diff
@@ -42,4 +51,152 @@ def get_ga4gh_signed_url(object_id, access_id):
...
+        )
+    except ValidationError as e:
+        return jsonify(e.errors()), 400
+
+    if request.headers.get("Authorization") and bulk_request.passports:
+        raise UserError(
+            "You cannot supply both GA4GH passports and a token "
+            "in the Authorization header of a request."
+        )
+
```
SpencerAxelrod, line 170: put the passports enabled check here:

```python
if bulk_request.passports and not config["GA4GH_PASSPORTS_TO_DRS_ENABLED"]:
    raise NotSupported(
        "Using GA4GH Passports as a means of authentication and authorization "
        "is not supported by this instance of Gen3."
    )
```

fence/blueprints/data/indexd.py
```diff
@@ -182,6 +184,71 @@ def get_signed_url_for_file(
...
+
+    # Collect X-Forwarded headers
+    x_forwarded_headers = [
+        f"{header}:{value}" for header, value in flask.request.headers if "X-" in header
+    ]
+
+    auth_info = _get_auth_info_for_id_or_from_request(
+        sub_type=int, db_session=db_session
+    )
+    flask.g.audit_data = {
```
SpencerAxelrod: ok so I think there is a specific case here where if they authenticate with a Passport but no bearer token, _get_auth_info_for_id_or_from_request will set `"username": auth_info["username"]` for the audit log to be anonymous. So we should check for passports before here, and if they are using passports set the `flask.g.audit_data` `username` using `users_from_passports.items()`

fence/blueprints/ga4gh.py
jacob50231: Kyle and I decided to ignore passports as it is not implemented in the single file download

fence/blueprints/data/indexd.py
```diff
@@ -470,6 +595,334 @@ def generate_aws_presigned_url_for_part(
...
+            raise UnavailableError("Fail to reach id service to find data location")
+
+        if res.status_code == 200:
+            try:
+                docs = res.json()
+                if not isinstance(docs, list):
+                    raise ValueError("Response is not a list")
+
+                # Build dict keyed by did - only returns docs that were found
+                index_dict = {doc["did"]: doc for doc in docs}
```
SpencerAxelrod, line 666: the single object drs route can handle both prefixed and non-prefixed requests e.g. `GET /objects/guid123/access/s3` and `GET /objects/dg.4503/guid123/access/s3`

Right now trying to use bulk drs route without prefix doesn't work e.g.

```
POST /objects/access
{
  "bulk_object_access_ids": [
    { "bulk_object_id": "guid123", "bulk_access_ids": ["s3"] }
  ]
}
```

@k-burt-uch I think we need to make sure the indexd `/bulk/documents` endpoint has the same prefix resolution if we want to handle this, which is annoying

fence/blueprints/ga4gh.py
```diff
@@ -42,4 +58,157 @@ def get_ga4gh_signed_url(object_id, access_id):
         drs="True",
     )
 
-    return flask.jsonify(result)
+    return jsonify(result)
+
+
+class BulkObjectAccessIds(BaseModel):
```
Avantol13, line 31: I'm not sure I like these models mingled in with the blueprint code. Could we at least move them to the top of the file above the routes? Otherwise I'm open to having a separate module for ga4gh_data_models.py or somethings

fence/blueprints/data/indexd.py
```diff
@@ -470,6 +595,334 @@ def generate_aws_presigned_url_for_part(
...
+                    for resource in authz_list:
+                        if resource in auth_mapping:
+                            methods_list = auth_mapping[resource]
+
+                            if any(
+                                m.get("service") == "fence"
+                                and "read-storage" in m.get("method", "")
+                                for m in methods_list
+                            ):
+                                authz_to_authorized_username[authz_key] = None
```
tianj7: does the JWT contain no username information? I thought fence token should have this.

jacob50231: Yup the JWT contains username information, but in the JWT path I'm only caching the authorization result. Itt falls back to _get_auth_info_for_id_or_from_request() which reads the username/sub from the request token, so I stored None here instead of carrying the username through the cache. If you'd prefer consistency with the passport flow, I can populate the username from the JWT and store it here as well but it wouldn't add any functionality and that authz_to_authorized_username isn't used anywhere else or fed to audit service.

fence/blueprints/data/indexd.py
```diff
@@ -63,7 +67,7 @@
     "az": {"upload": "PUT", "download": "GET"},
 }
 
-SUPPORTED_PROTOCOLS = ["s3", "http", "ftp", "https", "gs", "az"]
+SUPPORTED_PROTOCOLS = ["s3", "http", "ftp", "https", "gs", "az", "vec"]
```
Avantol13: we don't need vec anymore, the production version of Gen3 Embeddings relies on https

fence/blueprints/data/indexd.py
```diff
@@ -185,6 +189,95 @@ def get_signed_url_for_file(
     return {"url": signed_url}
 
 
+@enable_audit_logging
+def bulk_get_signed_url_for_file(
+    file_ids,
```
Avantol13: can we use the name `guids`?

fence/blueprints/data/indexd.py
```diff
@@ -185,6 +189,95 @@ def get_signed_url_for_file(
...
+    requested_protocol=None,
+    db_session=None,
+    bucket=None,
+    drs="False",
```
Avantol13: why a string?

fence/blueprints/data/indexd.py
```diff
@@ -185,6 +189,95 @@ def get_signed_url_for_file(
...
+    ga4gh_passports=None,
+):
+    requested_protocol = requested_protocol or flask.request.args.get("protocol", None)
```
Avantol13: why the flexible input with default as the request? I would prefer we extract the request info as "high level" as possible and then pass it in here

fence/blueprints/data/indexd.py
```diff
@@ -185,6 +189,95 @@ def get_signed_url_for_file(
...
+    r_pays_project = flask.request.args.get("userProject", None)
+    db_session = db_session or current_app.scoped_session()
+
+    # default to signing the URL
+    force_signed_url = True
+    no_force_sign_param = flask.request.args.get("no_force_sign")
```
Avantol13: this pattern of checking the request in the non-top level blueprint route should be avoided. Put all the arg parsing in `download_bulk_files` and use this as a lower-level, not-request-aware helper function. Then you don't need all this defaulting and checking request and setting a new var. Just have the parsing outside this and pass it in

fence/blueprints/data/indexd.py
```diff
@@ -185,6 +189,95 @@ def get_signed_url_for_file(
...
+    if no_force_sign_param and no_force_sign_param.lower() == "true":
+        force_signed_url = False
+
+    # Collect X-Forwarded headers
+    x_forwarded_headers = [
```
Avantol13: ooh we should not be doing this multiple times for diff endpoints. Don't we have this for other signed URLs already? Can you make a utility function instead?

fence/blueprints/data/indexd.py
Avantol13: agreed, this should fail fast

fence/blueprints/data/indexd.py
Avantol13: need a docstring

fence/blueprints/data/indexd.py
```diff
@@ -185,6 +189,95 @@ def get_signed_url_for_file(
...
+    prepare_bulk_presigned_url_audit_log(requested_protocol, indexed_files)
+    signed_urls, users, failed_file_ids = indexed_files.get_signed_urls(
+        requested_protocol,
+        expires_in,
+        force_signed_url=force_signed_url,
+        r_pays_project=r_pays_project,
+        users_from_passports=users_from_passports,
+    )
+
+    return {"urls": signed_urls, "failed_file_ids": failed_file_ids}
```
Avantol13: can we rename across the board to `guids` for this new code?

fence/blueprints/data/indexd.py
```diff
@@ -473,6 +600,334 @@ def generate_aws_presigned_url_for_part(
...
+            raise UnavailableError(res.text)
+
+    def get_signed_urls(
+        self,
+        protocol,
+        expires_in,
+        force_signed_url=True,
+        r_pays_project=None,
+        users_from_passports=None,
+    ):
```
Avantol13, line 681: docstring required for all public functions

fence/blueprints/data/indexd.py
```diff
@@ -473,6 +600,334 @@ def generate_aws_presigned_url_for_part(
...
+                elif isinstance(file_authz, str):
+                    authz_key = (file_authz,)
+                else:
+                    logger.error(f"Unexpected authz format for {file_id}: {file_authz}")
+                    continue
+                if authz_key not in self.auth_roles:
+                    authz_needing_check.add(authz_key)
+                    authz_to_file_ids[authz_key].append(file_id)
+
+        # BULK AUTHORIZATION: Get authorization results for all authz at once
```
Avantol13: all these comments creating logical blocks of code are smells that should lead to creating actual helper functions to break up this high-level logic into a series of calls - it would make this flow much more readable

fence/blueprints/data/indexd.py
```diff
@@ -473,6 +600,334 @@ def generate_aws_presigned_url_for_part(
...
+                        # Check which authz this user has access to
+                        for authz_key in authz_needing_check:
+                            if authz_key not in authz_to_authorized_username:
+                                authz_list = (
+                                    list(authz_key)
+                                    if isinstance(authz_key, tuple)
+                                    else authz_key
+                                )
+                                # Check if user has read-storage on any of the resources in this authz
+                                for resource in authz_list:
```
Avantol13: this nesting is brutal for readability, we need to either invert logic and fail early/skip or we need helper functions

fence/blueprints/data/indexd.py
```diff
@@ -473,6 +600,334 @@ def generate_aws_presigned_url_for_part(
...
+                                            for m in methods_list
+                                        ):
+                                            authz_to_authorized_username[authz_key] = (
+                                                username
+                                            )
+                                            break
+                    except Exception as e:
+                        logger.error(f"Failed to get auth mapping for {username}: {e}")
+            else:
+                # JWT-based authorization: reuse auth_mapping like passport flow
```
Avantol13: see previous comments, this could be another function

fence/blueprints/data/indexd.py
```diff
@@ -473,6 +600,334 @@ def generate_aws_presigned_url_for_part(
...
+
+        failed_file_ids = [
+            {
+                "error_code": error_code,
+                "object_ids": object_ids,
+            }
+            for error_code, object_ids in failed_file_ids_map.items()
+        ]
+
+        return signed_urls, users, failed_file_ids
```
Avantol13: I don't really like returning more than two things without creating a class or dataclass or something, especially if we're not using type hinting. Can you create a dataclass for this stuff?

fence/blueprints/data/indexd.py
```diff
@@ -473,6 +600,334 @@ def generate_aws_presigned_url_for_part(
...
+
+    def _get_signed_urls(
+        self,
+        protocol,
+        file_id,
+        expires_in,
+        force_signed_url,
+        r_pays_project,
+        authorized_user=None,
+    ):
```
Avantol13, line 755: docstring

fence/blueprints/ga4gh.py
```diff
@@ -42,4 +107,115 @@ def get_ga4gh_signed_url(object_id, access_id):
...
+        raise NotSupported(
+            "Using GA4GH Passports as a means of authentication and authorization "
+            "is not supported by this instance of Gen3."
+        )
+
+    # Validate bulk request size against maxBulk config
+    total_requested_count = sum(
+        len(ids) for ids in bulk_request.map_access_to_object_ids().values()
+    )
+    max_bulk_requests = config.get("MAX_BULK_DRS_REQUESTS", 100)
```
Avantol13: don't default here with a magic number. Config guarantees that this exists so rely on the actual default in the YAML and just config["MAX...

fence/blueprints/ga4gh.py
```diff
@@ -42,4 +107,115 @@ def get_ga4gh_signed_url(object_id, access_id):
...
+        "summary": {
+            "requested": total_requested_count,
+            "resolved": total_resolved,
+            "unresolved": total_unresolved,
+        },
+        "unresolved_drs_objects": unresolved_drs_objects,
+        "resolved_drs_object_access_urls": resolved_urls,
+    }
+
+    return jsonify(response), 200
```
Avantol13, line 268: is this really supposed to 200 if there are no resolved URLs?

fence/blueprints/data/indexd.py
```diff
@@ -2218,3 +2167,17 @@ def verify_data_upload_bucket_configuration(bucket):
             f"Buckets configgured in ALLOWED_DATA_UPLOAD_BUCKETS: {s3_buckets}"
         )
         raise Forbidden(f"Uploading to bucket '{bucket}' is not allowed")
+
+
+def collect_x_forwarded_headers(headers):
```
Avantol13, line 2275: docstrings in all public functions please

fence/blueprints/data/bulk_helpers.py
```diff
@@ -0,0 +1,260 @@
...
+                file_id,
+                expires_in,
+                force_signed_url,
+                r_pays_project,
+                authorized_user,
+            )
+            signed_urls.append({"drs_object_id": file_id, "url": signed_url})
+            users.append(authorized_user)
+        except Exception as exc:
+            from fence.errors import NotFound
```
Avantol13: import at top of file

.github/workflows/integration_tests.yaml
```diff
@@ -8,6 +8,7 @@ jobs:
     uses: uc-cdis/.github/.github/workflows/integration_tests.yaml@master
     with:
       SERVICE_TO_TEST: fence
+      TEST_REPO_BRANCH: feat/DRS_15
```
k-burt-uch: Remove this line before merging

fence/blueprints/ga4gh.py
```diff
@@ -1,12 +1,77 @@
...
 )
 
-blueprint = flask.Blueprint("ga4gh", __name__)
+
+class BulkObjectAccessIds(BaseModel):
+    bulk_object_id: str
+    bulk_access_ids: list[str]
+
+
+class BulkObjectAccessRequest(BaseModel):
```
k-burt-uch, line 42: Add docs to these classes and leave a TODO to remove for the Fast API conversion. Include the ticket https://ctds-planx.atlassian.net/browse/PD-151

k-burt-uch (approved):

PR #1346, opened 2026-04-21
Apply cascading authz for visas

k-burt-uch:

### Bug Fixes
- Fix that applies child authorizations for parent studies when passports are configured without dbGaP sftp sync.


Avantol13 (approved):
lgtm. I double checked that this wouldn't result in a pure dbgap sync making 2 calls, but this code is only used for single_user_sync-ing so we should be good.

PR #1358, opened 2026-06-10
Optionally fetch userinfo from OIDC IdP

Jasper51297:
This pull request enables Fence to propagate specific attributes from upstream OIDC Identity Providers (IdPs) into the ID tokens issued to clients. Previously, certain attributes returned by external providers during the OIDC flow were captured by Fence but not made accessible to the downstream applications.

This change introduces a configurable mapping that allows administrators to specify which fields from the upstream `userinfo` response should be included as claims in the Fence-generated ID token.

https://openid.net/specs/openid-connect-core-1_0.html#UserInfo

Minimal Fence (helm) example:
```
fence:
  FENCE_CONFIG:
    OPENID_CONNECT:
      sram:
        name: 'SRAM'
        client_id: 'myclientid'
        redirect_url: '{{BASE_URL}}/login/sram/login'
        discovery_url: 'https://proxy.sram.surf.nl/.well-known/openid-configuration'
        scope: 'openid profile email voperson_external_id'
        user_id_field: 'userinfo.email'
        email_field: 'userinfo.email'
```

### New Features
- Added `idp_attr_mapping` configuration to `OPENID_CONNECT`.
- Support for including upstream IdP attributes in Fence ID tokens.

### Breaking Changes
None

### Bug Fixes
None

### Improvements
- Improved user session metadata handling to store upstream `userinfo`.

### Dependency updates
None

### Deployment changes
None


nrandorf:
Good afternoon, are you able to address the three failing pytests?

Jasper51297:
Sorry, did not get notified some tests were failing. This should have been resolved now.

nrandorf:
Thank you for the update!

fence/resources/openid/idp_oauth2.py
```diff
@@ -271,21 +324,25 @@ def get_auth_info(self, code):
                     )
                     raise Exception(e)
 
-            if claims.get(user_id_field):
-                if user_id_field == "email" and not claims.get("email_verified"):
```
Avantol13, line 275: You cannot remove this check, it is required when the email field is found and the verified field is found. Google does this to protect against unverified emails

fence/resources/openid/idp_oauth2.py
```diff
@@ -271,21 +324,25 @@ def get_auth_info(self, code):
...
-                    return {"error": "Email is not verified"}
+            user_id_value = self._get_claim_value(user_id_field, claims, userinfo)
+
+            if user_id_value:
```
Avantol13, line 329: why is this whole return gated behind finding that user_id_value now? Before I believe it just prompted the additional email check?

fence/resources/openid/idp_oauth2.py
```diff
@@ -100,6 +100,46 @@ def get_jwt_keys(self, jwks_uri):
...
+        Args:
+            raw_access_token (str): The access token to use for authentication.
+
+        Returns:
+            dict: The JSON response from the userinfo endpoint, or an empty dict if the request fails.
+        """
+        userinfo_endpoint = self.get_value_from_discovery_doc("userinfo_endpoint", "")
+        if userinfo_endpoint and raw_access_token:
+            header = {"Authorization": "Bearer " + raw_access_token}
+            res = requests.get(userinfo_endpoint, headers=header, proxies=self.get_proxies())
```
Avantol13, line 116: I would recommend adding exponential backoff and error handling explicitly since this is an external call.

fence/resources/openid/idp_oauth2.py
```diff
@@ -100,6 +100,46 @@ def get_jwt_keys(self, jwks_uri):
...
+            if res.status_code == 200:
+                self.logger.info(f"Successfully retrieved userinfo from {userinfo_endpoint}")
+                return res.json()
+            else:
+                self.logger.error("Unable to get userinfo: status_code: {}, message: {}".format(res.status_code, res.text))
+        return {}
```
Avantol13, line 122: failure case return pattern should be avoided. Reverse this logic, fail fast on errors and the default return should be the positive case

tests/login/test_microsoft_login.py
Avantol13, line 1: We need new unit tests to test the new functionality. Change the config and test that what you added achieves the intended results

fence/resources/openid/idp_oauth2.py
```diff
@@ -271,21 +324,25 @@ def get_auth_info(self, code):
...
                     "refresh_token": refresh_token,
                     "iat": claims.get("iat"),
                     "exp": claims.get("exp"),
                     "groups": groups,
                     "group_prefix": group_prefix,
-                    "org": claims.get(organization_claim_field),
-                    "firstname": claims.get(firstname_claim_field),
-                    "lastname": claims.get(lastname_claim_field),
-                    "email": claims.get(email_claim_field),
+                    organization_claim_field: self._get_claim_value(organization_claim_field, claims, userinfo),
```
Avantol13, line 338: This could cause implicit overriding of dictionary fields and confuses purpose. `organization_claim_field` is intended to be the field within the IdP claims to get the organization information from, we are standardizing the location in Gen3's response to "org". Now you are mirroring and duplicating the parsing if the "org" field is in fact "org", e.g. this conflicts with the line explicitly standardizing on "org" below.

Can you please explain more context around why this is required? Can you not simply provide this existing configuration to get the org information from the right claim and then parse the "org" from the "org" field?

Avantol13 (changes requested):

PR #1361, opened 2026-06-16
Feature/onboarding version seen

sonali-shaw:

### New Features
- the onboardingVersionSeen field has been added to the additional_info of a user, allowing the frontend to determine what phase of onboarding to display to the user


michaelfitzo:
Closed after discussion with user that it was meant for a separate fork instead of to master

PR #1366, opened 2026-08-05
Fix authz check when requesting a task token

paulineribeyre:

### Bug Fixes
- Fix authz check when requesting a task token

### DIFF
--- a/fence/authz/auth.py
+++ b/fence/authz/auth.py
@@ -12,7 +12,7 @@ from fence.jwt.utils import get_jwt_header
 logger = get_logger(__name__)
 
 
-def authorize(resource, method):
+def authorize(resource, method, token=None):
     """
     Check with arborist to verify the authz for a request. Throws a ``Forbidden`` error if the user is not authorized to access the resource.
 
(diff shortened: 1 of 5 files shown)


fence/authz/auth.py
```diff
@@ -25,17 +25,20 @@ def authorize(resource, method):
...
             " this endpoint is unavailable"
         )
-    if "Authorization" not in flask.request.headers:
-        logger.debug("request missing Authorization header; treating as anonymous")
-        token = None
-    else:
-        token = get_jwt_header()
+    if not token:
+        if "Authorization" not in flask.request.headers:
+            logger.debug("request missing Authorization header; treating as anonymous")
```
paulineribeyre, line 38: There's no access token in the request when you use an API key to get a token. So all task token requests were treated as anonymous.

fence/blueprints/storage_creds/api.py
```diff
@@ -204,15 +204,21 @@ def post(self):
...
-                )
 
         result = create_user_access_token(
             flask.current_app.keypairs[0],
             api_key,
             expires_in,
             task_token_type=task_token_type,
         )
+
+        # we generate the result token BEFORE checking access, because a token is required to
```
Avantol13, line 215: but this means someone without access can spam this endpoint and we'll keep creating tokens (which is sorta an expensive operation). partially why I shifted to requiring it right away. The client should be able to use the API key to easily get an access token. I guess they could spam that endpoint to keep getting access tokens which is a similar problem... so maybe this is fine

paulineribeyre: I see what you're saying. Having users get a regular token first, and then attach it to the task token request might make some sense, but like you said it doesn't get rid of the problem and it complicates UX

Avantol13: Arguably a user shouldn't be hitting this themselves though- should be our clients, but if API access is open - it is still a weird difference from the non task token request

Avantol13 (approved):

fence/blueprints/storage_creds/api.py
paulineribeyre: for future reference about this discussion, see [here](https://cdis.slack.com/archives/C02SH3UB2T0/p1786055672081909?thread_ts=1785953406.385109&cid=C02SH3UB2T0)

PR #1383, opened 2026-09-21
Fix issue where session cookie's expiration does not match JWT token's

craigrbarnes:
Link to JIRA ticket if there is one: N/A

# Bug Report: Session Cookie Expires Attribute Stale Relative to JWT exp

fence/resources/user/user_session.py
Severity: Medium — causes premature browser-side session termination

---

## Background

Fence stores sessions as JWTs embedded in a browser cookie named fence. On every response, the session is renewed: a new JWT is signed with exp = now + SESSION_TIMEOUT, and the cookie is rewritten with the new JWT as its value and the new exp as its Expires attribute.

The cookie Expires is read from session.session_token["exp"] via get_expiration_time().

---

## Root Cause

get_updated_token() signs a new JWT and stores the encoded string in self._encoded_token, but never updates self.session_token (the decoded dict). As a result, get_expiration_time() still reads the previous token's exp when setting the cookie Expires.

save_session():
  1. token = session.get_updated_token(app)
     └─ creates new JWT (exp = now + SESSION_TIMEOUT)
     └─ sets self._encoded_token = new_token
     └─ ⚠️  self.session_token still holds the OLD decoded payload

  2. response.set_cookie(
         value=token,                              # ← NEW JWT (correct exp)
         expires=get_expiration_time(session)      # ← reads OLD session_token["exp"]
     )

---

## Effect

The cookie's Expires attribute is set to the previous renewal's expiration. Each renewal the cookie value advances by SESSION_TIMEOUT, but the Exces by the time elapsed since the last request. Under normal interactive use (requests every few minutes) the drift accumulates until the browser
discards the cookie well before the JWT inside it expires.

In the observed token:
- JWT exp: 2026-09-19 01:21:36 UTC
- Cookie Expires: 2026-09-19 01:01:35 UTC (20 min earlier)

The browser dropped the session 20 minutes before fence would have considered it expired.

---

## Demonstration

```python
import time
import jwt  # PyJWT

# Simulate the stale-exp bug

SESSION_TIMEOUT = 1800  # seconds

def create_session_token(expires_in, context):
    """Simplified stand-in for fence's create_session_token."""
    now = int(time.time())
    payload = {
        "pur": "session",
        "iat": now,
        "exp": now + expires_in,
        "context": context,
    }
    # In real fence this is RS256-signed; using HS256 here for demo simplicity
    return jwt.encode(payload, "secret", algorithm="HS256")


class UserSession:
    def __init__(self, encoded_token):
        self._encoded_token = encoded_token
        self.session_token = jwt.decode(
            encoded_token, "secret", algorithms=["HS256"]
        )

    # --- BUGGY VERSION ---
    def get_updated_token_buggy(self):
        if self._encoded_token:
            token = create_session_token(SESSION_TIMEOUT, self.session_token["context"])
            self._encoded_token = token
            # BUG: self.session_token is never updated
        return self._encoded_token

    # --- FIXED VERSION ---
    def get_updated_token_fixed(self):
        if self._encoded_token:
            token = create_session_token(SESSION_TIMEOUT, self.session_token["context"])
            self._encoded_token = token
            self.session_token = jwt.decode(  # ← sync the decoded dict
                token, algorithms=["HS256"],
                options={"verify_signature": False},
            )
        return self._encoded_token

    def get_cookie_expires(self):
        """Mirrors get_expiration_time() — reads session_token["exp"]."""
        return self.session_token["exp"]


# --- Reproduce the bug ---

initial_token = create_session_token(SESSION_TIMEOUT, {"session_started": int(time.time())})
session = UserSession(initial_token)

time.sleep(2)  # Simulate time passing between requests

# First renewal
session.get_updated_token_buggy()
jwt_exp_1 = jwt.decode(session._encoded_token, "secret", algorithms=["HS256"])["exp"]
cookie_exp_1 = session.get_cookie_expires()

print("After 1st renewal (buggy):")
print(f"  JWT exp in cookie value : {jwt_exp_1}")
print(f"  Cookie Expires attribute: {cookie_exp_1}")
print(f"  Drift                   : {jwt_exp_1 - cookie_exp_1}s  ← should be 0")

time.sleep(2)

# Second renewal — drift grows
session.get_updated_token_buggy()
jwt_exp_2 = jwt.decode(session._encoded_token, "secret", algorithms=["HS256"])["exp"]
cookie_exp_2 = session.get_cookie_expires()

print("\nAfter 2nd renewal (buggy):")
print(f"  JWT exp in cookie value : {jwt_exp_2}")
print(f"  Cookie Expires attribute: {cookie_exp_2}")
print(f"  Drift                   : {jwt_exp_2 - cookie_exp_2}s  ← grows each renewal")

# --- Verify the fix ---

session2 = UserSession(initial_token)
time.sleep(2)
session2.get_updated_token_fixed()
jwt_exp_f = jwt.decode(session2._encoded_token, "secret", algorithms=["HS256"])["exp"]
cookie_exp_f = session2.get_cookie_expires()

print("\nAfter 1st renewal (fixed):")
print(f"  JWT exp in cookie value : {jwt_exp_f}")
print(f"  Cookie Expires attribute: {cookie_exp_f}")
print(f"  Drift                   : {jwt_exp_f - cookie_exp_f}s  ← always 0")

```

### Example output:
After 1st renewal (buggy):
  JWT exp in cookie value : 1789780898
  Cookie Expires attribute: 1789779096   ← old token's exp
  Drift                   : 1802s        ← grows with each renewal

After 2nd renewal (buggy):
  JWT exp in cookie value : 1789782700
  Cookie Expires attribute: 1789779096   ← still the original exp!
  Drift                   : 3604s

After 1st renewal (fixed):
  JWT exp in cookie value : 1789780898
  Cookie Expires attribute: 1789780898   ← always in sync
  Drift                   : 0s

---

## Fix Applied

In get_updated_token(), after signing the new JWT, the decoded payload is synced back into self.session_token:
```python
self.session_token = jwt.decode(
    token,
    algorithms=["RS256"],
    options={"verify_signature": False},
)
```
Signature verification is intentionally skipped here — the token was just signed by this process, so re-verifying would be redundant.


wangfan860 (approved):

fence/resources/user/user_session.py
```diff
@@ -106,6 +106,14 @@ def get_updated_token(self, app):
                 self.session_token["context"],
             )
             self._encoded_token = token
+            # Sync the decoded dict so get_expiration_time reads the new exp,
+            # not the previous token's exp (which would make the cookie Expires
+            # attribute stale relative to the JWT value in the cookie).
+            self.session_token = jwt.decode(
```
BinamB, line 112: When we're setting the initial session token [here](https://github.com/uc-cdis/fence/blob/master/fence/resources/user/user_session.py#L95), we're calling `validate_jwt` [here](https://github.com/uc-cdis/fence/blob/master/fence/resources/user/user_session.py#L85). To be consistent with how we're setting the session_token, i think we should use `validate_jwt` here as well.

Avantol13: We don't need to. We just generated it like 5 lines above, there's no need to waste CPU cycles to validate it in this situation.

BinamB (approved):

